PR #1053, opened 2014-12-15
deprecates commodity checking and related API in bid portfolios

gidden:
For @mbmcgarry, this removes commodity checking in bid portfolios. This is a check that used to necessary, bid portfolios used to require a single commodity, but that has been relaxed, and this was missed on that update. This PR preserves ABI stability, noting what is deprecated.


rwcarlsen:
This could use a rebase.  I wouldn't normally mind to much, but it's been long enough (sorry) that this is quite a bit behind.

gidden:
rebased

PR #1054, opened 2014-12-16
lazy decay option and misc cleanups and tweaks

rwcarlsen:
* lazy decay option decays materials every time the Material::comp function is
  called.

* Added ability to call Material::Decay with a negative time argument that
  causes the decay operation to run with the material's context's current
  time.  Throws if the material has a NULL context.

* fixed sim_init tests - they previously counted on optional normalization for
  material compositions that has now changed.  Note that API is identical -
  the doc comments for the relevant functions explicitly states that returned
  composition are not normalized.

### DIFF
--- a/src/composition.cc
+++ b/src/composition.cc
@@ -64,7 +64,7 @@ const CompMap& Composition::mass() {
 Composition::Ptr Composition::Decay(int delta) {
   int tot_decay = prev_decay_ + delta;
   if (decay_line_->count(tot_decay) == 1) {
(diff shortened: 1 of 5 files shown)


rwcarlsen:
With @gidden's test modifications in cycamore, this no longer causes failures and should be good for merge.

rwcarlsen:
ping - ready.

rwcarlsen:
merge please - @scopatz asked if @gidden could pull the trigger when ready.

gidden:
looks good and ABI stability passes

PR #1062, opened 2014-12-29
Raw Recorder

scopatz:
This adds a subclass of recorder that does not inject the SimId and gives you raw table writing. This is needed for the more general case of writing to the database, and in particular for cymetric or any other analysis tool. In such situations SimIds are not necessarily semantically meaningful. For example, aggregation over many simulations will make any SimId field meaningless. Since folks may want to store any kind of data a raw recorder is needed. A subclass API turned out to be cleaner than adding new member functions.

This is dependent on #1061, which should be merged in first.

### DIFF
--- a/src/datum.h
+++ b/src/datum.h
@@ -14,6 +14,7 @@ namespace cyclus {
 /// Recorder for recording.
 class Datum {
   friend class Recorder;
+  friend class RawRecorder;
 
  public:
   typedef std::pair<const char*, boost::spirit::hold_any> Entry;
--- a/src/datum.h
+++ b/src/datum.h
@@ -36,6 +36,7 @@ class Datum {
   /// shape for this field. This is only useful for variable length data types
   /// such as string and blob. If a shape is provided, this field and value
   /// is interpreted as having a fixed length (of the value given). If the
+
   /// pointer is NULL or the entry is less than one (<1), the field is interpreted
   /// as inherently variable length, which may affect persistance. This is a
   /// vector of ints (a shape) rather than an int (a length) to accomodate
--- a/src/datum.h
+++ b/src/datum.h
@@ -14,7 +14,6 @@ namespace cyclus {
 /// Recorder for recording.
 class Datum {
   friend class Recorder;
-  friend class RawRecorder;
 
  public:
   typedef std::pair<const char*, boost::spirit::hold_any> Entry;
--- a/src/hdf5_back.cc
+++ b/src/hdf5_back.cc
@@ -1541,6 +1541,33 @@ void Hdf5Back::CreateTable(Datum* d) {
   schemas_[d->title()] = dbtypes;
 }
 
+std::map<std::string, DbTypes> Hdf5Back::ColumnTypes(std::string table) {
+  using std::string;
+  int i;
+  char* colname;
+  hid_t dset = H5Dopen2(file_, table.c_str(), H5P_DEFAULT);
+  hid_t dt = H5Dget_type(dset);
+  hsize_t ncols = H5Tget_nmembers(dt);
+  string fieldname;
+  string fieldtype;
+  LoadTableTypes(table, dset, ncols);
+  DbTypes* dbtypes = schemas_[table];
+
+  // create return value
+  std::map<string, DbTypes> rtn;
+  for (i = 0; i < ncols; ++i) {
+    colname = H5Tget_member_name(dt, i);
+    fieldname = string(colname);
+    free(colname);
+    rtn[fieldname] = dbtypes[i];
+  }
+
+  // close and return
+  H5Tclose(dt);
+  H5Dclose(dset);
+  return rtn;
+}
+
 void Hdf5Back::WriteGroup(DatumList& group) {
   std::string title = group.front()->title();
   const char * c_title = title.c_str();
--- a/src/recorder.cc
+++ b/src/recorder.cc
@@ -10,17 +10,23 @@
 
 namespace cyclus {
 
-Recorder::Recorder() : index_(0) {
+Recorder::Recorder() : index_(0), inject_sim_id_(true) {
   uuid_ = boost::uuids::random_generator()();
   set_dump_count(kDefaultDumpCount);
 }
 
-Recorder::Recorder(unsigned int dump_count) : index_(0) {
+Recorder::Recorder(bool inject_sim_id) : index_(0), inject_sim_id_(inject_sim_id) {
+  uuid_ = boost::uuids::random_generator()();
+  set_dump_count(kDefaultDumpCount);
+}
+
+Recorder::Recorder(unsigned int dump_count) : index_(0), inject_sim_id_(true) {
   uuid_ = boost::uuids::random_generator()();
   set_dump_count(dump_count);
 }
 
-Recorder::Recorder(boost::uuids::uuid simid) : index_(0), uuid_(simid) {
+Recorder::Recorder(boost::uuids::uuid simid) : index_(0), uuid_(simid), \
+                                               inject_sim_id_(true) {
   set_dump_count(kDefaultDumpCount);
 }
 
@@ -52,7 +58,9 @@ void Recorder::set_dump_count(unsigned int count) {
   data_.reserve(count);
   for (int i = 0; i < count; ++i) {
     Datum* d = new Datum(this, "");
-    d->AddVal("SimId", uuid_);
+    if (inject_sim_id_) {
+      d->AddVal("SimId", uuid_);
+    }
     data_.push_back(d);
   }
   dump_count_ = count;
@@ -61,8 +69,13 @@ void Recorder::set_dump_count(unsigned int count) {
 Datum* Recorder::NewDatum(std::string title) {
   Datum* d = data_[index_];
   d->title_ = title;
-  d->vals_.resize(1);
-  d->shapes_.resize(1);
+  if (inject_sim_id_) {
+    d->vals_.resize(1);
+    d->shapes_.resize(1);
+  } else {
+    d->vals_.resize(0);
+    d->shapes_.resize(0);
+  }
 
   index_++;
   return d;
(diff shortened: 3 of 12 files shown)


scopatz:
ping, this has been updated and is ready for another look.

scopatz:
Ping! This needs a prompt review, please.

scopatz:
Ping again!

src/datum.h
```diff
@@ -14,6 +14,7 @@ namespace cyclus {
 /// Recorder for recording.
 class Datum {
   friend class Recorder;
+  friend class RawRecorder;
```
gidden: this looks to now be extraneous

src/hdf5_back.cc
```diff
@@ -1541,6 +1541,33 @@ void Hdf5Back::CreateTable(Datum* d) {
   schemas_[d->title()] = dbtypes;
 }
 
+std::map<std::string, DbTypes> Hdf5Back::ColumnTypes(std::string table) {
```
gidden: does this function belong in this PR? perhaps it should be rebased on the current develop head?

scopatz: It has been rebased. 

```bash
$ git pull upstream develop 
From github.com:cyclus/cyclus
 * branch            develop    -> FETCH_HEAD
Already up-to-date.
scopatz@ares ~/cyclus develop $ git checkout rr
Switched to branch 'rr'
Your branch is up-to-date with 'origin/rr'.
scopatz@ares ~/cyclus rr $ git rebase develop 
Current branch rr is up to date.
```

Not sure why this is here. Maybe @rwcarlsen rebased and force pushed? It is on develop. https://github.com/cyclus/cyclus/commit/8c72577f0fb29c03668a2ee1513270fa50d3dea5

src/recorder.h
```diff
@@ -75,6 +79,12 @@ class Recorder {
   /// returns the unique id associated with this cyclus simulation.
   boost::uuids::uuid sim_id();
 
+  /// returns whether or not the unique simulation id will be injected.
+  bool inject_sim_id() { return inject_sim_id_; };
+
+  /// sets whether or not the unique simulation id will be injected.
+  void set_inject_sim_id(bool x) { inject_sim_id_ = x; };
```
gidden: You expressed displeasure [previously](https://github.com/cyclus/cyclus/pull/641#discussion_r8024852) at this pattern. It was decided at that point to reject the `set_` prefixes in the kernel.

scopatz: I was matching the surrounding code. But yes, I can move this to an overload. And I do hate this pattern :).

src/recorder.h
```diff
@@ -46,16 +46,20 @@ class Recorder {
...
+
   /// create a new recorder with the given dump count and random
   /// simulation id.
   Recorder(unsigned int dump_count);
 
-  /// create a new recorder with default dump frequency and the specified
-  /// simulation id.
+  /// create a new recorder with default dump frequency. the specified
+  /// simulation id, and simulation id injection.
   Recorder(boost::uuids::uuid simid);
```
gidden: @scopatz, you can use default arguments [in constructors](http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Default_Arguments). I think some slight retooling here (as you see fit) would be nice, rather than having on constructor per argument. Also, I think it would be nice to have a constructor that takes all arguments (non default).

gidden: @scopatz, you can use default arguments [in constructors](http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Default_Arguments). I think some slight retooling here (as you see fit) would be nice, rather than having on constructor per argument. Also, I think it would be nice to have a constructor that takes all arguments (non default).

src/recorder.cc
```diff
@@ -61,8 +69,13 @@ void Recorder::set_dump_count(unsigned int count) {
 Datum* Recorder::NewDatum(std::string title) {
   Datum* d = data_[index_];
   d->title_ = title;
-  d->vals_.resize(1);
-  d->shapes_.resize(1);
+  if (inject_sim_id_) {
```
gidden: I think a ternary operator would be cleaner here, i.e. `int sz = inject_sim_id_ ? 1 : 0;`.

src/recorder.h
scopatz: Good to know that they are allowed. However, it is allowed written and consistent with what was there before. If more constructors are not blocking then I can't do this.  I am effectively out of time for this and for any nice-to-have features.  I need to finish writing skeleton cymetric docs.

gidden: Then, at minimum, I would suggest a single constructor with all possible
(non default) args.

On Sun, Jan 4, 2015 at 9:01 PM, Anthony Scopatz <notifications@github.com>
wrote:

> In src/recorder.h
> <https://github.com/cyclus/cyclus/pull/1062#discussion-diff-22446380>:
>
> >    /// create a new recorder with the given dump count and random
> >    /// simulation id.
> >    Recorder(unsigned int dump_count);
> >
> > -  /// create a new recorder with default dump frequency and the specified
> > -  /// simulation id.
> > +  /// create a new recorder with default dump frequency. the specified
> > +  /// simulation id, and simulation id injection.
> >    Recorder(boost::uuids::uuid simid);
>
> Good to know that they are allowed. However, it is allowed written and
> consistent with what was there before. If more constructors are not
> blocking then I can't do this. I am effectively out of time for this and
> for any nice-to-have features. I need to finish writing skeleton cymetric
> docs.
>
> —
> <https://github.com/cyclus/cyclus/pull/1062/files#r22446380>.
>



-- 
Matthew Gidden
Ph. 225.892.3192

src/recorder.cc
scopatz: That put another variable on the stack in an often called function and I don't think the C/C++?Javascript ternary operator is readable.

src/recorder.h
scopatz: Why? This is the minimum solution?

scopatz: And matches the pattern established for this class....

src/recorder.cc
gidden: I disagree, but I've expressed my opinion. @rwcarlsen is welcome to comment. Further, we value vertical space and dislike premature optimization. If you'd really like to optimize, another state variable could be added (`resize_`), and it could be set in constructors.

src/recorder.h
gidden: You are the first user to find a case limiting for this class. A future user may want similar functionality, but with a different `dump count`, for example.

src/recorder.cc
scopatz: I don't think that we should be dogmatic about these things. No matter what way you cut it, this is minor. Also it is strictly in the implementation so anyone is welcome to change it in the future.

src/recorder.h
scopatz: Then that person can write their own constructor and test it, like I did here. I am seriously getting mixed messages about the scope of this work.

rwcarlsen:
rebase on develop HEAD please.

rwcarlsen:
why does this have col-types stuff in it still? - it is rebased - hmm....

rwcarlsen:
why does this have col-types stuff in it still? - it is rebased - hmm....

src/recorder.cc
rwcarlsen: I am partial to ifs myself, but I don't think this is a big deal either way.

src/recorder.h
rwcarlsen: Currently, there is no way to create a non-injecting recorder with a specific simulation id.  And rather than having a bunch of constructors trying to anticipate every permutation of configuration - of many, many argument constructors - why not just add a single member function (and add+modify no constructors) like "NoIdInject()"?

scopatz:
@rwcarlsen - other than the weird rebase stuff, do you think this is ready to merge in?

src/recorder.h
gidden: With the getter/setter (already implemented), I'm happy with this.

gidden:
I'm good for merging.

src/recorder.h
```diff
@@ -97,7 +107,7 @@ class Recorder {
   /// Unregisters all backends and resets.
   void Close();
 
- private:
```
rwcarlsen: back to private again now that there are no subclasses?

scopatz: I just saw this. I'll change it back.

src/recorder.h
```diff
@@ -75,6 +79,12 @@ class Recorder {
   /// returns the unique id associated with this cyclus simulation.
   boost::uuids::uuid sim_id();
 
+  /// returns whether or not the unique simulation id will be injected.
+  bool inject_sim_id() { return inject_sim_id_; };
+
+  /// sets whether or not the unique simulation id will be injected.
+  void inject_sim_id(bool x) { inject_sim_id_ = x; };
```
rwcarlsen: This needs to call `set_dump_count` or the datums will be generated wrong - and the resizing could result in random junk being recorded to database.

src/recorder.h
scopatz: changed.

tests/recorder_tests.cc
```diff
@@ -56,6 +56,15 @@ TEST(RecorderTest, Manager_GetSetDumpFreq) {
   EXPECT_EQ(m.dump_count(), cyclus::kDefaultDumpCount);
 }
 
+TEST(RecorderTest, InjectSimId) {
```
rwcarlsen: This test needs to check whether a recorded datum actually has a sim id injected or not.  Such a test would have caught the missing `set_dump_count` call in the `inject_sim_id(yourval)` function.

src/recorder.h
rwcarlsen: Ah - yep - I didn't see that initially.  But it does have a bug - see my other comment.

src/recorder.h
rwcarlsen: or even segfaults maybe.

rwcarlsen:
This looks good to me - just needs the bug fix and test for said bug.

src/recorder.h
scopatz: Just calling `set_dump_count()` will lose data, which is why I didn't do it. I'll `Flush()` and then set, though, as that is probably the safest option.

scopatz:
Ok the bug fix and stronger tests are now in.

scopatz:
This passes CI.

PR #1064, opened 2015-01-07
modify resbuf code generation. Add missing ResCast funcs.

rwcarlsen:
ResBuf prior to this commit was broken.  It tried to use the an overload
of the ResCast function in resource.h that didn't exist.  Since the ResCast
overload needed isn't terribly likely to be needed outside of resource
buffer(s), I decided it fit well in the toolkit.

Previously, the code generated by cycpp for setting resource buffer capacities
in InitFrom (db) assumed that the 'capacity' key for the buffer member
dictionary was a literal value stored in the database explicitly.  This
relaxes that assumption and instead assumes the 'capacity' key is literal code
that should be pasted in.  Now arbitrary expressions, etc. can be used to
compute capacities based on other member variables.  The tradeoff is that all
referenced member variables in these expressions must be defined/pragma'd
prior/above the resource buffer.

### DIFF
--- a/cli/cycpp.py
+++ b/cli/cycpp.py
@@ -949,11 +949,11 @@ class InitFromDbFilter(CodeGeneratorFilter):
             impl += ind + '{0} = qr.GetVal<{1}>("{0}");\n'.format(member, tstr)
         for b, info in cap_buffs.items():
             if isinstance(info['type'], STRING_TYPES):  # ResourceBuff
-                impl += ind + ('{0}.set_capacity(qr.GetVal<double>'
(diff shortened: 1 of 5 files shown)


scopatz:
Hi @rwcarlsen, could we get some unit tests for the new `toolkit::ResCast()` functions, please? Thanks.

scopatz:
Looks good otherwise.

rwcarlsen:
added tests.

scopatz:
Thanks @rwcarlsen

PR #1065, opened 2015-01-07
added pair<int, string> for starters

scopatz:
First type.

### DIFF
--- a/src/hdf5_back.cc
+++ b/src/hdf5_back.cc
@@ -532,6 +532,33 @@ QueryResult Hdf5Back::Query(std::string table, std::vector<Cond>* conds) {
               row[j] = x;
             break;
           }
+          case PAIR_INT_STRING: {
+            size_t nullpos;
+            unsigned int strlen = col_sizes_[table][j] - sizeof(int);
+            int xfirst = *reinterpret_cast<int*>(buf + offset);
+            string s = string(buf + offset + sizeof(int), strlen);
+            nullpos = s.find('\0');
+            if (nullpos != std::string::npos)
+              s.resize(nullpos);
+            pair<int, string> x = std::make_pair(xfirst, s);
+            is_row_selected =
+                CmpConds<pair<int, string> >(&x, &(field_conds[qr.fields[j]]));
+            if (is_row_selected)
+              row[j] = x;
+            break;
+          }
+          case PAIR_INT_VL_STRING: {
+            unsigned int itemsize = sizeof(int) + CYCLUS_SHA1_SIZE;
+            jlen = col_sizes_[table][j] / itemsize;
+            pair<int, string> x = std::make_pair(
+              *reinterpret_cast<int*>(buf + offset),
+              VLRead<string, VL_STRING>(buf + offset + sizeof(int)));
+            is_row_selected =
+                CmpConds<pair<int, string> >(&x, &(field_conds[qr.fields[j]]));
+            if (is_row_selected)
+              row[j] = x;
+            break;
+          }
           case MAP_INT_INT: {
             map<int, int> x = map<int, int>();
             jlen = col_sizes_[table][j] / (2*sizeof(int));
@@ -1223,6 +1250,24 @@ void Hdf5Back::CreateTable(Datum* d) {
       H5Tinsert(field_types[i], "first", 0, H5T_NATIVE_INT);
       H5Tinsert(field_types[i], "second", sizeof(int), H5T_NATIVE_INT);
       opened_types_.insert(field_types[i]);
+    } else if (valtype == typeid(pair<int, string>)) {
+      shape = shapes[i];
+      if (shape.empty() || shape[0] < 1) {
+        dbtypes[i] = PAIR_INT_VL_STRING;
+        field_types[i] = H5Tcreate(H5T_COMPOUND, sizeof(int) + CYCLUS_SHA1_SIZE);
+        H5Tinsert(field_types[i], "first", 0, H5T_NATIVE_INT);
+        H5Tinsert(field_types[i], "second", sizeof(int), sha1_type_);
+        opened_types_.insert(field_types[i]);
+        dst_sizes[i] = (sizeof(int) + CYCLUS_SHA1_SIZE);
+      } else {
+        dbtypes[i] = PAIR_INT_STRING;
+        hid_t str_type = CreateFLStrType(shape[0]);
+        field_types[i] = H5Tcreate(H5T_COMPOUND, sizeof(int) + shape[0]);
+        H5Tinsert(field_types[i], "first", 0, H5T_NATIVE_INT);
+        H5Tinsert(field_types[i], "second", sizeof(int), str_type);
+        opened_types_.insert(field_types[i]);
+        dst_sizes[i] = (sizeof(int) + shape[0]);
+      }
     } else if (valtype == typeid(std::map<int, int>)) {
       shape = shapes[i];
       hid_t item_type = H5Tcreate(H5T_COMPOUND, sizeof(int) * 2);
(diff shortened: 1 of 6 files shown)


scopatz:
Note that I have just added map<pair<int, string>, double>

rwcarlsen:
If this is to support cycamore archetypes, then it must also be implemented in sqlite backend before a release.

scopatz:
Obviously, but that does not affect this PR, which simply adds types to the typesystem. Traditionally, we have dealt with the backends separately.

scopatz:
Requesting that either @mbmcgarry or @jdangerx review this, as it is pretty innocuous and they should get practice reviewing and merging PRs. Thanks!

mbmcgarry:
I will ask @rcarlsen for some pro-tips on reviewing PRs and then take a look at it :)

scopatz:
ping @mbmcgarry @jdangerx

src/hdf5_back.cc
```diff
@@ -532,6 +532,33 @@ QueryResult Hdf5Back::Query(std::string table, std::vector<Cond>* conds) {
               row[j] = x;
             break;
           }
+          case PAIR_INT_STRING: {
+            size_t nullpos;
+            unsigned int strlen = col_sizes_[table][j] - sizeof(int);
+            int xfirst = *reinterpret_cast<int*>(buf + offset);
```
mbmcgarry: Google style guide says don't use unsigned int.

scopatz: Link?

mbmcgarry: On 01/09/15, Anthony Scopatz  wrote:
> 
> In src/hdf5_back.cc(https://github.com/cyclus/cyclus/pull/1065#discussion-diff-22731263):
> 
> > @@ -532,6 +532,33 @@ QueryResult Hdf5Back::Query(std::string table, std::vector<Cond>* conds) { > row[j] = x; > break; > } > + case PAIR_INT_STRING: { > + size_t nullpos; > + unsigned int strlen = col_sizes_[table][j] - sizeof(int); > + int xfirst = *reinterpret_cast<int*>(buf + offset); 
> 
> Link?

http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Integer_Types
subsection "On Unsigned Integers"

scopatz: There is no avoiding it here. The C++ STL doesn't listen to the GCSG and the primary place this variable is used is in a string constructor (http://www.cplusplus.com/reference/string/string/string/) which requires an unsigned int. Good to know about this restriction and now I have another reason to hate the GCSG :).

src/hdf5_back.cc
```diff
@@ -2866,4 +3108,42 @@ Hdf5Back::VLBufToVal<std::map<std::string, std::string> >(const hvl_t& buf) {
...
+
+hvl_t Hdf5Back::VLValToBuf(const std::map<std::pair<int, std::string>, double>& x) {
+  // VL_MAP_PAIR_INT_STRING_DOUBLE implemented as VL_MAP_PAIR_INT_VL_STRING_DOUBLE
+  hvl_t buf;
+  Digest keyhash;
+  buf.len = x.size();
+  size_t itemsize = sizeof(int) + CYCLUS_SHA1_SIZE + sizeof(double);
+  size_t nbytes = itemsize * buf.len;
+  buf.p = new char[nbytes];
+  unsigned int i = 0;
```
jdangerx: Is `unsigned int` here unavoidable too, or is this a situation where we can bow to GCSG and have `i` be signed?

jdangerx:
Looks good to me, beyond that unsigned int question.

src/hdf5_back.cc
scopatz: Here is not technically unavoidable, but it is a little silly.  `sizeof()` (http://en.cppreference.com/w/cpp/language/sizeof) returns a size_t, which is defined by the platform to be an unsigned integral type.  So making i signed would mean that the addition would have to cast one or the other of the values.

jdangerx: Got it - I was wondering if that was the case. Carry on, then.

jdangerx:
Thanks @scopatz !

mbmcgarry:
Ran into errors on the unit_tests, but they may be due to my version of hdf5 (1.8.13). Updating now to re-test.

mbmcgarry:
The unit tests have errors for the new variable types on my mac, but there are similar errors for existing variable types that were also present on the develop branch.  Robert and I aren't sure if this is a yosemite issue or something else, but they do not appear to be a result of this PR.

https://gist.github.com/mbmcgarry/c572b0e6b4c0e3617af7

scopatz:
What happens when you run just `$ cyclus_unit_tests --gtest_filter=Hdf5BackTests.ReadWriteVLMapVLStringDouble`

mbmcgarry:
Here you go:

https://gist.github.com/mbmcgarry/1bcad0e34974bbad8b09

scopatz:
Maybe it isn't flushing.  Try adding Flush() calls in the test after writing but before reading.

mbmcgarry:
Does this mean, do something like:

m.Flush() ;

around line 962 in hdf5_back_tests.cc ?

scopatz:
nevermind. It already is flushing where it should.  what does google say?

mbmcgarry:
This discussion suggests it may be a problem with HDF5 version 1.8.14:

https://groups.google.com/forum/#!topic/julia-users/rlNs3m8xXJ0

But I get the same error messages on the miniconda build. What version of HDF5 is minconda using? I can't see it on binstar, I think because it's a private package.

-Meghan

scopatz:
Try downgrading to HDF5 1.8.13 then...

mbmcgarry:
I have downgraded to 1.8.13 but it did not resolve the failed unit tests.  (However, it did allow Cycamore to build, which was not working with 1.8.14).

-Meghan


[  FAILED  ] 9 tests, listed below:
[  FAILED  ] Hdf5BackTests.ReadWriteVLMapVLStringDouble
[  FAILED  ] Hdf5BackTests.ReadWriteVLMapStringVLString
[  FAILED  ] Hdf5BackTests.ReadWriteVLMapVLStringString
[  FAILED  ] Hdf5BackTests.ReadWriteMapVLStringVLString
[  FAILED  ] Hdf5BackTests.ReadWriteVLMapVLStringVLString
[  FAILED  ] Hdf5BackTests.ReadWriteMapPairIntStringDouble
[  FAILED  ] Hdf5BackTests.ReadWriteVLMapPairIntStringDouble
[  FAILED  ] Hdf5BackTests.ReadWriteMapPairIntVLStringDouble
[  FAILED  ] Hdf5BackTests.ReadWriteVLMapPairIntVLStringDouble

scopatz:
if you have have pytables installed, what does the following return?

```bash
 $ python -c "import tables; print(tables.which_lib_version('zlib'))"
```

mbmcgarry:
Took forever to install pytables and related dependencies.  Here’s the result:


mbmcgarry@dyn-72-33-204-90 ~/git/cyclus/build/bin (develop) $ python -c "import tables; print(tables.which_lib_version('zlib'))"
(1, '1.2.5', None)

scopatz:
Hmm, are you still having issues after installing pytables?

mbmcgarry:
Yes, unfortunately.

mbmcgarry:
(and yes, I did a clean rebuild after installing pytables)

mbmcgarry:
Though I did get this very strange error when installing pytables:

ld: warning: ignoring file /opt/local/lib/libhdf5.dylib, file was built for x86_64 which is not the architecture being linked (i386): /opt/local/lib/libhdf5.dylib

mbmcgarry:
Here is my config.log file from the hdf5 build.

PR #1066, opened 2015-01-09
ResMap

scopatz:
Preliminary cut.  Docstrings need clean up. No code gen.

### DIFF
--- a/cli/cycpp.py
+++ b/cli/cycpp.py
@@ -897,7 +897,7 @@ class InitFromCopyFilter(CodeGeneratorFilter):
                 impl += info[self.pragmaname]
             elif info['type'] not in BUFFERS:
                 impl += ind + "{0} = m->{0};\n".format(member)
-            elif 'capacity' in info:
+            else:
                 cap_buffs[member] = info
 
         for b, info in cap_buffs.items():
@@ -916,9 +916,7 @@ class InitFromCopyFilter(CodeGeneratorFilter):
         CYCNS + '::toolkit::ResourceBuff': ("{var}.set_capacity("
                                             "m->{var}.capacity());\n"),
         CYCNS + '::toolkit::ResBuf': "{var}.capacity(m->{var}.capacity());\n",
-        CYCNS + '::toolkit::ResMap': (
-            "{var}.capacity(m->{var}.capacity());\n"
-            '{var}.obj_ids(m->obj_ids());\n'),
+        CYCNS + '::toolkit::ResMap': '{var}.obj_ids(m->obj_ids());\n',
         }
 
 class InitFromDbFilter(CodeGeneratorFilter):
@@ -984,7 +982,6 @@ class InitFromDbFilter(CodeGeneratorFilter):
         CYCNS + '::toolkit::ResourceBuff': '{var}.set_capacity({capacity});\n',
         CYCNS + '::toolkit::ResBuf': '{var}.capacity({capacity});\n',
         CYCNS + '::toolkit::ResMap': (
-            '{var}.capacity({capacity});\n'
             '{var}.obj_ids(qr.GetVal<{tstr}>("{var}"))\n;'),
         }
 
--- a/cli/cycpp.py
+++ b/cli/cycpp.py
@@ -89,9 +89,11 @@ PRIMITIVES = {'bool', 'int', 'float', 'double', 'std::string', 'cyclus::Blob',
               'boost::uuids::uuid', }
 
 BUFFERS = {'{0}::toolkit::ResourceBuff'.format(CYCNS),
+           '{0}::toolkit::ResBuf'.format(CYCNS),
            ('{0}::toolkit::ResBuf'.format(CYCNS), CYCNS + '::Resource'),
            ('{0}::toolkit::ResBuf'.format(CYCNS), CYCNS + '::Product'),
            ('{0}::toolkit::ResBuf'.format(CYCNS), CYCNS + '::Material'),
+           '{0}::toolkit::ResMap'.format(CYCNS),
            }
 
 TEMPLATES = {'std::vector', 'std::set', 'std::list', 'std::pair',
@@ -905,7 +907,8 @@ class InitFromCopyFilter(CodeGeneratorFilter):
             if t_impl is None:
                 msg = 'type {0!r} could not be found for InitFromCopy() code gen.'
                 raise TypeError(msg.format(t_info))
-            impl += ind + t_impl.format(var=b)
+            t_impl = t_impl.format(var=b)
+            impl += ind + t_impl.replace('\n', '\n' + ind).strip(' ')
 
         return impl
 
@@ -913,7 +916,9 @@ class InitFromCopyFilter(CodeGeneratorFilter):
         CYCNS + '::toolkit::ResourceBuff': ("{var}.set_capacity("
                                             "m->{var}.capacity());\n"),
         CYCNS + '::toolkit::ResBuf': "{var}.capacity(m->{var}.capacity());\n",
-        CYCNS + '::toolkit::ResMap': "{var}.capacity(m->{var}.capacity());\n",
+        CYCNS + '::toolkit::ResMap': (
+            "{var}.capacity(m->{var}.capacity());\n"
+            '{var}.obj_ids(m->obj_ids());\n'),
         }
 
 class InitFromDbFilter(CodeGeneratorFilter):
@@ -951,14 +956,18 @@ class InitFromDbFilter(CodeGeneratorFilter):
                 impl += info[self.pragmaname]
                 continue
             t = info['type']
-            if t in BUFFERS:
-                if 'capacity' in info:
-                    cap_buffs[member] = info
+            key = t if isinstance(t, STRING_TYPES) else t[0]
+            if key in BUFFERS:
+                if 'capacity' not in info:
+                    info = dict(info)
+                    info['capacity'] = 1e300
+                cap_buffs[member] = info
                 continue
             tstr = type_to_str(t)
             if tstr.endswith('>'):
                 tstr += ' '
             impl += ind + '{0} = qr.GetVal<{1}>("{0}");\n'.format(member, tstr)
+
         for b, info in cap_buffs.items():
             t_info = info['type']
             key = t_info if isinstance(t_info, STRING_TYPES) else t_info[0]
@@ -966,14 +975,17 @@ class InitFromDbFilter(CodeGeneratorFilter):
             if t_impl is None:
                 msg = 'type {0!r} could not be found for InitFromDb() code gen.'
                 raise TypeError(msg.format(t_info))
-            impl += ind + t_impl.format(var=b, capacity=info['capacity'])
+            t_impl = t_impl.format(var=b, tstr=type_to_str(t_info), **info)
+            impl += ind + t_impl.replace('\n', '\n' + ind).strip(' ')
 
         return impl
 
     res_impl = {
         CYCNS + '::toolkit::ResourceBuff': '{var}.set_capacity({capacity});\n',
         CYCNS + '::toolkit::ResBuf': '{var}.capacity({capacity});\n',
-        CYCNS + '::toolkit::ResMap': '{var}.capacity({capacity});\n',
+        CYCNS + '::toolkit::ResMap': (
+            '{var}.capacity({capacity});\n'
+            '{var}.obj_ids(qr.GetVal<{tstr}>("{var}"))\n;'),
         }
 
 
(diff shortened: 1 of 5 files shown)


src/toolkit/res_map.h
```diff
@@ -0,0 +1,189 @@
...
+///  private:
+///   ...
+///   cyclus::toolkit::ResMap    <cyclus::Material> inventory_;
+///   cyclus::toolkit::ResMap    <cyclus::Material> outventory_;
+/// };
+/// @endcode
+///
+/// In this example, if there is sufficient material in inventory_, 2703 kg is
+/// removed as a single object that is then placed in another buffer
+/// (outventory_) each time step.
```
rwcarlsen: doc comment needs to be updated with an appropriate example for ResMap.

gidden: I think the comment needs to effectively be rewritten for ResMap, but @scopatz pointed that out in his original comment.

rwcarlsen:
At a first glance, this looks okay, but there is an unaddressed issue relating to init/restart.  The resource buffer's only state is its ordered list of resources - there is some other state, but that state can be initialized from the ordered list of resources.  The kernel has no mechanism for initializing extra information w.r.t. resource holding classes.  More particularly, the std::map presents a problem.  The dirty flag doesn't affect observable behavior of the map beyond performance, so it is fine.

The best solution I can think of is to initialize The ResMap with a pointer to another, normal, member variable map that is persisted.  Something like this:

```c++
MyFacility : public Facility {
 public:
  EnterNotify() {
    mymap.Init(&foomap);
  }
 private:
  ResMap<std::string,Material> mymap;

  #pragma cyclus var {}
  std::map<std::string, Material::Ptr> foomap;
}
```
But even if we do something like that, this only provides a one-to-one mapping.  Otherwise map keys get overwritten.  As written, for example, I cannot associate a group of resources with a single key/metadatum.  Each key can only return one resource.  I think that prevents this from being useful for one of the largest use cases.  Do you have any ideas for how to address this?

scopatz:
> At a first glance, this looks okay, but there is an unaddressed issue relating to init/restart.

I said there was no code generation yet.  I would highly appreciate if you read my original post. An apology would be nice.

> Each key can only return one resource. I think that prevents this from being useful for one of the largest use cases. Do you have any ideas for how to address this?

I am not. This is not meant to be a ResBufMap.  That is a separate data structure with significantly different logic. Furthermore, it doesn't support the use case that I have. Furthermore, we all agreed that we need to be brutal about paring down the feature set. ResMaps are the minimal use case that I need. If it turns out that I do need a ResBufMap, I'll write that too.

rwcarlsen:
>> At a first glance, this looks okay, but there is an unaddressed issue relating to init/restart.

> I said there was no code generation yet. I would highly appreciate if you read my original post. An apology would be nice.

I know there is no code generation.  And I did read your original post.  I am sorry you were offended - I had no such intent.  I was just considering this PR with an eye toward how we will handle init/restart when the time comes - which it must soon if we are to use this at all.  The ResBuf has no state beyond resources - so code-gen for it is relatively simple.  That will not be the case here.  Are we okay with that?  I'm not sure it is a direction we want to go without some consideration.  We need to make sure that the code-gen problems here are solvable in way that works well with the kernel, ui's, and doesn't unnecessarily tangle up the cycpp code.

scopatz:
> I am sorry you were offended - I had no such intent. 

Apology accepted. 

> The ResBuf has no state beyond resources - so code-gen for it is relatively simple. That will not be the case here. Are we okay with that? I'm not sure it is a direction we want to go without some consideration. We need to make sure that the code-gen problems here are solvable in way that works well with the kernel, ui's, and doesn't unnecessarily tangle up the cycpp code.

Well, for starters, the user/archetype developer should not have to create any supplementary data structures on their Agents to persist/depersist ResMaps.  I think that defeats the purpose of having this be a managed data structure for convience. It should be sufficient for ResMaps to be state variables. In fact, I think a lof of the issues I was having with Katy's FCO Separations facility was that it was trying to manage metadata on top of internal (non-state var) ResourceBuffs.  The logic is intense.

I really see three options for how to automatically manage persistence and persistence of all Resource container classes (ResourceBuff, ResBuf, ResMap, ResBufMap, etc).

1. Make the classes DB-aware and responsible for persisting and depersisting themselves. I think we are in agreement that this is a Bad Idea (tm).
2. Code generate specifically for these classes. For better or worse, this is what we do now.
3. Make these classes proper types in the typesystem (a la `cyclus::Blob`) and let the backends handle everything.

I am dead set against (1), while (2) & (3) I don't have much preference for. Abstractly though I probably prefer (3), it is not worth changing how ResBuf and ResourceBuff.  Furthermore, I think it is important to be consistent in how these things are handled, which is why I was going to go with code generation.

If you have any other suggestions or strong feelings please let me know. Otherwise, I'll just go with code gen as it is the established mechanism.

rwcarlsen:
1 is no go for me too.  3 done naively won't work because we don't use normal mechanisms to store resource state - we use the kernel's Resources and Compositions tables because those tables already exists (no need to duplicate) store.  So doing 3 right would be tricky and take some work to make everything interoperate with init and the kernel properly.  2 (codegen) sounds good to me.

src/toolkit/res_map.h
```diff
@@ -0,0 +1,191 @@
...
+class ResMap {
+ public:
+  ResMap() : capacity_(INFINITY), dirty_quantity_(true), quantity_(0) {
+    Warn<EXPERIMENTAL_WARNING>("ResMap is experimental and its API may be "
+                               "subject to change");
+  }
+
+  virtual ~ResMap() {}
+
+  typedef typename std::map<K, typename R::Ptr> map_type;
```
rwcarlsen: You need to use Resource::obj_id's here.  Pointers are not the consistent from snapshot/init.  So you will also need to add another map from obj_id to resource_ptr.  The InitInv codegen for ResMap will need to populate the obj_id->ptr map while the InitFrom(db) codegen will populate the K->obj_id map.

rwcarlsen:
My only api concern is that it will be easy for users to "take resources out" without actually removing them from the map.  This could result in all sorts of difficult to debug problems.  ResBuf is careful to not allow users to see any of the contained resources without them actually being removed from the buffer.  The only exception is a single Peek function- which has a special name that helps make the distinction clear to the user and is not the "normal" way to use the buffer.

src/toolkit/res_map.h
rwcarlsen: "Pointers are not consistent across snapshot+init" is what I meant :)

scopatz:
Hi @rwcarlsen - I think that the API distinctions are warranted. This is supposed to be a generic container and implemented a std::map-esque interface.  It is not meant to be a buffer. Each key can be used however the user desires. Each item is supposed to be considered heterogeneously from every other item.  This is very different from a buffer, where every element is nominally "of the same kind" as every other element, and may naturally be combined with one another.  

I am not super worried because people know that when they access an element of a map, the element doesn't magically disappear from the map. If they truly want to remove the key they have to use the `erase()` member function.  That is the same behaviour here. 

Thanks for the point about `obj_ids`, etc.  I figured that could be handling in code gen, but I guess we'll see.

scopatz:
Also I am not opposed to adding a `Pop()` method in addition to what is here now.

rwcarlsen:
Pop would be nice - because it both erases and returns the stored value.
Code gen can't get around pointers being non-determinstic.  I almost missed
the issue myself while looking over ResMap.  Codegen will help us to
persist the ResBuf's internal state, but we still need to have a
cross-snapshot-init state to give it.

scopatz:
Ok, this is ready for another look.

rwcarlsen:
Looks good.  The last thing I noticed is that ResMap allows you to break its capacity because it doesn't recompute total quantity stored with the overloaded `operator[]`.  So I could:

```c++
resmap["foo"] = very_large_material_breaking_capacity_limit;
```

without any feedback.   This issue should either be addressed, or the concept of capacity could also be removed from ResMap and just keep quantity.

And a `int count()` function would be nice too.

scopatz:
> Looks good. The last thing I noticed is that ResMap allows you to break its capacity because it doesn't recompute total quantity stored with the overloaded operator[]. So I could:

```c++
resmap["foo"] = very_large_material_breaking_capacity_limit;
```

> without any feedback. This issue should either be addressed, or the concept of capacity could also be removed from ResMap and just keep quantity.

Not entirely true. An error will be thrown the next time they ask for the quantity().  I am not certain that the capacity has meaning in all cases for a ResMap. I only kept it in there because I thought that it was required to be a resource container. I'd be happy eliminating it.  Strict capacity checks and nice indexing are basically mutually exclusive in C++ because of the way lvalues work. The alternative is to reimplement a mapping algorithm from just to get capacity checks.

> And a int count() function would be nice too.

There is `size()`. I don't want to provide two APIs for the same idea.

rwcarlsen:
Oh - I didn't notice "size" - sorry.  Capacity could still be useful but
also confusing with the delayed error.  I'm sort of on the fence.  You can
decide.  If you prefer no capacity you are welcome to remove it.

scopatz:
> Oh - I didn't notice "size" - sorry. 

No worries. 

>  Capacity could still be useful but
also confusing with the delayed error.  I'm sort of on the fence.  You can
decide.  If you prefer no capacity you are welcome to remove it.

I think I'll remove it. Without loss of generality we can add it back in later. In some sense there are two capacities anyway: the total capacity and the capacity for each key. This really only addressed the first one, and poorly at that.

scopatz:
updated and removed capacities

scopatz:
Passes CI

scopatz:
I am a little disappointed that this has seen no movement in a whole week.

cli/cycpp.py
```diff
@@ -89,13 +89,16 @@
...
            ('{0}::toolkit::ResBuf'.format(CYCNS), CYCNS + '::Resource'),
            ('{0}::toolkit::ResBuf'.format(CYCNS), CYCNS + '::Product'),
            ('{0}::toolkit::ResBuf'.format(CYCNS), CYCNS + '::Material'),
+           '{0}::toolkit::ResMap'.format(CYCNS),
            }
 
 TEMPLATES = {'std::vector', 'std::set', 'std::list', 'std::pair',
-             'std::map', '{0}::toolkit::ResBuf'.format(CYCNS),}
+             'std::map', '{0}::toolkit::ResBuf'.format(CYCNS),
+             CYCNS + '::toolkit::ResMap',}
```
jdangerx: Is there a reason for this to be a `+` instead of a `.format()` ? Doesn't seem like a big deal but just wondering.

scopatz: It was just shorter and faster:

```python
In [10]: %timeit "Hello" + "::There"
10000000 loops, best of 3: 20.9 ns per loop

In [11]: %timeit "{0}::There".format("Hello")
1000000 loops, best of 3: 145 ns per loop
```

gidden: this dictionary (and others) seems to use the `format` method, perhaps there's an argument for consistency?

scopatz: Is this blocking?

gidden: Just a style comment, dove-tailing off @jdangerx. @rwcarlsen seems to have
the most skin in this game, perhaps he should do a final review and merge.

On Sat, Jan 24, 2015 at 12:56 PM, Anthony Scopatz <notifications@github.com>
wrote:

> In cli/cycpp.py
> <https://github.com/cyclus/cyclus/pull/1066#discussion_r23498478>:
>
> >             }
> >
> >  TEMPLATES = {'std::vector', 'std::set', 'std::list', 'std::pair',
> > -             'std::map', '{0}::toolkit::ResBuf'.format(CYCNS),}
> > +             'std::map', '{0}::toolkit::ResBuf'.format(CYCNS),
> > +             CYCNS + '::toolkit::ResMap',}
>
> Is this blocking?
>
> —
> <https://github.com/cyclus/cyclus/pull/1066/files#r23498478>.
>



-- 
Matthew Gidden
Ph. 225.892.3192

scopatz: I don't think that we should limit ourselves to just myself, @gidden, and @rwcarlsen as mergers. That is not a healthy community, especially when we have had consistent excessive delays in reviewing, let alone merging. I trust @gonuke, @katyhuff, @rakhimov, @mbmcgarry, @opotowsky, @jdangerx, and all other people with push access just as much as I trust myself.

gidden: (Sorry, that was not at all meant to deter @jdangerx from also
reviewing/merging)

On Sat, Jan 24, 2015 at 12:58 PM, Matthew Gidden <gidden@wisc.edu> wrote:

> Just a style comment, dove-tailing off @jdangerx. @rwcarlsen seems to have
> the most skin in this game, perhaps he should do a final review and merge.
>
> On Sat, Jan 24, 2015 at 12:56 PM, Anthony Scopatz <
> notifications@github.com> wrote:
>
>> In cli/cycpp.py
>> <https://github.com/cyclus/cyclus/pull/1066#discussion_r23498478>:
>>
>> >             }
>> >
>> >  TEMPLATES = {'std::vector', 'std::set', 'std::list', 'std::pair',
>> > -             'std::map', '{0}::toolkit::ResBuf'.format(CYCNS),}
>> > +             'std::map', '{0}::toolkit::ResBuf'.format(CYCNS),
>> > +             CYCNS + '::toolkit::ResMap',}
>>
>> Is this blocking?
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/cyclus/cyclus/pull/1066/files#r23498478>.
>>
>
>
>
> --
> Matthew Gidden
> Ph.D. Candidate, Nuclear Engineering
> The University of Wisconsin -- Madison
> Ph. 225.892.3192
>



-- 
Matthew Gidden
Ph. 225.892.3192

scopatz: Again, is this blocking.

jdangerx: Not really - I was just wondering why it was +, since as Matt pointed out it's format everywhere else.

cli/cycpp.py
```diff
@@ -1380,19 +1411,34 @@ def impl(self, ind="  "):
...
                 continue
-
             if self.pragmaname in info:
                 impl += info[self.pragmaname]
                 continue
+
+            # get expression
             t = info["type"]
+            key = t if isinstance(t, STRING_TYPES) else t[0]
             if t in BUFFERS:
```
jdangerx: Should this be `if key in BUFFERS:`? Seems to be trying to do something similar to line 1305.

jdangerx:
Seems good to me - a little consistency stuff in cycpp.py but that's not super important, in my opinion. Looks like  @rwcarlsen 's issues have been resolved, also, with the removal of capacity.

jdangerx:
Thanks, @scopatz !

PR #1067, opened 2015-01-09
Transmute resets prev_decay_time now

rwcarlsen:
Previously, Transmute left prev_decay_time alone resulting in strange/wrong
results if, for example, a material had accumulated several time steps of
delta t since last decay - and then was transmuted - say burned as reactor
fuel.  The resulting material would have still remembered its previous decay
delta t.  Causing bad physics.

This fix changes the behavior to assume that the Transmute caller has
calculated the new target composition to be accurate for the current
simulation time - i.e. no accumulated decay delta t.

### DIFF
--- a/src/material.cc
+++ b/src/material.cc
@@ -128,6 +128,17 @@ void Material::Absorb(Material::Ptr mat) {
 void Material::Transmute(Composition::Ptr c) {
   comp_ = c;
   tracker_.Modify();
+
+  // Presumably the user has chosen the new composition to be accurate for
+  // the current simulation time.  The next call to decay should not include
+  // accumulated decay delta t from a composition that no longer exists in
+  // this material.  This ---------+ condition allows testing to work.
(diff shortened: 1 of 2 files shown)


scopatz:
Seems reasonable, waiting on CI.

rwcarlsen:
Actually - also wait - I want to push up a test for this.

scopatz:
Also reasonable.

rwcarlsen:
test added.  This is built on #1054 now and so should be merged after it.

scopatz:
Seems reasonable. Waiting for #1054 to be merged.

rwcarlsen:
1054 is in and this has been rebased.  waiting on CI.

scopatz:
Looks like we lost the run id.

scopatz:
relaunching

scopatz:
Thanks

PR #1078, opened 2015-01-28
Much Improved Default Values

scopatz:
This PR fixes significant issues with the default value handling in cycpp.  This was discovered by trying to install bright-lite with the develop version of cyclus. Perhaps @FlanFlanagan would like to take a look.


scopatz:
That is review and merge.

PR #1079, opened 2015-02-05
JSON to XML, and Back Again

scopatz:
This PR implements a mechanism to help retain the last shred of my sanity by virtue of not having to look at thousands of lines of XML.  This implements converters between JSON and XML formats, following some common conventions [1,2]. As a sanity-retention device, I largely consider this a stop-gap to having Python input files.

This PR adds the command line options `--xml-to-json` and `--json-to-xml`. This also enables a simulation to be run from a properly formed JSON input file. During initialization, the JSON is converted into XML prior to the normal schema validation routines.

Lastly, this adds interleaves to schema elements that are order independent and should have been interleaved a while ago.

Note that over the course of this PR's development I found out that we hit the hard drive several times, by loading and reloading the same input file.  This should be cleaned up eventually.

1. https://github.com/Cheedoong/xml2json#xml2json
2. http://www.utilities-online.info/xmltojson/#.VNO1KnBGh5Q


scopatz:
Note that this has an ~2x LOC reduction:

```bash
$ wc -l *
  181 bl-two-stream-fr.json
  371 bl-two-stream-fr.xml
  334 forced-deploy.json
  729 forced-deploy.xml
  311 mktdriverinst-example.json
  522 mktdriverinst-example.xml
```

Also this passes CI.

mbmcgarry:
Well, I can't comment on the details of the implementation, but nothing jumps out as obviously problematic.  Sounds like it will be sanity-saving indeed.

scopatz:
Well, I'd be happy if you merged it :)

PR #1080, opened 2015-02-06
hotfix for reading in variable-length data

scopatz:
Minor implementation fix that enables reading in variable-length data in outside of a simulation context.  This is ridiculously urgent. Since it affects @opotowsky, she should probably review, though given the urgency any one is welcome to do the honors. Thanks.


opotowsky:
Hi @scopatz! Thank you for this, and I'm on it

opotowsky:
My fledgling understanding of c++ and batlab agree that this is merge ready

PR #1082, opened 2015-02-12
TimeSeries

scopatz:
This implements a time series convention helpers.

### DIFF
--- a/src/toolkit/timeseries.cc
+++ b/src/toolkit/timeseries.cc
@@ -1,0 +1,33 @@
+#include "error.h"
+
+namespace cyclus {
+namespace toolkit {
+
+void RecordTimeSeries(TimeSeriesTypes tstype, cyclus::Agent* agent,
+                      double value) {
+  using std::string;
+  string strtype;
+  switch (tstype) {
+    case POWER : {
+      strtype = "Power";
+      break;
+    }
+    default: {
+      throw cyclus::ValueError("TimeSeriesType not understood!");
+    }
+  }
+  RecordTimeSeries(strtype, agent, value);
+}
+
+void RecordTimeSeries(std::string tstype, cyclus::Agent* agent, double value) {
+  using std::string;
+  std::string tblname = "TimeSeries" + tstype;
+  agent->context()->NewDatum(tblname)
+       ->AddValue("AgentId", agent->id())
+       ->AddValue("Time", agent->context()->time())
+       ->AddValue("Value", value)
+       ->Record();
+}
+
+}  // namespace toolkit
+}  // namespace cyclus
--- a/src/toolkit/timeseries.cc
+++ b/src/toolkit/timeseries.cc
@@ -1,33 +1,10 @@
-#include "error.h"
-
 namespace cyclus {
 namespace toolkit {
 
-void RecordTimeSeries(TimeSeriesTypes tstype, cyclus::Agent* agent,
-                      double value) {
-  using std::string;
-  string strtype;
-  switch (tstype) {
-    case POWER : {
-      strtype = "Power";
-      break;
-    }
-    default: {
-      throw cyclus::ValueError("TimeSeriesType not understood!");
-    }
-  }
-  RecordTimeSeries(strtype, agent, value);
+void RecordTimeSeries<POWER>(cyclus::Agent* agent, double value) {
+  RecordTimeSeries<double>("Power", agent, value);
 }
 
-void RecordTimeSeries(std::string tstype, cyclus::Agent* agent, double value) {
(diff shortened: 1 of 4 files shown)


tests/toolkit/timeseries_tests.cc
```diff
@@ -0,0 +1,27 @@
...
+  RecordTimeSeries<POWER>(a, 42.0);
+  delete a;
+}
+
+TEST(TimeSeriesTests, RawPower) {
+  TestContext tc;
+  Agent* a = new TestAgent(tc.get());
+  RecordTimeSeries<double>("Power", a, 42.0);
+  delete a;
+}
```
mbmcgarry: Are you looking at the output database separately to see that the information was actually recorded?

src/toolkit/timeseries.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include <string>
+
+#include "agent.h"
+#include "context.h"
+
+namespace cyclus {
+namespace toolkit {
+
+/// Time series types to be used in the RecordTimeSeries() fucntions.
```
mbmcgarry: should be functions

tests/toolkit/timeseries_tests.cc
scopatz: No, this doesn't write out to a real database. It just implements the correct interface for testing.  This strategy is used in other parts of the test suite. We test database writing in other places as well.

src/toolkit/timeseries.h
scopatz: Thanks!

tests/toolkit/timeseries_tests.cc
mbmcgarry: Looks reasonable from my limited perspective.

scopatz:
I think this is good to go, but did anyone else want to take a look?

rwcarlsen:
I'd like to take a closer look tomorrow morning if you don't mind waiting
 until then.

scopatz:
I think that we'll be hitting the need for this tomorrow afternoon.  So before then would be great

src/toolkit/timeseries.cc
```diff
@@ -0,0 +1,13 @@
+#include "timeseries.h"
+
+namespace cyclus {
+namespace toolkit {
+
+template <>
+void RecordTimeSeries<POWER>(cyclus::Agent* agent, double value) {
+  RecordTimeSeries<double>("Power", agent, value);
```
rwcarlsen: Not a stopper, but FYI - you could just `RecordTimeSeries("Power", agent, value);` and the compiler will infer the templated type for the function.

src/toolkit/timeseries.h
```diff
@@ -0,0 +1,37 @@
...
+#include "agent.h"
+#include "context.h"
+
+namespace cyclus {
+namespace toolkit {
+
+/// Time series types to be used in the RecordTimeSeries() functions.
+/// These types have the following unit which *must* be adhered to strictly:
+/// - POWER [MWe]
+enum TimeSeriesTypes {
```
rwcarlsen: This should be singular - it makes the it feel more natural when defining/using variables of this type - e.g.

```c++
TimeSeriesTypes mytype; // this is awkward

TimeSeriesType mytype; // more natural
```

tests/toolkit/timeseries_tests.cc
rwcarlsen: While we shouldn't be (re)testing the database writing/querying here, I will note that these tests are not particularly useful - they test nothing about the functions except that they can be called without throwing exceptions or segfaulting - which IMO, is hardly better than having no tests at all w.r.t. how much tests can actually do.

src/toolkit/timeseries.cc
scopatz: "explicit is better than implicit," or so they tell me :)

src/toolkit/timeseries.h
scopatz: Yup you are right.

tests/toolkit/timeseries_tests.cc
scopatz: In my opinion, the existence of a valid test is always better than no test, irrespective of the depth of functionality it tests. Unit tests in particular, should aim to test as simple of a behavior as possible.  I think that it is also worth noting that these underlying functions don't actually do that much and aren't that big.  They exist to establish a convention, not implement a lot of behavior.  I think that this is actually a pretty decent example of a small function with small tests implemented via mocked behavior.

scopatz:
OK! I think this is ready. Thanks for the review.

PR #1083, opened 2015-02-12
fix segfault from double decommissioning

rwcarlsen:


scopatz:
Thanks!

scopatz:
Thanks @rwcarlsen

PR #1084, opened 2015-02-13
mock simulation now can reconstruct resources for you

rwcarlsen:


rwcarlsen:
wait a bit on this.

rwcarlsen:
ready now.

rwcarlsen:
I'd really like to get this in soon if someone has time for a review.  Thanks.

gidden:
@rwcarlsen and I were talking in the office. To summarize: by adding a QBackend to the signature of the load methods, all *required* `SimInit` state is taken away. Thus, there is an argument for also making them static functions.

rwcarlsen:
done.  also - we may want to make LoadComposition public in the future, but no need to yet.

gidden:
So these now cause an ABI test failure for me. However, API stability is maintained. That's fine by me, but I don't want to merge in a PR that breaks a test. Can @scopatz chime in here?

rwcarlsen:
@scopatz - how long is this going to have to wait?

scopatz:
Oh sorry, yeah. We shouldn't merge anything in which breaks a test. So this PR should either fix the test or wait until develop progress such that the test is fixed.

rwcarlsen:
This PR doesn't break a test that wasn't already broken.  Especially very broken with clang.

scopatz:
But this causes a new breakage, so it still introduces an error.

rwcarlsen:
ETA on the fix?

scopatz:
I'd rather discuss that on the mailing list thread if it is not going to happen as part of this PR.

rwcarlsen:
Why would it be a part of this PR? - it has nothing to do with this PR and you were going to do it.

scopatz:
Then why are we talking about it on this PR? And yes I offered to do it.

mbmcgarry:
I've reviewed the code and tested it.
Is this OK to merge now @scopatz ?

scopatz:
I'd personally like to try this on my machine, given all the tsuris about it.  I'll do that now.

scopatz:
This works for me now. +1 to merge

PR #1085, opened 2015-02-17
updates decommisioning behavior such that an agent with a lifetime of 1 ...

gidden:
...built at t_0 will also be decommissioned at t_0, such that it will experience one full time step

This adds to the agent API an `exit_time` function. `-1` is returned if the agent has an 'infinite' lifetime. Otherwise, the total number of time steps - 1 is returned, per the associated [discussion](https://groups.google.com/forum/#!topic/cyclus-dev/jt29aJGkIdM).

### DIFF
--- a/src/agent.h
+++ b/src/agent.h
@@ -381,12 +381,12 @@ class Agent : public StateWrangler, virtual public Ider {
 
   /// Returns the default time step at which this agent will exit the
   /// simulation (-1 if the agent has an infinite lifetime).
-  //
-  // The "- 1" is because deomissioning happens at the end of a time step.
-  // For a lifetime of 1, we expect an agent to go through only 1 entire
-  // time step.  In this case, the agent should be decommissioned on the
-  // same time step it was created.  So we need "- 1" for lifetime to have
-  // this correct meaning.
+  ///
+  /// Deomissioning happens at the end of a time step. With a lifetime of 1, we
+  /// expect an agent to go through only 1 entire time step. In this case, the
+  /// agent should be decommissioned on the same time step it was
+  /// created. Therefore, for agents with non-infinite lifetimes, the exit_time
+  /// will be the enter time plus its lifetime less 1.
   inline const int exit_time() const {
     if (lifetime() == -1)
       return -1;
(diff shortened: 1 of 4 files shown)


gidden:
Note that whichever PR is chosen (if any), @scopatz should definitely at least chime in, as he is able.

src/agent.h
```diff
@@ -379,6 +379,20 @@ class Agent : public StateWrangler, virtual public Ider {
   /// decommissioning (-1 if the agent has an infinite lifetime).
   inline const int lifetime() const { return lifetime_; }
 
+  /// Returns the default time step at which this agent will exit the
+  /// simulation (-1 if the agent has an infinite lifetime).
+  //
```
rwcarlsen: triple slashes all the way down.

rwcarlsen:
Could you update the source and sink agents in cyclus to use the exit_time now too?

rwcarlsen:
I like this better than my PR - I'll revert the commit in mine.

gidden:
done @rwcarlsen

src/agent.h
```diff
@@ -379,6 +379,18 @@ class Agent : public StateWrangler, virtual public Ider {
...
+  /// Returns the default time step at which this agent will exit the
+  /// simulation (-1 if the agent has an infinite lifetime).
+  ///
+  /// Deomissioning happens at the end of a time step. With a lifetime of 1, we
+  /// expect an agent to go through only 1 entire time step. In this case, the
+  /// agent should be decommissioned on the same time step it was
+  /// created. Therefore, for agents with non-infinite lifetimes, the exit_time
+  /// will be the enter time plus its lifetime less 1.
+  inline const int exit_time() const {
+    lifetime() == -1 ? return -1 : return enter_time_ + lifetime_ - 1;
```
rwcarlsen: this doesn't compile.  I'd prefer

```c++
if (lifetime_ < 0) {
  return -1
}
return enter_time_+lifetime_-1;
```

rwcarlsen: But at least fix compile error :-)

gidden:
I didn't mean to leave in the ternary operator. Good catch!

scopatz:
Please update the archetype developer's guide with this information. Thanks!

PR #1090, opened 2015-02-20
Fix feed qty

mbmcgarry:
Added a toolkit function NonFissileMultiplier that is the ratio of total fuel over fissile material.  This is used when calculating necessary feed quantities for input materials into the enrichment facility that include components beyond U-235 and U-238.  Needed for the new cycamore EnrichmentFacility.

Also updated documentation for UraniumAssay and FeedQty to clarify that they assume the material is entirely comprised of fissile isotopes of uranium.

### DIFF
--- a/src/toolkit/enrichment.cc
+++ b/src/toolkit/enrichment.cc
@@ -75,6 +75,12 @@ double TailsQty(double product_qty, const Assays& assays) {
   return product_qty * factor;
 }
 
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+double NonFissileMultiplier(cyclus::Material::Ptr matl) {
+  cyclus::toolkit::MatQuery mq(matl);
+  return mq.qty()/(mq.mass(922350000) + mq.mass(922380000));
+}
+
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 double ValueFunc(double frac) {
   if (frac < 0) {
--- a/src/toolkit/enrichment.h
+++ b/src/toolkit/enrichment.h
@@ -31,7 +31,7 @@ class Assays {
 };
 
 /// @param mat the material inquired about
-/// @return the atom percent of U-235 w.r.t Uranium in a material
+/// @return the atom percent of U-235 w.r.t U-235+U-238 in a material
 double UraniumAssay(Material::Ptr mat);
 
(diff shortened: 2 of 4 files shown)


rwcarlsen:
For this to be general enough for the toolkit, I'd want it to return something like the fraction of the material that is fissile - including all fissile isotopes.

src/toolkit/enrichment.cc
```diff
@@ -75,6 +75,12 @@ double TailsQty(double product_qty, const Assays& assays) {
   return product_qty * factor;
 }
 
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+double NonFissileMultiplier(cyclus::Material::Ptr matl) {
+  cyclus::toolkit::MatQuery mq(matl);
+  return mq.qty()/(mq.mass(922350000) + mq.mass(922380000));
```
gidden: U-238 is not a fissile isotope.

src/toolkit/enrichment.h
```diff
@@ -57,6 +58,14 @@ double FeedQty(double product_qty, const Assays& assays);
 /// whose units match those of the given product
 double TailsQty(double product_qty, const Assays& assays);
 
+/// @param feed_matl pointer to material
+/// @return ratio of (total mass)/(fissile mass)
+/// Can be used to convert amnt of necessary fissile material
+/// (from FeedQty) into necessary amount of raw material when
+/// non-fissile components are present.  Returns 1 when material
+/// is entirely comprised of U-235 + U-238
+double NonFissileMultiplier(cyclus::Material::Ptr matl);
```
gidden: I don't think this function is doing what it's documentation states. Perhaps we can discuss its purpose and decide the best course of action (rename the function, etc.).

scopatz: I agree that this needs some sort of rename, or redefinition.

rwcarlsen:
This needs to be reverted.

rwcarlsen:
It got piggy'd on #1092 which had it in its history.

mbmcgarry:
Sorry, I didn’t mean to link the two PRs.  Yes, this should be reverted and I’ll come up with a different solution to this problem.

-Meghan


On Feb 26, 2015, at 9:58 PM, Anthony Scopatz <notifications@github.com<mailto:notifications@github.com>> wrote:


It got piggy'd on #1092<https://github.com/cyclus/cyclus/pull/1092> which had it in its history.

PR #1092, opened 2015-02-26
Xml annotation

mbmcgarry:
- Fixed bug that prevented xml annotations for schema to be put in header file. 
- updated documentation for enrichment toolkit functions to clarify which U isotopes are assumed in calculations


scopatz:
Thanks for taking care of this @mbmcgarry!

PR #1094, opened 2015-03-02
re-amalgamates pyne

scopatz:
Should fix or help fix the nightlies.


scopatz:
This likely needs a new nuc data to be uploaded, which shouldn't be done until this is merged.

PR #1096, opened 2015-03-06
Use boost serialization to simplify sqlite backend dramatically.

rwcarlsen:
Not only is the code much more simple and less error prone, but it is much more trivial to add new nested template types.  And it is now very easy to inspect container variable values in the output db by hand.


gidden:
this is definitely much simpler! do you know what effects the new strategy has on performance?

gidden:
@rwcarlsen and I discussed this in the office. His intuition is that this will make snapshotting/restart somewhat slower, but that our I/O is dominated by composition recording during a simulation.

scopatz:
This also added a new type to the type system.  Documentation is needed.

PR #1098, opened 2015-03-10
recursive container support for cycpp

rwcarlsen:
Also cleaned up xml schema generation.  Now it uses xml.dom.minidom to prettify and auto-escapes quotes and backslashes (e.g. for newlines).  This results in custom-schema snippets being much easier for archetype devs to write (and also makes the cyclus.github.com docs about them un-incorrect).

Full disclosure - there are some lines much longer than 80 chars - most/all of these are in the tests and are long because they are 'generated' snippets from the filters used as hybrid regression/unit tests (the unit test factor comes from the fact that I carefully verified the generated snippets to be correct).  In case we ever tweak the filters in the future, we can easily regenerate the changed "regression" snippets and verify correctness.  This is why I don't think it prudent to make them shorter.

### DIFF
--- a/cli/cycpp.py
+++ b/cli/cycpp.py
@@ -55,6 +55,7 @@ from argparse import ArgumentParser, RawDescriptionHelpFormatter
 from pprint import pprint, pformat
 import textwrap
 import difflib
+import xml.dom.minidom
 
 try:
     import simplejson as json
@@ -111,6 +112,28 @@ WRANGLERS = {
 
 ENTITIES = [('cyclus::Region', 'region'), ('cyclus::Institution', 'institution'), 
             ('cyclus::Facility', 'facility'), ('cyclus::Agent', 'archetype')]
+
+def escape_xml(s, ind='    '):
+    """Escapes xml string s, prettifies it and puts in c++ string lit form."""
+
+    s = xml.dom.minidom.parseString(s)
+    s = s.toprettyxml()
+    s = s.replace('"', '\\"')
+
+    lines = s.splitlines()
+    lines = lines[1:] # remove initial xml version tag
+
+    clean = []
+    for line in lines:
+        if line.strip() != '':
+            clean.append(line)
+
+    for i, line in enumerate(clean):
+        if line.strip() == '':
+            continue
+        clean[i] = '{0}"{1}\\n"'.format(ind, line.rstrip())
+
+    return '\n'.join(clean)
 
 def prepare_type(cpptype, othertype):
     """Updates othertype to conform to the length of cpptype using None's.
@@ -1203,7 +1226,12 @@ class InfileToDbFilter(CodeGeneratorFilter):
         alias = prepare_type(t, alias)
         if alias[1] == None:
             alias[1] = 'val'
-        s = '{0}::InfileTree* sub = sub->SubTree("{1}");\n'.format(CYCNS, alias[0])
+
+        # the extra assignment (bub, sub) is because we want the intial sub
+        # rhs to be from outer scope - otherwise the newly defined sub will be
+        # in scope causing segfaults
+        s = '{0}::InfileTree* bub = sub->SubTree("{1}");\n'.format(CYCNS, alias[0])
+        s += '{0}::InfileTree* sub = bub;\n'.format(CYCNS)
         s += 'int n = sub->NMatches("{0}");\n'.format(alias[1])
         s += ind + '{0} {1};\n'.format(type_to_str(t), member)
         s += ind + '{0}.resize(n);\n'.format(member)
@@ -1218,7 +1246,11 @@ class InfileToDbFilter(CodeGeneratorFilter):
         alias = prepare_type(t, alias)
         if alias[1] == None:
             alias[1] = 'val'
-        s = '{0}::InfileTree* sub = sub->SubTree("{1}");\n'.format(CYCNS, alias[0])
+        # the extra assignment (bub, sub) is because we want the intial sub
+        # rhs to be from outer scope - otherwise the newly defined sub will be
+        # in scope causing segfaults
+        s = '{0}::InfileTree* bub = sub->SubTree("{1}");\n'.format(CYCNS, alias[0])
+        s += '{0}::InfileTree* sub = bub;\n'.format(CYCNS)
         s += 'int n = sub->NMatches("{0}");\n'.format(alias[1])
         s += ind + '{0} {1};\n'.format(type_to_str(t), member)
         s += ind + 'for (i = 0; i < n; ++i) {\n'
@@ -1232,7 +1264,11 @@ class InfileToDbFilter(CodeGeneratorFilter):
         alias = prepare_type(t, alias)
         if alias[1] == None:
             alias[1] = 'val'
-        s = '{0}::InfileTree* sub = sub->SubTree("{1}");\n'.format(CYCNS, alias[0])
+        # the extra assignment (bub, sub) is because we want the intial sub
+        # rhs to be from outer scope - otherwise the newly defined sub will be
+        # in scope causing segfaults
+        s = '{0}::InfileTree* bub = sub->SubTree("{1}");\n'.format(CYCNS, alias[0])
+        s += '{0}::InfileTree* sub = bub;\n'.format(CYCNS)
         s += 'int n = sub->NMatches("{0}");\n'.format(alias[1])
         s += ind + '{0} {1};\n'.format(type_to_str(t), member)
         s += ind + 'for (i = 0; i < n; ++i) {\n'
@@ -1248,7 +1284,11 @@ class InfileToDbFilter(CodeGeneratorFilter):
             alias[1] = 'first'
         if alias[2] == None:
             alias[2] = 'second'
-        s = '{0}::InfileTree* sub = sub->SubTree("{1}");\n'.format(CYCNS, alias[0])
+        # the extra assignment (bub, sub) is because we want the intial sub
+        # rhs to be from outer scope - otherwise the newly defined sub will be
+        # in scope causing segfaults
+        s = '{0}::InfileTree* bub = sub->SubTree("{1}");\n'.format(CYCNS, alias[0])
+        s += '{0}::InfileTree* sub = bub;\n'.format(CYCNS)
         s += ind + self.read_member('first', alias[1], t[1], uitype[1], ind+'  ')
         s += ind + self.read_member('second', alias[2], t[2], uitype[2], ind+'  ')
         s += ind + '{0} {1}(first, second);\n'.format(type_to_str(t), member)
@@ -1261,7 +1301,11 @@ class InfileToDbFilter(CodeGeneratorFilter):
             alias[1] = 'key'
         if alias[2] == None:
             alias[2] = 'val'
-        s = '{0}::InfileTree* sub = sub->SubTree("{1}");\n'.format(CYCNS, alias[0])
+        # the extra assignment (bub, sub) is because we want the intial sub
+        # rhs to be from outer scope - otherwise the newly defined sub will be
+        # in scope causing segfaults
+        s = '{0}::InfileTree* bub = sub->SubTree("{1}");\n'.format(CYCNS, alias[0])
+        s += '{0}::InfileTree* sub = bub;\n'.format(CYCNS)
         s += 'int n = sub->NMatches("{0}");\n'.format(alias[1])
         s += ind + '{0} {1};\n'.format(type_to_str(t), member)
         s += ind + 'for (i = 0; i < n; ++i) {\n'
@@ -1286,8 +1330,7 @@ class InfileToDbFilter(CodeGeneratorFilter):
         impl += self.shapes_impl(ctx, ind)
 
         # read data from infile onto class
-        impl += ind + 'tree = tree->SubTree("config/*");\n'
-        impl += ind + '{0}::InfileTree* sub = tree;\n'.format(CYCNS)
+        impl += ind + '{0}::InfileTree* sub = tree->SubTree("config/*");\n'.format(CYCNS)
         impl += ind + 'int i;\n'
         impl += ind + 'int n;\n'
         for member, info in ctx.items():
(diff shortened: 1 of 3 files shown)


mbmcgarry:
The XML schema cleanup portion is working well with the custom schema for the Cycamore Enrichment Facility.  Nice modification @rwcarlsen  !

rwcarlsen:
Good to hear :)

cli/cycpp.py
```diff
@@ -112,6 +113,44 @@
...
+    """Updates othertype to conform to the length of cpptype using None's.
+    """
+    if not isinstance(cpptype, STRING_TYPES):
+        if isinstance(othertype, STRING_TYPES):
+            othertype = [othertype]
+
+        if othertype is None:
+            othertype = []
+            for v in cpptype:
+                othertype.append(None)
```
scopatz: This should be `othertype = [None] * len(cpptype)`

cli/cycpp.py
```diff
@@ -112,6 +113,44 @@
...
+        elif len(othertype) < len(cpptype):
+            for i in range(len(cpptype)-len(othertype)):
+                othertype.append(None)
```
scopatz: same idea here.

cli/cycpp.py
```diff
@@ -1045,224 +1084,236 @@ def _val(self, t, val=None, name=None, uitype=None):
...
-    def _val_uuid(self, t, val='', name=None, uitype=None):
+    def _val_uuid(self, t, val='', name='uuidvar', uitype=None):
+        v = 'boost::uuids::uuid {0} = '.format(name)
         if isinstance(val, STRING_TYPES):
-            v = '"{0}"'.format(val)
+            v += '"{0}"'.format(val)
         elif isinstance(val, uuid.UUID):
-            v = [x+y for x,y in zip(val.hex[::1], val.hex[1::2])]
-            v = '{0x' + ', 0x'.join(v) + '}'
+            l = [x+y for x,y in zip(val.hex[::1], val.hex[1::2])]
```
scopatz: PEP8: space after commas

cli/cycpp.py
```diff
@@ -1045,224 +1084,236 @@ def _val(self, t, val=None, name=None, uitype=None):
...
+
         return v
 
-    def _val_set(self, t, val=frozenset(), name=None, uitype=None):
-        vtype = t[1]
-        vuitype = None if uitype is None else uitype[1]
-        v = ''
+    def _val_set(self, t, val=frozenset(), name='setvar', uitype=None):
+        uitype = prepare_type(t, uitype)
+        elemname = 'elem'
```
scopatz: Didn't this used to be val?  Won't it break a lot of input files to change now?

cli/cycpp.py
```diff
@@ -1045,224 +1084,236 @@ def _val(self, t, val=None, name=None, uitype=None):
...
         if uitype == 'nuclide':
-            template = ('pyne::nucname::id({cycns}::{query}<std::string>({tree}, '
-                        '"{alias}"{default}{index}))')
-            if d is not None:
-                kw['default'] = ', "{0}"'.format(d)
+            template = ('pyne::nucname::id({cycns}::Query<std::string>({tree}, '
+                        '"{alias}"{index}))')
         else:
-            template = '{cycns}::{query}<{type}>({tree}, "{alias}"{default}{index})'
+            template = '{cycns}::Query<{type}>({tree}, "{alias}"{index})'
```
scopatz: Why remove the optional query? Is there something I am not seeing?

cli/cycpp.py
```diff
@@ -1382,18 +1458,81 @@ def _type(self, cpp, given=None):
...
+                name = names[0]
+
+            keynames = 'key' if isinstance(cpptype[1], STRING_TYPES) else ['key']
+            if names[1] is not None:
+                keynames = names[1]
+
+            valnames = 'val' if isinstance(cpptype[2], STRING_TYPES) else ['val']
+            if names[1] is not None:
+                valnames = names[2]
+
```
scopatz: PEP8 excessive vertical whitespace

cli/cycpp.py
rwcarlsen: This is in InfileToDb - and this variable name is referring to a temporary c++ var - nothing to do with schema tag names.

cli/cycpp.py
rwcarlsen: If the statevar is optional, the top level impl, now catches this in an if block.  It checks to see if the variable was defined in the schema before running this cpp code.  So This code only ever runs after we know that the optional value has *already* been defined in the input file.  So it was actually a redundant check before.  We don't need the default, because that is constructed and assigned by code that gets inserted into the `else` of the if block that checks to see if the value was specified in the input file.

cli/cycpp.py
rwcarlsen: fixed

cli/cycpp.py
rwcarlsen: done

cli/cycpp.py
rwcarlsen: done

cli/cycpp.py
rwcarlsen: done.

rwcarlsen:
ready.

cli/cycpp.py
scopatz: OK

rwcarlsen:
Thanks for the review @scopatz - good questions.

cli/cycpp.py
```diff
@@ -1305,8 +1352,34 @@ def impl(self, ind="  "):
...
+                labels = info.get('alias', None)
+                if labels is None:
+                    labels = member if isinstance(t, STRING_TYPES) else [member]
                 reader = self.readers.get(t, self.readers.get(t[0], None))
-                impl += reader(member, alias, t, d, uitype, ind)
+
+                # generate condition to choose default vs given val if default exists
+                if d is not None:
+                    name = labels if isinstance(t, STRING_TYPES) else labels[0]
+                    impl += ind + 'if (sub->NMatches("{0}") > 0) {{\n'.format(name)
```
scopatz: Is this what checks that the statevar is in the input file?

rwcarlsen: Yes.

scopatz: So this part of the code - or its body - does not produce properly indented, human readable code.  I just opened up the generated file and I did not find it easy to read or understand.  From sink.h, here is what I am seeing:

```C++
  virtual void InfileToDb(cyclus::InfileTree* tree, cyclus::DbInit di) {
    cyclus::Facility::InfileToDb(tree, di);
    cyclus::InfileTree* sub = tree->SubTree("config/*");
    int i;
    int n;
    if (sub->NMatches("recipe_name") > 0) {
{      std::string recipe_name_val = cyclus::Query<std::string>(sub, "recipe_name");
recipe_name = recipe_name_val;}    } else {
      std::string recipe_name_tmp("");recipe_name = recipe_name_tmp;    }
{cyclus::InfileTree* bub = sub->SubTree("in_commods");
cyclus::InfileTree* sub = bub;
int n = sub->NMatches("val");
    std::vector< std::string > in_commods_val;
    in_commods_val.resize(n);
    for (i = 0; i < n; ++i) {
    std::string elem;{        std::string elem_in = cyclus::Query<std::string>(sub, "val", i);
elem = elem_in;}      in_commods_val[i] = elem;
    }
in_commods = in_commods_val;}{    double capacity_val = cyclus::Query<double>(sub, "capacity");
capacity = capacity_val;}    if (sub->NMatches("max_inv_size") > 0) {
{      double max_inv_size_val = cyclus::Query<double>(sub, "max_inv_size");
max_inv_size = max_inv_size_val;}    } else {
      double max_inv_size_tmp = 1e+299;max_inv_size = max_inv_size_tmp;    }
    di.NewDatum("Info")
    ->AddVal("recipe_name", recipe_name)
    ->AddVal("in_commods", in_commods)
    ->AddVal("capacity", capacity)
    ->AddVal("max_inv_size", max_inv_size)
    ->Record();
  };
```

scopatz:
np. happy to look at it since it is a pretty major refactor of one of my refactors :)

cli/cycpp.py
rwcarlsen: Fixed.  Although I think it reasonable to assume people debugging codegen'd code just use something like astyle or clang-format.  We shouldn't invest too much worrying about these solved problems.

gidden: Separately, *we* could clang-format code-gen'd code as a post-process (if
it exists on the system). Just a thought..

On Fri, Mar 13, 2015 at 9:59 AM, Robert Carlsen <notifications@github.com>
wrote:

> In cli/cycpp.py
> <https://github.com/cyclus/cyclus/pull/1098#discussion_r26392184>:
>
> >                  reader = self.readers.get(t, self.readers.get(t[0], None))
> > -                impl += reader(member, alias, t, d, uitype, ind)
> > +
> > +                # generate condition to choose default vs given val if default exists
> > +                if d is not None:
> > +                    name = labels if isinstance(t, STRING_TYPES) else labels[0]
> > +                    impl += ind + 'if (sub->NMatches("{0}") > 0) {{\n'.format(name)
>
> Fixed. Although I think it reasonable to assume people debugging codegen'd
> code just use something like astyle or clang-format. We shouldn't invest
> too much worrying about these solved problems.
>
> —
> <https://github.com/cyclus/cyclus/pull/1098/files#r26392184>.
>



-- 
Matthew Gidden
Ph. 225.892.3192

scopatz: The last thing that a human who is debugging wants to do is run another piece of software :)

scopatz: Also it looses all value as a teaching tool if you have to run a formatted on in.

rwcarlsen: I understand the viewpoint.  But the people 'debugging' codegen'd code are most likely cyclus core developers - the same people maintaining the code generator.  There is a non-trivial cost in code-generator readability, ease of maintainability, etc. No free lunch.  I think the cost is higher than the benefit in this case.  But as I said - fixed - just wanted to  make my opinion known :-)

rwcarlsen:
CI passes... but should it?   ;-)

scopatz:
Is this ready to look at or are you still working on it @rwcarlsen?

rwcarlsen:
ready.

tests/cycpp_tests.py
```diff
@@ -581,24 +843,25 @@ def test_nuclide_uitype():
...
-                '  x = pyne::nucname::id(cyclus::Query<std::string>(tree, "x"));\n'
-                '  di.NewDatum("Info")\n'
-                '  ->AddVal("x", x)\n'
-                '  ->Record();\n')
+    exp_impl = (
+        '  cyclus::InfileTree* sub = tree->SubTree("config/*");\n'
+        '  int i;\n'
+        '  int n;\n'
+        '{  int x_val = pyne::nucname::id(cyclus::Query<std::string>(sub, "x"));\n'
+        'x = x_val;}  di.NewDatum("Info")\n'
```
scopatz: This does not seem properly formatted.  I was seeing the same thing in the output of the sink.h

rwcarlsen:
ping - this should be for sure ready now :-)

cli/cycpp.py
```diff
@@ -112,6 +113,41 @@
...
 
+def escape_xml(s, ind='    '):
+    """Escapes xml string s, prettifies it and puts in c++ string lit form."""
+
+    s = xml.dom.minidom.parseString(s)
+    s = s.toprettyxml(indent='    ')
+    s = s.replace('"', '\\"')
+
+    lines = s.splitlines()
+    lines = lines[1:] # remove initial xml version tag
```
scopatz: PEP8 two spaces before comment.

cli/cycpp.py
```diff
@@ -112,6 +113,41 @@
...
+
+    clean = []
+    for line in lines:
+        if line.strip() != '':
+            clean.append(line)
```
scopatz: This should be a list comprehension and the predicate should be `len(line.strip()) == 0`

cli/cycpp.py
```diff
@@ -112,6 +113,41 @@
...
+
+    for i, line in enumerate(clean):
+        clean[i] = '{0}"{1}\\n"'.format(ind, line.rstrip())
```
scopatz: Again a list comprehension or a `map()` would do well here.

tests/cycpp_tests.py
```diff
@@ -390,74 +393,162 @@ def test_itdbfilter():
...
-                '  int n;\n'
-                '  x = cyclus::Query<int>(tree, "x");\n'
-                'THINGFISH\n'
-                '  di.NewDatum("Info")\n'
-                '  ->AddVal("x", x)\n'
-                'ABSOLUTELY FREE\n'
-                '  ->Record();\n')
-    yield assert_equal, exp_impl, impl
+    exp_impl = (
+        '  int rawcycpp_shape_y[1] = {42};\n  cycpp_shape_y = std::vector<int>(rawcycpp_shape_y, rawcycpp_shape_y + 1);\n  cyclus::InfileTree* sub = tree->SubTree("config/*");\n  int i;\n  int n;\n  {\n    int x_val = cyclus::Query<int>(sub, "x");\n    x = x_val;\n  }\nTHINGFISH\n  di.NewDatum("Info")\n  ->AddVal("x", x)\nABSOLUTELY FREE\n  ->Record();\n'
```
scopatz: This statement is too long for one line and it is on three line.  You may as well split up the string since you already have the grouping parentheses.

scopatz:
The output looks much better. Thanks

scopatz:
Just a few more minor issues inline.

tests/cycpp_tests.py
rwcarlsen: since this was a pprint-generated string, it seems a bit counterproductive to modify it.  Some versions of pprint automagically do line breaks, the version I generated this with does not.  I'd rather leave this particular line alone.

cli/cycpp.py
rwcarlsen: done

cli/cycpp.py
rwcarlsen: done

cli/cycpp.py
rwcarlsen: done

tests/cycpp_tests.py
scopatz: fine.

scopatz:
Is this ready?

scopatz:
for another look?

rwcarlsen:
should be.

rwcarlsen:
rebased

rwcarlsen:
Thanks for the review @scopatz.

PR #1101, opened 2015-03-12
Agent::lifetime_ now protected to support deploy inst lifetime setting.

rwcarlsen:
Also - smarter decom handling in inst class - now supports non-facility subclass children.


gidden:
looks fine to me

gidden:
this passes for me

PR #1102, opened 2015-03-12
protected access is not enough - need a setter for Agent::lifetime

rwcarlsen:
to enable other agents to manually specify lifetimes.


scopatz:
This could use a unit test.

rwcarlsen:
done.

PR #1103, opened 2015-03-13
added map<string, map<int, double> > and corresponding VL permutations

rwcarlsen:


gidden:
Mmm, I know that moving the case/switch statement into the macro was done in an effort to reduce vertical space, etc., but I think here it obfuscates what's actually going on. My vote would be to move them out of the macro, but I'd happily be outvoted.

rwcarlsen:
My reasoning was that the macro is defined in the immediate vicinity of its
use.  And this makes it less likely that updates with new types will have
problems.  New updates/additions only require one line that is very
difficult to get wrong - easier to get right than before IMO.  If the macro
were defined elsewhere, I would absolutely agree that obfuscation becomes
problematic.

gidden:
ok

scopatz:
In any event, the macro names are probably insufficiently scary http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Macro_Names

scopatz:
That is: 

> Do not just #undef an existing macro before replacing it with your own; instead, pick a name that's likely to be unique.

http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Preprocessor_Macros

rwcarlsen:
I didn't undef macros before replacing them.  I undef'd my own macros after
using them.  I actually went out of my way to follow the style guide.  And
I think the names are fine.  The style guide never says macro names should
be scary.

scopatz:
Yes, the style guide does.  It says:

> If you do, they're like this: MY_MACRO_THAT_SCARES_SMALL_CHILDREN.

 http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Macro_Names

scopatz:
At the very least, I think that `CYCLUS_` should be prepended.

rwcarlsen:
That doesn't say to give them scary names.  It just gives one scary example
for humor.  We certainly can't name all macros with that exact name.  A
prefix is not necessary because the macros are immediately un-defined in
the same file (in the same function even) after their usage.

scopatz:
I disagree that the comment was just for humor.  It was just using humor to relate an idea. The problem is that if you define a macro that overwrites a previously defined macro, undefining will delete both.  Namespacing (the concept, not the syntax) prevents that. Furthermore, the macro names here are very generic (COMMA, LOADVAR) and so namespacing becomes even more important.  This avoids an unlikely, extremely painful, and 100% preventable class of bugs.

rwcarlsen:
CYCLUS_ prefix added.

gonuke:
FWIW, I do not think that the style guide is advocating scary names at all!

    You're not really going to define a macro, are you? If you do, they're like this: MY_MACRO_THAT_SCARES_SMALL_CHILDREN.

Careful parsing of this suggests that all macros are interpreted as scary and should thus be avoided, if possible, rather than that all macros should be given scary names.  The examples that immediately follow this are more evidence to this interpretation.

I do think namespacing is important, so I'm glad to see those added.

scopatz:
This failed CI.  Relaunching.

scopatz:
Failed again.  I think that this is a conda-build update which broke our selectors.  I'll try to fix.

src/query_backend.h
```diff
@@ -165,8 +165,28 @@ enum DbTypes {
...
   // append new types only:
+
+  // map<string, vector<double> >
+  MAP_STRING_VECTOR_DOUBLE,
+  MAP_STRING_VL_VECTOR_DOUBLE,
+  VL_MAP_STRING_VECTOR_DOUBLE,
+  MAP_VL_STRING_VECTOR_DOUBLE,
+  MAP_VL_STRING_VL_VECTOR_DOUBLE,
+  VL_MAP_STRING_VL_VECTOR_DOUBLE,
+  VL_MAP_VL_STRING_VECTOR_DOUBLE,
```
scopatz: You can only append to this Enum.  Otherwise the dbtype system breaks.

scopatz:
Also note that the dbtypes.json file in the docs needs to have a corresponding PR before this can be merged. Thanks!

src/query_backend.h
rwcarlsen: But this is entirely inter-release.  I think this is okay.  We never promise to be api compatible on develop branch - we promise to on release to release.

rwcarlsen: s/inter/intra/

scopatz: You are right about compatability.  Can you point me to the commit where `VL_MAP_VL_STRING_VL_VECTOR_DOUBLE` was first added?

scopatz: Please.

rwcarlsen:
Can we wait until we know all the dbtypes for the entire release? - It is not unlikely there will be more before then as cycamore gets finished up.

scopatz:
Not really.  The dbtypes.json file is an interface for cymetric.  So it is needed to for cymetric development to be tested with the new types that have been added.  Obviously the pivot table should remain defaulted to v1.2, but the file itself needs to be updated.

src/query_backend.h
rwcarlsen: https://github.com/cyclus/cyclus/pull/1096

scopatz: OK

rwcarlsen:
dbtypes PR in cyclus/cyclus.github.com#126

scopatz:
Looks good locally.  Relaunching.

scopatz:
Thanks @rwcarlsen

PR #1105, opened 2015-03-18
added -D to install.py

scopatz:
Useful for passing definitions directly down to CMake. Taken from pyne.


gonuke:
I've tested two null cases successfully:
* leaving out the -D doesn't cause problems
* passing a variable that is irrelevant (-DPPHW_FOO=BAR) doesn't cause problems.

Any suggestions for a non-NULL test, even one that just sets a variable to some otherwise default value?

scopatz:
`-DCMAKE_BUILD_TYPE=Release` should work

gonuke:
Looks good to me.

PR #1106, opened 2015-03-20
LTEQ Demand Constraint

gidden:
Hi folks,

This PR adds the ability for consumers to supply a LTEQ constraint. For example, a reactor can now make a demand for fuel with total fission products less than a threshold. Support has been added for both the Greedy and mathematical programming solvers.  

I added basic unit tests, a "trivial" integration test (integrating exchange layer to formulation layer and back), and a less trivial integration test.

### DIFF
--- a/src/capacity_constraint.h
+++ b/src/capacity_constraint.h
@@ -68,6 +68,7 @@ class CapacityConstraint {
   CapacityConstraint(double capacity, typename Converter<T>::Ptr converter)
       : capacity_(capacity),
         converter_(converter),
+        cap_type_(NONE),
         id_(next_id_++) {
     if (capacity_ <= 0)
       throw ValueError("Capacity is not positive, no trades will be executed");
@@ -77,6 +78,7 @@ class CapacityConstraint {
   /// that simply returns 1)
   explicit CapacityConstraint(double capacity)
       : capacity_(capacity),
+        cap_type_(NONE),
         id_(next_id_++) {
     if (capacity_ <= 0)
       throw ValueError("Capacity is not positive, no trades will be executed");
@@ -87,13 +89,23 @@ class CapacityConstraint {
   CapacityConstraint(const CapacityConstraint& other)
       : capacity_(other.capacity_),
         converter_(other.converter_),
-        id_(next_id_++) {}
-
-  /// @return the constraints capacity
-  inline double capacity() const {
-    return capacity_;
+        cap_type_(other.cap_type_),
+        id_(next_id_++) {
+    assert(capacity_ > 0);
   }
 
+  /// capacity getters/setters
+  // @{
+  inline double capacity() const { return capacity_; }
+  inline void capacity(double c) { capacity_ = c; }
+  // @}
+  
+  /// type getters/setters
+  // @{
+  inline cap_t cap_type() const { return cap_type_; }
+  inline void cap_type(cap_t t) const { cap_type_ = t; }
+  // @}
+  
   /// @return the converter
   inline typename Converter<T>::Ptr converter() const {
     return converter_;
@@ -113,6 +125,7 @@ class CapacityConstraint {
 
  private:
   double capacity_;
+  cap_t cap_type_;
   typename Converter<T>::Ptr converter_;
   int id_;
   static int next_id_;
@@ -125,7 +138,8 @@ template<class T>
 inline bool operator==(const CapacityConstraint<T>& lhs,
                        const CapacityConstraint<T>& rhs) {
   return  ((lhs.capacity() == rhs.capacity()) &&
-           (*lhs.converter() == *rhs.converter()));
+           (*lhs.converter() == *rhs.converter()) &&
+           (lhs.cap_type() == rhs.cap_type()));
 }
 
 /// @brief CapacityConstraint-CapacityConstraint comparison operator, allows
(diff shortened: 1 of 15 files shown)


src/capacity_constraint.h
```diff
@@ -63,36 +64,69 @@ struct TrivialConverter : public Converter<T> {
 template <class T>
 class CapacityConstraint {
  public:
-  /// @brief constructor for a constraint with a non-trivial converter
+  /// constructor
+  /// @param capacity a capacitating value
+  /// @param type the type, e.g., LTEQ or GTEQ
+  /// @param converter a conversion function pointer
+  // @{
```
scopatz: triple slash?

src/capacity_types.h
```diff
@@ -0,0 +1,16 @@
+#ifndef CYCLUS_SRC_CAPACITY_TYPES_H_
+#define CYCLUS_SRC_CAPACITY_TYPES_H_
+
+namespace cyclus {
+
+/// the type of capacity, currently less-than-or-equal-to (LTEQ) and
+/// greater-than-or-equal-to (GTEQ) are supported
+enum CapType {
+  LTEQ = 0,
+  GTEQ,
```
scopatz: I think that the standard abbreviations are GE and LE.  That would be more consistent with the same names they are given in the query filters.

src/exchange_graph.h
```diff
@@ -135,6 +137,10 @@ class ExchangeNodeGroup {
   const std::vector<double>& capacities() const { return capacities_; }
   std::vector<double>& capacities() { return capacities_; }
 
+  /// @brief the kind of capacities (i.e., LTEQ, GTEQ, EQ)
```
scopatz: EQ doesn't seem to be a valid kind.  NONE perhaps?

src/exchange_translator.h
```diff
@@ -134,6 +134,25 @@ inline void AddBid(ExchangeTranslationContext<T>& translation_ctx,
...
+  for (c_it = s.begin(); c_it != s.end(); ++c_it) {
+    cap_t type = c_it->cap_type(); 
+    switch(type) {
+      case NONE:
+        // use default value for backwards compatability
+        g->AddCapacity(c_it->capacity()); 
+        break;
+      default:
+        g->AddCapacity(c_it->capacity(), type);
+        break;
```
scopatz: I am not sure you need the break in a default. Not that it hurts :).

src/greedy_solver.cc
```diff
@@ -101,30 +103,43 @@ double GreedySolver::Capacity(ExchangeNode::Ptr n, const Arc& a, bool min_cap,
...
   for (int i = 0; i < unit_caps.size(); i++) {
     grp_cap = group_caps[i];
     u_cap = unit_caps[i];
-    cap = grp_cap / u_cap;
+    // special case for unlimited capacities
+    cap = grp_cap == stdmax ? stdmax : grp_cap / u_cap;
     CLOG(cyclus::LEV_DEBUG1) << "Capacity for node: ";
     CLOG(cyclus::LEV_DEBUG1) << "   group capacity: " << grp_cap;
     CLOG(cyclus::LEV_DEBUG1) << "    unit capacity: " << u_cap;
     CLOG(cyclus::LEV_DEBUG1) << "         capacity: " << cap;
```
scopatz: printing the `grp_cap` and `cap` here seems redundant now.

src/prog_translator.h
```diff
@@ -33,8 +34,22 @@ class ProgTranslator {
...
   };
 
+  /// a helper function to translate capacities
+  ///
+  /// @param grp the exchange group
+  /// @param request whether this is a request or not
+  /// @param faux_id a false arc id for surplus supply
+  /// @param cap_rows the capacity rows which will be appended to
+  /// @param ctx the translation context
+  static void XlateCaps(ExchangeNodeGroup* grp,
```
scopatz: I am still not a huge fan of using `X` to mean `Trans`.  (GCSG: http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#General_Naming_Rules).  However, there is wiggle room here since it is more important to be consistent. I dunno. Your call. My 2 drachmas.

src/greedy_solver.h
```diff
@@ -96,36 +96,34 @@ class GreedySolver: public ExchangeSolver {
...
   /// capacities, or the ExchangeNode's remaining qty -- whichever is smaller.
   /// @{
+  double Capacity(ExchangeNode::Ptr n, const Arc& a, double curr_qty);
+  inline double Capacity(ExchangeNode::Ptr n, const Arc& a) {
+    return Capacity(n, a, 0.0);
+  }
+  /// @}
+
+  /// @warning deprecated!
+  /// @{
```
scopatz: Shouldn't these issue a deprecation warning then?

scopatz:
Hi @gidden! Thanks for putting this in.  It looks very useful.  There are a couple of small-ish issues to be dealt with in-line.  The only thing I would request is some more detailed developer docs about the method.  The implementation of the constraints didn't match exactly the math in my mind originally, but does make a certain amount of sense. A write up of what is being implemented I think would have helped. This can be either a mathematical formulation or an algorithmic one.  Thanks!

gidden:
Hi @scopatz. Thanks for the review! I agree that we should have more developer docs. I made #1107 to that effect, and am fine with it blocking this PR for now. The docs in question will likely come from my thesis and/or DRE publication, so should be relatively easy to generate. I will try to get to them post haste.

scopatz:
Hi @gidden! Ok!  Sounds great!

gidden:
moving to 1.4

PR #1108, opened 2015-03-26
fix handling of nested template parsing. fixed alias handling.

rwcarlsen:
This is needed for the separations facility - so quicker review is better.

Previously template parsing assumed that all commas were at the current nesting level for
template args.  This is not true when templates become nested like this:

```c++
  std::map<std::string, std::pair<int, double> >
```

Previously the parse would splitting on the comma inside the pair template
brackets as a part of the outer map's template args.  This is now fixed.
The new recursive support for templated types enabled recursive specification
of alias's.  There were a few spots in the code that weren't correctly updated
to deal with this.  Those are also addressed here.

### DIFF
--- a/cli/cycpp.py
+++ b/cli/cycpp.py
@@ -492,7 +492,9 @@ class VarDeclarationFilter(Filter):
         annotations['index'] = len(state.context[classname]['vars'])
         state.context[classname]['vars'][vname] = annotations
         if 'alias' in annotations:
-            state.context[classname]['vars'][annotations['alias']] = vname
+            alias = annotations['alias']
+            alias = alias if isinstance(alias, STRING_TYPES) else alias[0]
+            state.context[classname]['vars'][alias] = vname
         state.var_annotations = None
 
     def transform_pass3(self, statement, sep):
@@ -732,7 +734,7 @@ class StateAccumulator(object):
                 msg = ("{i}The type of {c}::{n} ({t}) is not a recognized "
                        "primitive type: {p}.").format(
                     i=self.includeloc(), t=t, n=name, c=self.classname(),
-                    p=", ".join(sorted(self.supported_types)))
+                    p=pformat(self.supported_types))
                 raise TypeError(msg)
         return t
 
@@ -1361,10 +1363,6 @@ class InfileToDbFilter(CodeGeneratorFilter):
                 # this member is a variable alias pointer
                 continue
 
-            alias = member
-            if 'alias' in info:
-                alias = info['alias']
-
             if self.pragmaname in info and 'read' in info[self.pragmaname]:
                 impl += info[self.pragmaname]['read']
                 continue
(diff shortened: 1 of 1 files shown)


cli/cycpp.py
```diff
@@ -1986,39 +1984,45 @@ def outter_split(s, open_brace='(', close_brace=')', separator=','):
...
+    j = s.rfind(close_brace)
+    t = [s[:i]]
+    inner = s[i+1:j]
+    t.extend(parse_arg(inner, open_brace, close_brace, separator))
     return t
 
+def parse_arg(s, open_brace, close_brace, separator):
+    """Takes a string containing one or more c++ template args and returns
+    a list of the argument values as strings.
+    """
```
gidden: could you add an example in the documentation?

gidden: (please)

rwcarlsen: Do you mean just in the docstring?

gidden: yup

gidden:
Also, I don't see this fixing the example you gave, e.g.

```
In [7]: from cycpp import parse_template as new

In [8]: from cycppdev import parse_template as old

In [9]: t = 'std::map<std::string, std::pair<int, double> >'

In [10]: new(t) == old(t)
Out[10]: True
```

cli/cycpp.py
```diff
@@ -1986,39 +1984,45 @@ def outter_split(s, open_brace='(', close_brace=')', separator=','):
...
-                                close_brace=close_brace, separator=separator))
+
+    i = s.find(open_brace)
+    j = s.rfind(close_brace)
+    t = [s[:i]]
+    inner = s[i+1:j]
+    t.extend(parse_arg(inner, open_brace, close_brace, separator))
     return t
 
+def parse_arg(s, open_brace, close_brace, separator):
```
gidden: can you match the parent function's signature, please

rwcarlsen:
@gidden - you're right - my example isn't broken.  Here is an example where old cycpp doesn't work:

```python
>>> import cycpp
>>> import cycpp_new
>>> t = 'std::map<std::string, std::pair<double, std::map<int, double> > >'
>>> cycpp.parse_template(t)
['std::map', 'std::string']
>>> cycpp_new.parse_template(t)
['std::map', 'std::string', ['std::pair', 'double', ['std::map', 'int', 'double']]]
```

gidden:
Ok, thanks!

gidden:
Is there a reason `parse_arg` is returning a list of lists? E.g.,
```python
In [19]: from cycpp import parse_arg

In [20]: t = 'std::pair<int, double>'

In [21]: parse_arg(t, '<', '>', ',')
Out[21]: [['std::pair', 'int', 'double']]
```

rwcarlsen:
When parse_arg is called, we are already inside another template's args.  So parse_arg would only ever be called for a type like `std::vector<std::pair<int, double> >`.  And parse_template extends parse_arg's return to a growing recursive type.  In this case, template would have assembled `[std::vector]`.  Then parse arg returns `[[std::pair, int, double]]`.  parse_template then adds each element in the returned list (only one element) resulting in `[std::vector, [std::pair, int, double]]`.  This is necessary because parse_arg has to handle cases where there is more than one template arg - e.g. `std::map<int, std::pair<int, double>` - then parse_arg gets `int, std::pair<int, double` passed to it and returns `[int, [std::pair, int double]]`.  It's sort of confusing - you woudnt' believe how long it took me to write the code/fix the bug.

gidden:
Gotcha. I see that `parse_arg` is only called at `t.extend(parse_arg(inner, open_brace, close_brace, separator))`. Either the documentation for `parse_arg` should be updated (it returns a list of lists) or it should be called as `t.extend([parse_arg(inner, open_brace, close_brace, separator)])` or an equivalent. I personally like the latter, but would accept the former with a docstring update.

rwcarlsen:
I'll try to make the doc comments more clear (tomorrow).  parse_arg doesn't
always return a list of lists - it might only return a list in cases where
there is not template nesting at the current level of recursion in the
parse_[arg/template] calls (e.g. if given "int, double").

cli/cycpp.py
rwcarlsen: done

cli/cycpp.py
rwcarlsen: done.

rwcarlsen:
Should be ready now.  The examples in the docstrings are doctests.

gidden:
Perfect! Thanks for working through the review. I think this will be really nice for anyone else wanting to use/update these functions.

PR #1110, opened 2015-04-01
Add internal tag to cycpp. Modify derived_init tag behavior:

rwcarlsen:
'internal' tag now makes state variables invisible in the xml schema, but
still persisted/loaded normally.  'internal' vars can have default values.
'derived_init' vars now show up in the xml schema as optional and the
derived_init value is overridden if a user specifies the value in the schema.
If a var is both internal and derived_init, then it is not in the schema but
can still have a default value specified.

### DIFF
--- a/cli/cycpp.py
+++ b/cli/cycpp.py
@@ -1526,13 +1526,14 @@ class InfileToDbFilter(CodeGeneratorFilter):
                 continue
             d = info['default'] if 'default' in info else None
 
-            if 'internal' in info:
+            if 'internal' in info and info['internal'] == True:
                 # do NOT combine above and below ifs into a single if
                 if d is not None:
                     mname = member + '_tmp'
                     impl += self._val(t, val=d, name=mname, uitype=uitype, ind=ind)
                     impl += ind + '{0} = {1};\n'.format(member, mname)
-                # else: skip else clause below
+                else:
+                    raise RuntimeError('state variables marked as internal must have a default')
             else:
                 labels = info.get('alias', None)
                 if labels is None:
--- a/cli/cycpp.py
+++ b/cli/cycpp.py
@@ -1525,8 +1525,14 @@ class InfileToDbFilter(CodeGeneratorFilter):
             if key in BUFFERS:
                 continue
             d = info['default'] if 'default' in info else None
-            if 'derived_init' in info:
-                impl += ind + info['derived_init'] + '\n'
+
+            if 'internal' in info:
+                # do NOT combine above and below ifs into a single if
+                if d is not None:
+                    mname = member + '_tmp'
+                    impl += self._val(t, val=d, name=mname, uitype=uitype, ind=ind)
+                    impl += ind + '{0} = {1};\n'.format(member, mname)
+                # else: skip else clause below
             else:
                 labels = info.get('alias', None)
                 if labels is None:
@@ -1557,6 +1563,10 @@ class InfileToDbFilter(CodeGeneratorFilter):
                     impl += ind + '{0} = {1};\n'.format(member, mname)
                     ind = ind[:-2]
                     impl += ind + '}\n'
+
+            # this must run after all other codegen for the current statevar:
+            if 'derived_init' in info:
+                impl += ind + info['derived_init'] + '\n'
 
         # write obj to database
         impl += ind + 'di.NewDatum("Info")\n'
@@ -1732,7 +1742,7 @@ class SchemaFilter(CodeGeneratorFilter):
 
             if key in BUFFERS:  # buffer state, skip
                 continue
-            if 'derived_init' in info:  # derived state, skip
+            if 'internal' in info:
                 continue
 
             opt = True if 'default' in info else False
--- a/tests/cycpp_tests.h
+++ b/tests/cycpp_tests.h
@@ -97,6 +97,19 @@ class Villan: public mi6::Spy {
     }
   std::string enemy;
 
+  #pragma cyclus var {"internal": False}
+  int foo1;
+  #pragma cyclus var {"internal": False, "derived_init": "foo3 = std::exp(2);"}
+  int foo3;
+  #pragma cyclus var {"internal": False, "default": 7}
+  int foo5;
+  #pragma cyclus var {"internal": True, "default": 7}
+  int foo6;
+  #pragma cyclus var {"internal": False, "derived_init": "foo3 = std::exp(2);", "default": 7}
+  int foo7;
+  #pragma cyclus var {"internal": True, "derived_init": "foo4 = std::exp(2);", "default": 7}
+  int foo8;
+
   #pragma cyclus var {\
     "default": [1, 2, 5],\
     }
--- a/tests/cycpp_tests.py
+++ b/tests/cycpp_tests.py
@@ -1005,6 +1005,142 @@ def test_infiletodb_read_map():
 
     yield assert_equal, exp, obs
 
+def test_internal_schema():
+    cases = [
+        {'params': {'internal': True, 'default': 7}, 'want':
+            '  return ""\n    "<interleave/>\\n";\n'
+            },
+        {'params': {'derived_init': 'x *= 42;'}, 'want':
+            ('  return ""\n'
+             '    "<interleave>\\n"\n'
+             '    "    <element name=\\"x\\">\\n"\n'
+             '    "        <data type=\\"int\\"/>\\n"\n'
+             '    "    </element>\\n"\n'
+             '    "</interleave>\\n";\n')
+            },
+        {'params': {'derived_init': 'x *= 42;', 'default': 7}, 'want':
+            ('  return ""\n'
+             '    "<interleave>\\n"\n'
+             '    "    <optional>\\n"\n'
+             '    "        <element name=\\"x\\">\\n"\n'
+             '    "            <data type=\\"int\\"/>\\n"\n'
+             '    "        </element>\\n"\n'
+             '    "    </optional>\\n"\n'
+             '    "</interleave>\\n";\n')
+            },
+        {'params': {'internal': True, 'derived_init': 'x *= 42;', 'default': 7}, 'want':
+            '  return ""\n    "<interleave/>\\n";\n'
+            },
+    ]
+
+    for i, case in enumerate(cases):
+        keys = case['params'].copy()
+        m = MockCodeGenMachine()
+        params = {'type': 'int'}
+        params.update(keys)
+        m.context = {"MyFactory": OrderedDict([('vars', OrderedDict([
+                ('x', params),
+            ]))])}
+        f = SchemaFilter(m)
+        f.given_classname = 'MyFactory'
+
+        impl = f.impl()
+
+        want = case['want']
+        msg = 'case {0} failed\n    ---- got ----\n    {1}\n    ---- want ----\n    {2}'.format(i + 1, impl.replace('\n', '\n    '), want.replace('\n', '\n    '))
+        if want != impl:
+            pprint.pprint(impl)
+            assert_true(False, msg)
+
+def test_internal_infiletodb():
+    # the expected output (i.e. 'want':...) is set as 'throw' if the
+    # configuration should cause an exception.
+    cases = [
+        {'params': {'internal': True}, 'want': 'throw'},
+        {'params': {'internal': True, 'default': 7}, 'want':
+            ('  cyclus::InfileTree* sub = tree->SubTree("config/*");\n'
+             '  int i;\n'
+             '  int n;\n'
+             '  int x_tmp = 7;\n'
+             '  x = x_tmp;\n'
+             '  di.NewDatum("Info")\n'
+             '  ->AddVal("x", x)\n'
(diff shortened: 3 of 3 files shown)


rwcarlsen:
This will need docs on the website.

gidden:
I think this is a good start. I have a few suggestions. A variable should be able to have any combination of `default`, `internal`, and `derived_init` (although I think this should be renamed to `assign`, that can wait for a different pr). The specific difference here is that I don't think derived init should be overridden. 

so we have 7 combinations. Here's how I think they should work out.

default:
  - optional in schema

internal:
  - not in schema

derived_init:
  - non optional in schema, can be used for things like `foo_ = foo_ * bar_`, where the original `foo` and `bar` are supplied via input 

default and internal:
  - not in schema, default value used in init

default and derived_init:
  - optional in schema and same behavior as derived_init above

internal and derived_init:
  - how derived_init works right now, e.g., `foo_ = bar_` with bar supplied in input

default and internal and derived_init:
  - not in schema, default value used in a derivation
  - I can't think of why this would be any different than simply internal + derived_init

rwcarlsen:
On Wed, Apr 1, 2015 at 11:01 AM, Matthew Gidden <notifications@github.com>
wrote:

> I think this is a good start. I have a few suggestions. A variable should
> be able to have any combination of default, internal, and derived_init
> (although I think this should be renamed to assign, that can wait for a
> different pr). The specific difference here is that I don't think derived
> init should be overridden.
>
> so we have 7 combinations. Here's how I think they should work out.
>
> default:
>
>    - optional in schema
>
> internal:
>
>    - not in schema
>
> derived_init:
>
>    - non optional in schema, can be used for things like foo_ = foo_ *
>    bar_, where the original foo and bar are supplied via input
>
> default and internal:
>
>    - not in schema, default value used in init
>
> default and derived_init:
>
>    - optional in schema and same behavior as derived_init above
>
> This case is ambiguous.  If the user doesn't provide a value in xml, then
which do we execute - the default value or the derived init code?  What
happens if user supplies value in xml file - does it override derived_init,
default, both? I think this difficulty forces us to choose one to have
preference over the other if both are defined (both derived_init and
default).

scopatz:
Just glancing at this, I don't understand what you mean by tag.  Do you mean a key in a state var dict?

rwcarlsen:
@scopatz - sorry for the unclear description - yes state var dict key is
correct.

gidden:
Hi Robert,

The derived init code should always be run. If the variable is non-internal, then the derivation should be run on either the user-provided or default value.

Here's a contrived example. Consider a facility that generates power. The implementation assumes the units are in watts, but the input should be user-friendly, and thus defined in megawatts. Therefore, you would have something like

```
#pragma cyclus var {'default': 1000, 'units': MW: 'derived_init': 'power_ *= 1e3;'}
double power_;
```

rwcarlsen:
So you are suggesting just paste in derived_init code after all other
code-gen related to a state var - correct?  While I don't think that that
is necessarily better (I think both ways can lead to users/devs being
surprised), it seems reasonable.  I'll change it.

rwcarlsen:
derived init code is now always generated and placed after other state var generated code.

cli/cycpp.py
```diff
@@ -1568,10 +1578,10 @@ def impl(self, ind="  "):
 
             if key in BUFFERS:  # buffer state, skip
                 continue
-            if 'derived_init' in info:  # derived state, skip
+            if 'internal' in info:  # derived state, skip
                 continue
 
-            opt = True if 'default' in info else False
+            opt = True if 'default' in info or 'derived_init' in info else False
```
gidden: this line shouldn't change

gidden:
I feel uneasy about updating cycpp without adding tests, but I think the right place for testing this is by adding something to `cycpp_tests.h`. However, its usage seems to be deprecated... I started a thread [here](https://groups.google.com/forum/#!topic/cyclus-dev/vRiFZw9rPFo).

cli/cycpp.py
rwcarlsen: If we don't change this line, users will be required to put a value in input files for all derived init vars - that would be very bad IMO.  The  primary use case for derived init is to allow the dev to specify a special sort of default value so the user doesn't have to manually.

rwcarlsen: ...all derived init vars that don't have defaults that is.

gidden: We need dev's to be explicit. The only time optional tags should be provided is if there is a default value (therefore, setting it is optional).

gidden: In other words, the way `derived_init` works with this PR should be different than it works now. Cycamore facilities should have to add an 'internal' key to get the same behavior.

rwcarlsen: You've convinced me it is not as bad as I initially thought.  I'll change.
I don't have time to do the tests though.  I need to be moving on to other
things.

gidden: That's fine, the tests should have been easier (just adjust cycpp_tests.h).
Can you show me a build with an updated archetype that you're working on?

On Wed, Apr 1, 2015 at 4:32 PM, Robert Carlsen <notifications@github.com>
wrote:

> In cli/cycpp.py
> <https://github.com/cyclus/cyclus/pull/1110#discussion_r27614086>:
>
> >                  continue
> >
> > -            opt = True if 'default' in info else False
> > +            opt = True if 'default' in info or 'derived_init' in info else False
>
> You've convinced me it is not as bad as I initially thought. I'll change.
> I don't have time to do the tests though. I need to be moving on to other
> things.
> … <#14c76e625e819381_>
> On Wed, Apr 1, 2015 at 3:24 PM, Matthew Gidden <notifications@github.com>
> wrote: In cli/cycpp.py <
> https://github.com/cyclus/cyclus/pull/1110#discussion_r27613345>: >
> continue > > - opt = True if 'default' in info else False > + opt = True if
> 'default' in info or 'derived_init' in info else False In other words, the
> way derived_init works with this PR should be different than it works now.
> Cycamore facilities should have to add an 'internal' key to get the same
> behavior. — Reply to this email directly or view it on GitHub <
> https://github.com/cyclus/cyclus/pull/1110/files#r27613345>.
>
> —
> <https://github.com/cyclus/cyclus/pull/1110/files#r27614086>.
>



-- 
Matthew Gidden, Ph.D.
Ph. 225.892.3192

rwcarlsen: I don't currently have anything with the internal tag.  But if you want to
test how if it works in real archetypes (the real world) codegen - you can
add internal tag to my cycamore reactor's res_indexes and/or discharged
member vars.

gidden:
```
  File "/home/gidden/work/cyclus/cyclus/cli/cycpp.py", line 1584
    opt = True if 'default' in info:
                                   ^
```

cli/cycpp.py
```diff
@@ -1568,10 +1578,10 @@ def impl(self, ind="  "):
 
             if key in BUFFERS:  # buffer state, skip
                 continue
-            if 'derived_init' in info:  # derived state, skip
+            if 'internal' in info:  # derived state, skip
                 continue
 
-            opt = True if 'default' in info else False
+            opt = True if 'default' in info:
```
gidden: still not full reverted

gidden:
this looks good after the final fix. website and cycamore PRs will be needed in conjunction. I can review/pull them both when they're ready.

cli/cycpp.py
rwcarlsen: oops - that's a syntax error if I ever saw one.

gidden:
I'll note that this also (correctly) causes cycamore test failures. Yay!

scopatz:
I still haven't done a full review, but I do want to second @gidden's concern over tests.  New features must have tests.

scopatz:
Hi all, this looks fine, but certainly needs tests as it adds critical logic to the code generator.  This should be possible to unit test.

gidden:
@scopatz, it appears that the closest thing we have to unit tests in this regard are testing that the dictionary was [read in as expected](https://github.com/cyclus/cyclus/blob/develop/tests/cycpp_tests.py#L188). To wit, `derivedinit` has not been tested in the first place.

I think the most reasonable thing to do here is to update `cycpp_tests.h` and add a test in `cycpp_tests.py` that confirms preprocessing do not fail with the inclusion of these new types.

gidden:
So I now see that the `MockCodeGenMachine` could be used to do this. See #1112 for my arguments for/against.

scopatz:
You could always create a different set of objects that make the testing the piece under scrutiny easy.  The choice isn't only between using the MockCodeGenMachine and integrations tests.  I can't think of a situation where writing tests for this could take more than an hour...

gidden:
@rwcarlsen, I'd like to merge this if you find time for adding a test

gidden:
ping @rwcarlsen, move this to 1.4 if you won't have time to add a test, please

gidden:
Hi @rwcarlsen, it turns out this would be useful to the developer's tutorial. Would you mind terribly adding a test and opening a PR on the release branch?

rwcarlsen:
I'm not sure I'll get to it in time.  If at all, it will be mid to late
next week at the earliest.

gidden:
Ok. I'm still up in the air as far as how deep we should go about
init/restart anyway.

On Fri, May 8, 2015 at 2:55 PM, Robert Carlsen <notifications@github.com>
wrote:

> I'm not sure I'll get to it in time. If at all, it will be mid to late
> next week at the earliest.
>
> On Thu, May 7, 2015 at 3:18 PM, Matthew Gidden <notifications@github.com>
> wrote:
>
> > Hi @rwcarlsen <https://github.com/rwcarlsen>, it turns out this would be
> > useful to the developer's tutorial. Would you mind terribly adding a test
> > and opening a PR on the release branch?
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/cyclus/cyclus/pull/1110#issuecomment-100022055>.
>
> >
>
> —
> <https://github.com/cyclus/cyclus/pull/1110#issuecomment-100340352>.
>



-- 
Matthew Gidden, Ph.D.
Ph. 225.892.3192

gidden:
Note that this change will require:

- updating http://fuelcycle.org/arche/cycpp.html
- updating http://fuelcycle.org/arche/tutorial/toolkit.html

Does the "internal" variable hit the annotations? If so, cycic will require an update as well.

gidden:
Based on the [discussion](https://groups.google.com/forum/#!topic/cyclus-dev/elIYEYKQ0XY) previously, it was not clear to me that an "internal" variable also requires a "default". I claim that is an over specification (any internal variable can be initialized in a `Build` method as required. Is there a use case that is not covered by this?

rwcarlsen:
In order to maintain orthogonality of the meanings of the annotations, that is the result.  The meanings and how they intersect is easier to understand if internal means "not in schema" except for when it means "not in schema and with an implicit default value".  And default is the way you make something optional in the schema except for internal.  And the current behavior was patterned after your excellent feedback in a comment farther up.

rwcarlsen:
correction to above comment: s/easier/more confusing/

rwcarlsen:
And the way currently implemented, the preprocessor has fewer special/corner cases to handle and is easier to maintain correct and understand.

gidden:
Ah, so in the "excellent feedback" (thank you), I state that a variable should be able to have simply an `internal` tag. If a variable is `internal` it should not *also* be *required* to have a `default`.

rwcarlsen:
My other points about orthogonality of concepts, and simplicity in implementation still stand.

gidden:
I don't understand why the orthogonality of the concepts is not considered
covered in my previous feedback.

On Tue, May 26, 2015 at 11:22 AM, Robert Carlsen <notifications@github.com>
wrote:

> My other points about orthogonality of concepts, and simplicity in
> implementation still stand.
>
> —
> <https://github.com/cyclus/cyclus/pull/1110#issuecomment-105587727>.
>



-- 
Matthew Gidden, Ph.D.
Ph. 225.892.3192

rwcarlsen:
Orthogonality in this context meaning that the only way to get a default is
to have the default annotation.

rwcarlsen:
This should be ready.

The other reason default needs to be required with internal @gidden is that even though vars can be internal, they still need a column in the archetypes agentstate table.  So there must be a value that we stuff in the initial row of the table for that agent or the table will not be generated with a column for this variable.  And since it is difficult to guess/auto-generate a default for any arbitrary cpp data type, I've opted to just require that the archetype dev also give us a default.  In order to prevent errors, I modified cycpp to raise an exception with an informative error message if the dev tags a var as internal without specifying a default.

gidden:
So I'm ok with this conceptually. However, I'm curious why the explicit default is needed when you get the cpp default for free (at least for all the std and pod types)?

gidden:
Ah, ok, I mixed up who was doing the default generation. Understood.

gidden:
Thanks for coming back to this, it's great!

gidden:
So we will still need to update our website documentation.

From before:

```
Note that this change will require:

updating http://fuelcycle.org/arche/cycpp.html
updating http://fuelcycle.org/arche/tutorial/toolkit.html
```

Would you mind @rwcarlsen ?

PR #1115, opened 2015-04-06
updates preprocessor to handle recursive container indicies. previous be...

gidden:
...havior would over use the index variable and hang in certain cases


scopatz:
It seems like this redeclares `i`, which is an error in many cases.

gonuke:
This is valid C/C++ and should be in all cases.  The declaration of 'i' is local to the for loop under this change and not a global redeclaration at function scope.  Hopefully this is also correct algorithmically.

scopatz:
You can turn this behaviour off though with GCC's `-fno-for-scope`, as was the case in earlier C++ compilers (including GCC) https://gcc.gnu.org/onlinedocs/gcc-3.0.4/gcc_3.html#SEC8

In any event, I feel like it is pretty bad style to be redeclaring variables in an inner scope when the same variable exists at an outer scope, particularly if it is some throw away variable name like `i`.  This would be pretty annoying to debug.

scopatz:
Also, less of an issue, but redeclaring will add another variable to the stack.

scopatz:
Also, also, I was fooling around with C++98 recently for some reason, and saw this error.  Which is how I knew it could happen. That is to say, I have seen this come up in practice.

rwcarlsen:
You can also turn off stack safety and all sorts of other things that make
lots of code we write broken.  But this uses standard, default c++ and
compiler behavior in a ver common way.  And it is much more complicated and
less elegant, readable, etc. to code the alternative - which would involve
coming up with a robust way to create lots and lots of variables with
different names without clashing - and those variables would also be on the
stack anyway.  So no savings there.

rwcarlsen:
Sorry for my bad grammar and typos - I was trying to hurry so I could
return to watching the game.

scopatz:
To be clear, I do not mind:

```c++
for (int i = 0; i < 10; i++) {}
for (int i = 0; i < 10; i++) {}
for (int i = 0; i < 10; i++) {}
```

for all of the reasons that we have outlined here.  What I do mind is 

```c++
int i = 0;
for (int i = 0; i < 10; i++) {}
for (int i = 0; i < 10; i++) {}
for (int i = 0; i < 10; i++) {}
```

which is what this PR generates.  Even though this is valid C++, I think that this style can lead to hard-to-track down errors, especially in a code generated environment.  Also this very clearly has two `i` on the stack whenever you are in a for-loop scope.  I don't think that the extra stack space is a huge deal because (1) it is only an int, and (2) it may be optimized away if it unused. But if it is unused, why is it their at all?

> And it is much more complicated and
less elegant, readable, etc. to code the alternative - which would involve
coming up with a robust way to create lots and lots of variables with
different names without clashing

i0, i1, i2, i3, i4, .... it isn't that hard and is pretty simple to code. You just have to have an i-counter on the code generator.  This probably should be implemented for nested loops as 

```c++
for (int i = 0; i < 10; i++) {
  for (int i = 0; i < 10; i++) {
    for (int i = 0; i < 10; i++) {}
  }
}
```

is pretty bad style too, in my opinion.

gonuke:
I agree that this is bad style.  I hadn't realized that this reused the same variable name from higher scope.

Does GCSG say anything about this?  If not, it probably should... ;)

scopatz:
> Does GCSG say anything about this? If not, it probably should... ;)

It doesn't says to place the variable name in the smallest scope, which sort of covers this. http://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Local_Variables

rwcarlsen:
ROI seems low to me - esp given that we are over 1 month overdue for the
release and still not terribly close to being done.  I think Matt has made
a reasonable fix to buggy code and that it should be merged in.  Any who
have time to spare are welcome to spare it.  Right now I have ~10 PR's
rebased on top of each other and am cherry-picking left and right inside my
own repos between PR branches and a testing branches that need various
combinations of PR's - it is wasting a lot of my time.

scopatz:
It is literally a one-line fix. Other PRs should not affect the code quality of this PR. That is, each PR should be judged on its own merits.

gidden:
I think 329f5 addresses your concern

scopatz:
This looks great to me, @gidden!  Just the one minor division comment.

gidden:
don't merge this yet. there are more bugs =(

gidden:
ok, now we're good

rwcarlsen:
Good catch!

scopatz:
Thanks @gidden

PR #1120, opened 2015-04-08
Ok, even more robust recursive container support. This works for all pro...

gidden:
...posed cycamore modules. Specifically, n values and i indicies are unique to recursion levels and subtree indicies are explicitly passed in at each stage

I think this fixes everything. I've run it against every proposed cycamore input and we pass all tests.


rwcarlsen:
Still get error when compiling sepfac:

```
In file included from /home/r/src/cyclus/cycamore/src/separations.cc:1:
/home/r/src/cyclus/cycamore/src/separations.h:149:70: error: use of undeclared
      identifier 'i1'
              cyclus::InfileTree* bub = sub->SubTree("efficiencies", i1);
```

cli/cycpp.py
```diff
@@ -1361,21 +1364,24 @@ def read_map(self, member, alias, t, uitype=None, ind="  ", idx=None):
...
         # subtree must be specified if in recursive level
-        tree_idx = '0' if lev == 0 else 'i{}'.format(lev - 1)
+        tree_idx = idx or '0'
         s = ind + '{0}::InfileTree* bub = sub->SubTree("{1}", {2});\n'.format(
             CYCNS, alias[0], tree_idx)
         s += ind + '{0}::InfileTree* sub = bub;\n'.format(CYCNS)
-        s += ind + 'int n = sub->NMatches("{0}");\n'.format(alias[1])
-        s += ind + '{0} {1};\n'.format(type_to_str(t), member)
-        s += ind + 'for (int {idx} = 0; {idx} < n; ++{idx})'.format(idx=idx) + ' {\n'
         with self._nest_idx():
```
rwcarlsen: I think if we remove this with statement, it will work.

rwcarlsen: never mind - I was confused - ignore this comment.

rwcarlsen:
Sorry - git got me.  I had the old ppup branch sitting around and the merge with new one didn't work, but I didn't notice initially.  Everything seems to be working.  Merging.

PR #1124, opened 2015-04-10
faster resbuf

rwcarlsen:
This makes recipe based FCO sims go from ~15 min to ~2 min.  Also I threw in a one line fix for sim_init where prototypes added post simulation init weren't snapshotted properly (the new deployinst utilizes does this and won't work correctly without this fix).

### DIFF
--- a/src/toolkit/res_buf.h
+++ b/src/toolkit/res_buf.h
@@ -139,6 +139,7 @@ class ResBuf {
         rs_present_.erase(r);
       }
 
+      qty_ -= r->quantity();
       rs.push_back(r);
       left -= quan;
     }
@@ -181,6 +182,7 @@ class ResBuf {
     std::vector<typename T::Ptr> rs;
     for (int i = 0; i < n; i++) {
       typename T::Ptr r = rs_.front();
+      qty_ -= r->quantity();
       rs_.pop_front();
       rs.push_back(r);
       rs_present_.erase(r);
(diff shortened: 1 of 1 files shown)


src/context.cc
```diff
@@ -119,6 +119,7 @@ boost::uuids::uuid Context::sim_id() {
 
 void Context::AddPrototype(std::string name, Agent* p) {
   protos_[name] = p;
+  SimInit::SnapAgent(p);
```
gidden: why is this needed here now?

src/toolkit/res_buf.h
```diff
@@ -306,12 +311,12 @@ class ResBuf {
...
-    typename std::list<typename T::Ptr>::iterator it;
-    std::vector<double> qtys;
-    for (it = rs_.begin(); it != rs_.end(); ++it) {
-      qtys.push_back((*it)->quantity());
+    int n = rs_.size();
+    if (n == 0) {
+      qty_ = 0;
+    } else if (n == 1) {
+      qty_ = rs_.front()->quantity();
     }
```
gidden: just to confirm, nothing should be done if `rs_.size() > 1`?

rwcarlsen: correct.

rwcarlsen:
This should be ready.

gidden:
thanks @rwcarlsen!

PR #1129, opened 2015-04-20
added @rwcarlsen's fix for in situ prototype generation with a explanato...

gidden:
...ry comment

@vcloitre, this should fix your issue with deploy inst lifetimes. feel free to review and merge.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -119,7 +119,7 @@ boost::uuids::uuid Context::sim_id() {
 
 void Context::AddPrototype(std::string name, Agent* p) {
   protos_[name] = p;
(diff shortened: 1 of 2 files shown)


src/context.cc
```diff
@@ -119,6 +119,8 @@ boost::uuids::uuid Context::sim_id() {
 
 void Context::AddPrototype(std::string name, Agent* p) {
   protos_[name] = p;
+  // explicit call to support in situ prototype generation
```
rwcarlsen: could we make this say sth like `explicit snapshot required for post-xml-init added prototypes`

PR #1137, opened 2015-04-29
Store mode (def, decl, impl) on codegenfilter 

rwcarlsen:
...so filters can tweak behavior depending on mode. Snapshoting now omits return statement if operating in impl mode.


mbmcgarry:
Thanks @rwcarlsen !

PR #1138, opened 2015-04-29
some performance enhancements

rwcarlsen:
Speed up compmath ops, composition mass/atom funcs, and summing a bit.  This gives ~15% improvement on some FCO-like simulations. This is built on #1137 - so that needs to go in first.

### DIFF
--- a/src/comp_math.cc
+++ b/src/comp_math.cc
@@ -12,32 +12,29 @@ namespace compmath {
 
 CompMap Add(const CompMap& v1, const CompMap& v2) {
   CompMap out(v1);
-  CompMap vv2(v2);
   for (CompMap::const_iterator it = v2.begin(); it != v2.end(); ++it) {
     int nuc = it->first;
-    out[nuc] += vv2[nuc];
+    out[nuc] += it->second;
   }
   return out;
 }
 
 CompMap Sub(const CompMap& v1, const CompMap& v2) {
   CompMap out(v1);
-  CompMap vv2(v2);
   for (CompMap::const_iterator it = v2.begin(); it != v2.end(); ++it) {
     int nuc = it->first;
-    out[nuc] -= vv2[nuc];
+    out[nuc] -= it->second;
   }
   return out;
 }
 
 double Sum(const CompMap& v) {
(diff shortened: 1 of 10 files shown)


src/comp_math.cc
```diff
@@ -12,32 +12,29 @@ namespace compmath {
 
 CompMap Add(const CompMap& v1, const CompMap& v2) {
   CompMap out(v1);
-  CompMap vv2(v2);
   for (CompMap::const_iterator it = v2.begin(); it != v2.end(); ++it) {
     int nuc = it->first;
-    out[nuc] += vv2[nuc];
```
gidden: since this is an optimization PR, is there any reason to skip the `nuc` definition and just have `out[it->first] += it->second;`?

src/comp_math.cc
```diff
@@ -12,32 +12,29 @@ namespace compmath {
...
   }
   return out;
 }
 
 CompMap Sub(const CompMap& v1, const CompMap& v2) {
   CompMap out(v1);
-  CompMap vv2(v2);
   for (CompMap::const_iterator it = v2.begin(); it != v2.end(); ++it) {
     int nuc = it->first;
-    out[nuc] -= vv2[nuc];
```
gidden: same here

gidden:
comments inline, otherwise looks good

src/comp_math.cc
rwcarlsen: Not generally - the compiler does things like that for you (e.g. eliminating unused intermediate variables).  So in cases like this, I generally side on readability.

rwcarlsen:
I added a few small tweaks to the sqlite backend - telling it to use in-memory journaling rather than disk-based.  And also to not wait until transactions are flushed to disk before proceeding with other operations.  This reduces FCO-like (recipe based) run times by ~half!  The trade-off is data integrity if say cyclus crashes or the computer loses power suddenly mid-simulation.  I think the trade-off is definitely worth it.  HDF5 doesn't provide these safety features either.

rwcarlsen:
closed in favor of #1145.

PR #1141, opened 2015-04-29
No00

gidden:
trying to get rid of the `version=0.0` sillyness


gidden:
hmm, this fails to pick up the local cyclus with `version == 9999`. not sure why. chime in if you have any ideas @scopatz. i'm just going to close this for now as it can be worked on later.

PR #1143, opened 2015-04-30
Schema Fix for XML - JSON conversion

scopatz:
I'd be happy to put in an associated cycamore test to fix input files once this is merged in.  This is the smallest impact to the code base and retains consistency across in both JSON and XML forms.  (That is, there are no special tags in either JSON or XML to support the other format.)


gidden:
For those of us who are unclear on why the extra `<item>` level is needed for JSON conversion (like myself), could you expound a bit on that point please, @scopatz?

scopatz:
I am answering right now on the mailing list.

gidden:
Thank you =)

tests/cycpp_tests.py
```diff
@@ -585,16 +585,19 @@ def test_schemafilter():
         '    "<interleave>\\n"\n'
         '    "    <element name=\\"x\\">\\n"\n'
         '    "        <oneOrMore>\\n"\n'
-        '    "            <element name=\\"key\\">\\n"\n'
-        '    "                <data type=\\"int\\"/>\\n"\n'
-        '    "            </element>\\n"\n'
-        '    "            <element name=\\"val\\">\\n"\n'
-        '    "                <data type=\\"double\\"/>\\n"\n'
+        '    "            <element name=\\"item\\">\\n"\n'
+        '    "                <interleave>\\n"\n'
```
gonuke: Is this interleave important?  It seems that the interleave at this level would allow someone to create a file that would visually appear as value=key rather than key=value.  Is there a detail of JSON that requires this?

gonuke: Moreover, doesn't that make it somewhat harder to process since you'll be relying on other information to determine which is the key and which is the value, rather than relying on order?

tests/cycpp_tests.py
```diff
@@ -585,16 +585,19 @@ def test_schemafilter():
         '    "<interleave>\\n"\n'
         '    "    <element name=\\"x\\">\\n"\n'
         '    "        <oneOrMore>\\n"\n'
-        '    "            <element name=\\"key\\">\\n"\n'
-        '    "                <data type=\\"int\\"/>\\n"\n'
-        '    "            </element>\\n"\n'
-        '    "            <element name=\\"val\\">\\n"\n'
-        '    "                <data type=\\"double\\"/>\\n"\n'
+        '    "            <element name=\\"item\\">\\n"\n'
```
gonuke: I think I appreciate the semantic value of this level of nesting....  it basically fills the role of stl::pair<T1,T2> right?  (Does that interpretation make sense?)  Similarly, I can imagine how this level of nesting greatly simplifies the semantics of interpreting the file.

gonuke:
Thanks @scopatz - this makes it much more clear... I guess my inline comments were submitted a little too slowly :)

rwcarlsen:
If we decide to merge this, InfileToDbFilter will also needs significant updates before merging.

scopatz:
A more complete justification for the need for this change is https://groups.google.com/d/msg/cyclus-dev/WK7aP4yPz-I/rt3t-QfjyUwJ

scopatz:
Aslo, I suspect that these issues are part of what caused cycic to break (though that is just a guess).

scopatz:
> If we decide to merge this, InfileToDbFilter will also need significant updates before merging.

@rwcarlsen all of the tests pass as is.  Could you point me to the spot where you think the InfileToDbFilter will need to change?  Only item should affect anything.  Interleaving should be fine.

rwcarlsen:
We don't have any tests that round trip input file data through generated
code and check for correctness.  So the infileToDb and schema could be
completely out of sync and all tests in cyclus will pass.  The only way we
have currently that would catch this is the regression tests in cycamore.

Somewhere in the read_map function the extra <item> nesting will need to be
handled.  The changes will likely be tricky to get right.  Also, facilities
that use custom aliasing will need to have annotations updated with an
extra level of naming for <item> inserted (e.g. Separations).  The uitype
annotations for archetypes will also need to be updated to have an extra
level.

rwcarlsen:
We probably don't actually need/want to change uitype - never mind about that.

FlanFlanagan:
I will hold off on making changes to CycIC until this change is decided
upon.

Currently this

    <element name="in_commods">
        <oneOrMore>
            <element name="key">
                <data type="string"/>
            </element>
            <element name="val">
                <data type="double"/>
            </element>
        </oneOrMore>
    </element>

works in CycIC right now. What Anthony is proposing may involve a change to
the code, I'm fairly sure I've coded it to handle this level of depth. It
may cause some issue with the fancy CycIC ui components of the annotations
(not really in the code so much as in the annotations themselves).

Take for example the following structure.

                <element name="efficiencies">
                    <oneOrMore>
                        <element name="item">
                             </interleave>
                                <element name="comp">
                                   <data type="string"/>
                                </element>
                                <element name="eff">
                                  <data type="double"/>
                                </element>
                            </interleave>
                        </element>
                    </oneOrMore>
                </element>

Do we expect the <element name="item"> to use annotations?

scopatz:
This PR has been updated along with a companion PR to cycamore, cyclus/cycamore#330

gidden:
This and cyclus/cycamore#330 pass locally for me. @rwcarlsen, could you comment on your past concerns with the inclusion of 0bbcbde?

FWIW, all cycamore tests pass (including running all input files), and this is the output for the Separations annotations

```
$ cyclus --agent-annotations :cycamore:Separations
{
<snip>
  "streams": "streams_", 
  "streams_": {
   "alias": ["streams", "commod", ["info", "buf_size", ["efficiencies", "comp", "eff"]]], 
   "doc": "Output streams for separations. Each stream must have a unique name identifying the commodity on which its material is traded, a max buffer capacity in kg (neg values indicate infinite size), and a set of component efficiencies. 'comp' is a component to be separated into the stream (e.g. U, Pu, etc.) and 'eff' is the mass fraction of the component that is separated from the feed into this output stream. If any stream buffer is full, the facility halts operation until space becomes available. The sum total of all component efficiencies across streams must be less than or equal to 1 (e.g. sum of U efficiencies for all streams must be <= 1).", 
   "index": 9, 
   "type": [
    "std::map", 
    "std::string", 
    ["std::pair", "double", ["std::map", "int", "double"]]
   ], 
   "uitype": [
    "oneormore", 
    "outcommodity", 
    ["pair", "double", ["oneormore", "nuclide", "double"]]
   ]
  }, 
</snip>
}
```

@FlanFlanagan, would you mind testing this to see if it does involve a Cycic change? To the best of my knowledge, the injection of an `<item>` tag is needed specifically to denote a pair-like structure inside of a `oneOrMore` to do this full JSON/XML round trip conversion.  The update to cycpp uses the path `item/key<or alias>` and `item/val<or alias>` when reading in a map. Importantly, it knows that it is reading a map and thus to look for that specific tag. There will probably have to be explicit logic related to knowing the upper-level annotations in Cycic. In other words, in the example you gave, when your algorithm reaches the `efficiencies`  level and is moving downward, it will have to inject `item/` paths knowing that `efficiencies` has a type of map, much like @scopatz implemented in 0bbcbde for cycpp.

gidden:
Alternately, @FlanFlanagan, we could require the use of a convention to denote a map entry other than `item` since `item` could easily be used in non-map parts of a future archetype's schema. As a silly example, imagine if we denoted map item tags as `<cyclus_map_item>`. You could then identify that tag and special case your recursion.

Not that I think this is the right way to go,  just another option =)

FlanFlanagan:
I will work on this today and get back to you asap

rwcarlsen:
This looks really good @scopatz.  The only thing I would really like to see added is the ability for alias to specify the 'item' tag  name just like it is able to specify tag names of every other xml element.  So the separations alias would change from:

```python
["streams", "commod", ["info", "buf_size", ["efficiencies", "comp", "eff"]]]
```
to maybe something like:

```python
["streams", "commod", ["info", "buf_size", ["efficiencies", [eff, "comp", "frac"]]]]
```

Currently alias allows the archetype dev to customize *every* xml tag - so not being able to rename "item" is a bit of an eyesore and inconsistent with the spirit of the rest of alias functionality.

gonuke:
Is this necessary for v1.3?  While I think this issue may be important in the long run, round trip between XML<->JSON is never something that we have promised and potentially raises a variety of issues.  In fact, unless we essentially adopt JSON as an alternative I/O format, with full testing, I am concerned that we can't guarantee this anyway.

rwcarlsen:
@gonuke This was my original concern - letting json limitations limit the expresiveness of and driving the evolution of our main supported input format.

scopatz:
The reason this is needed is that the new code generation stuff in v1.3 breaks code in userspace.  Here is a fix that supports users.

gonuke:
I'm worried about the implications on Cycic, which is one layer closer to the core than userspace - @FlanFlanagan - any thoughts?

rwcarlsen:
Just to clarify - 1.3 (before this PR) codegen preserves 1.2 behavior (with generalizations for recursion).  I consider this PR a feature.  Robert F generates forms based on the xml schema (not json schema... I think) which shouldn't be any more problematic for him than before.  So this PR primarily serves to aid Anthony in his FCO efforts - for which he has been using json-based scenario description.

FlanFlanagan:
Well, it's certainly going to have implications to CycIC. There is no way
around that. Changing the schema to add additional interweaves and items
requires updates to the annotations (as Robert C has already pointed out).
If it goes it now or later it will still have the same implications. I'm
already fixing CycIC to handle the additionally structures added to the
codegen for the 1.3 cycamore facilities. So it's not horrible for me to fix
this too.

On a side note, where can I access these changes? Anthony's fork?

On Mon, May 4, 2015 at 3:03 PM, Robert Carlsen <notifications@github.com>
wrote:

> Just to clarify - 1.3 (before this PR) codegen preserves 1.2 behavior
> (with generalizations for recursion). I consider this PR a feature. Robert
> F generates forms based on the xml schema (not json schema... I think)
> which shouldn't be any more problematic for him than before. So this PR
> primarily serves to aid Anthony in his FCO efforts - for which he has been
> using json-based scenario description.
>
> —
> <https://github.com/cyclus/cyclus/pull/1143#issuecomment-98834301>.
>



-- 
Robert Flanagan
Graduate Research Assistant
Nuclear and Radiation Engineering Program
Department of Mechanical Engineering
The University of Texas at Austin

FlanFlanagan:
Or alternatively the output from a cyclus -m on this PR would be equally
useful.

On Mon, May 4, 2015 at 3:39 PM, Robert Flanagan <rf8465@utexas.edu> wrote:

> Well, it's certainly going to have implications to CycIC. There is no way
> around that. Changing the schema to add additional interweaves and items
> requires updates to the annotations (as Robert C has already pointed out).
> If it goes it now or later it will still have the same implications. I'm
> already fixing CycIC to handle the additionally structures added to the
> codegen for the 1.3 cycamore facilities. So it's not horrible for me to fix
> this too.
>
> On a side note, where can I access these changes? Anthony's fork?
>
> On Mon, May 4, 2015 at 3:03 PM, Robert Carlsen <notifications@github.com>
> wrote:
>
>> Just to clarify - 1.3 (before this PR) codegen preserves 1.2 behavior
>> (with generalizations for recursion). I consider this PR a feature. Robert
>> F generates forms based on the xml schema (not json schema... I think)
>> which shouldn't be any more problematic for him than before. So this PR
>> primarily serves to aid Anthony in his FCO efforts - for which he has been
>> using json-based scenario description.
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/cyclus/cyclus/pull/1143#issuecomment-98834301>.
>>
>
>
>
> --
> Robert Flanagan
> Graduate Research Assistant
> Nuclear and Radiation Engineering Program
> Department of Mechanical Engineering
> The University of Texas at Austin
>



-- 
Robert Flanagan
Graduate Research Assistant
Nuclear and Radiation Engineering Program
Department of Mechanical Engineering
The University of Texas at Austin

FlanFlanagan:
Update: The issue with the new cycamore facilities actually existed in the
"defaults" ui tag. Currently a lot of these are left blank in the cycamore
facilities or rather are represented by "[ ]". I think if the field has no
default(s) this should actually just be left out of the annotations all
together. I recommend this because currently if a field has a default value
CycIC set the user level of that field to be 1. This was done to reduce
clutter of the forms for initial users, since a default value implies that
it does not require filling out as far as cycic is concerned. This also
means that any field with a user level of greater than 0 requires default
values.

Thoughts?

On Mon, May 4, 2015 at 3:49 PM, Robert Flanagan <rf8465@utexas.edu> wrote:

> Or alternatively the output from a cyclus -m on this PR would be equally
> useful.
>
> On Mon, May 4, 2015 at 3:39 PM, Robert Flanagan <rf8465@utexas.edu> wrote:
>
>> Well, it's certainly going to have implications to CycIC. There is no way
>> around that. Changing the schema to add additional interweaves and items
>> requires updates to the annotations (as Robert C has already pointed out).
>> If it goes it now or later it will still have the same implications. I'm
>> already fixing CycIC to handle the additionally structures added to the
>> codegen for the 1.3 cycamore facilities. So it's not horrible for me to fix
>> this too.
>>
>> On a side note, where can I access these changes? Anthony's fork?
>>
>> On Mon, May 4, 2015 at 3:03 PM, Robert Carlsen <notifications@github.com>
>> wrote:
>>
>>> Just to clarify - 1.3 (before this PR) codegen preserves 1.2 behavior
>>> (with generalizations for recursion). I consider this PR a feature. Robert
>>> F generates forms based on the xml schema (not json schema... I think)
>>> which shouldn't be any more problematic for him than before. So this PR
>>> primarily serves to aid Anthony in his FCO efforts - for which he has been
>>> using json-based scenario description.
>>>
>>> —
>>> Reply to this email directly or view it on GitHub
>>> <https://github.com/cyclus/cyclus/pull/1143#issuecomment-98834301>.
>>>
>>
>>
>>
>> --
>> Robert Flanagan
>> Graduate Research Assistant
>> Nuclear and Radiation Engineering Program
>> Department of Mechanical Engineering
>> The University of Texas at Austin
>>
>
>
>
> --
> Robert Flanagan
> Graduate Research Assistant
> Nuclear and Radiation Engineering Program
> Department of Mechanical Engineering
> The University of Texas at Austin
>



-- 
Robert Flanagan
Graduate Research Assistant
Nuclear and Radiation Engineering Program
Department of Mechanical Engineering
The University of Texas at Austin

rwcarlsen:
The fields with an empty list default `[]` are were put there explicitly to make them optional.  Their default value is zero-entries and is unlikely to be changed.

scopatz:
> 1.3 (before this PR) codegen preserves 1.2 behavior (with generalizations for recursion).

That is not wholly true. The generalizations also add aliasing, which this PR fixes, as explained on the mailing list.  Also the recursive types are that broke this are new in v1.3.  But v1.3 is 2+ months late and so we were forced to use an interim version of cyclus that was compatible with cymetric (circa 2 months ago.)  The new recursive code generation is broken with respect to that as well, which again, this PR fixes.

> So this PR primarily serves to aid Anthony in his FCO efforts - for which he has been using json-based scenario description.

Please don't make this about me.  This is about all of the FCO transition analysis work.  This includes @deniadjokic, @opotowsky and @russellnibbelink.

rwcarlsen:
>> 1.3 (before this PR) codegen preserves 1.2 behavior (with generalizations for recursion).

> That is not wholly true. The generalizations also add aliasing, which this PR fixes, as explained on > the mailing list.  Also the recursive types are that broke this are new in v1.3.

The aliasing was also generalized recursively, but it was additional functionality added - not necessarily changing previous functionality.

> > So this PR primarily serves to aid Anthony in his FCO efforts - for which he has been using json-based scenario description.

> Please don't make this about me.  This is about all of the FCO transition analysis work.  This includes > @deniadjokic, @opotowsky and @russellnibbelink.

Apologies - I didn't intend it to come across that way.  I just was trying to provide more context for the issues this PR deals with.  You are correct - many other people are involved.  I perhaps abbreviated the involved parties inappropriately.

gidden:
While @FlanFlanagan test rides this to see how hard cycic will be to update, @scopatz, can you comment on the ability to alias the `<item>` tag? My inclination is to agree with @rwcarlsen that if it is required, it should be alias-able.

FlanFlanagan:
I have the 1.3 schema working but am having trouble finding a fix for the
schema changes that Anthony proposed.

It's definitely going to take at least another day to get that in, and I'm
already behind schedule on what I wanted to do this week. So, I'm for
putting that off until the next release...

On Tue, May 5, 2015 at 9:12 AM, Matthew Gidden <notifications@github.com>
wrote:

> While @FlanFlanagan <https://github.com/FlanFlanagan> test rides this to
> see how hard cycic will be to update, @scopatz
> <https://github.com/scopatz>, can you comment on the ability to alias the
> <item> tag? My inclination is to agree with @rwcarlsen
> <https://github.com/rwcarlsen> that if it is required, it should be
> alias-able.
>
> —
> <https://github.com/cyclus/cyclus/pull/1143#issuecomment-99089912>.
>



-- 
Robert Flanagan
Graduate Research Assistant
Nuclear and Radiation Engineering Program
Department of Mechanical Engineering
The University of Texas at Austin

gidden:
Hello everyone,

Based on our timeline, priorities, availability, and requirement to maintain consistency across projects, I am going to suggest waiting on this and cyclus/cycamore#330 until after 1.3.  We have had a fruitful discussion, but some issues are unresolved, namely compatibility with Cycic, but also the ability to alias `item`, item interleaving, and proper round-trip support testing. 

I think there is strong potential for each of these to be addressed, but not in the immediate time frame necessary for 1.3. I suggest we discuss the process of including these PRs in the larger cyclus ecosystem (specifically Cycamore and Cycic) on the list.

Cheers,
Matt

scopatz:
@gidden I completely disagree.  This is 100% needed for the FCO transition analysis use case, which is due at the end of May. I also don't understand why the transition analysis use case is less important is somehow less important than other use cases. 

I don't feel like this has been given a fair shot.  e.g. 

> unresolved, namely compatibility with Cycic, but also the ability to alias item, item interleaving, and proper round-trip support testing. 

1. compatibility with cycic forthcoming and the breakages caused by what is on v1.3 were only the first step there.
2. I was given less than a working day to work on item aliasing, which is frankly a feature request on top of what was supposed to be critical bugfix pr.
3. this PR solves the interleaving issue.
4. The related PR in cycamore along with this one addresses the round tripping issue.

Not merging this PR is tantamount to forcing a permanent fork of cyclus and cycamore for the relevant FCO transition analysis work.  I think it is important to remember who the sponsor and ultimate customer is in all of this.

scopatz:
If @FlanFlanagan is willing to explain the cycic issues to me, I am also happy to take a look. I have done some cycic work in the past.

deniadjokic:
Hi all, I can't say enough how much appreciate everyone working so hard on supporting Cyclus development for the FCO analysis. I'm not totally clued in on all the details of this PR, though I do understand that there is a tradeoff between meeting short-term goals and developing long-term code stability in this case. While I am generally a supporter of the latter paradigm, I do think that meeting this end-of-May deadline is incredibly important for Cyclus. Is there some sort of compromise (depending on everyone's near-term priorities) we could reach to keep issues directly related to the FCO work on one of the front burners for even just a couple of weeks, until after the FCO meeting?

FlanFlanagan:
It is currently causing an out of bounds indexing issue and I can't find
the root of it. I have a lot of other stuff to do on CycIC and I don't have
time to dedicate myself to ensuring this works. Eventually I'm sure I could
get to it, but not in the next week.

You're welcome to take a crack at fixing it if you'd like though.



On Tue, May 5, 2015 at 5:47 PM, Anthony Scopatz <notifications@github.com>
wrote:

> If @FlanFlanagan <https://github.com/FlanFlanagan> is willing to explain
> the cycic issues to me, I am also happy to take a look. I have done some
> cycic work in the past.
>
> —
> <https://github.com/cyclus/cyclus/pull/1143#issuecomment-99253339>.
>



-- 
Robert Flanagan
Graduate Research Assistant
Nuclear and Radiation Engineering Program
Department of Mechanical Engineering
The University of Texas at Austin

FlanFlanagan:
Perhaps someone could clarify to me why the FCO work can not function
without the JSON input file format (unless I'm mistaken the XML style is
working)? I'm willing to put in the time if it is that mission critical to
Cyclus as a whole.

On Tue, May 5, 2015 at 6:08 PM, Denia Djokic <notifications@github.com>
wrote:

> Hi all, I can't say enough how much appreciate everyone working so hard on
> supporting Cyclus development for the FCO analysis. I'm not totally clued
> in on all the details of this PR, though I do understand that there is a
> tradeoff between meeting short-term goals and developing long-term code
> stability in this case. While I am generally a supporter of the latter
> paradigm, I do think that meeting this end-of-May deadline is incredibly
> important for Cyclus. Is there some sort of compromise (depending on
> everyone's near-term priorities) we could reach to keep issues directly
> related to the FCO work on one of the front burners for even just a couple
> of weeks, until after the FCO meeting?
>
> —
> <https://github.com/cyclus/cyclus/pull/1143#issuecomment-99256347>.
>



-- 
Robert Flanagan
Graduate Research Assistant
Nuclear and Radiation Engineering Program
Department of Mechanical Engineering
The University of Texas at Austin

scopatz:
> Perhaps someone could clarify to me why the FCO work can not function without the JSON input file format

@FlanFlanagan we built a JSON based workflow for FCO. The current v1.3 rcs broke downstream user code - both cycic and the JSON stuff.  This is intended as a small, critical bugfix release to address these issues.  It is great that you updated cycic to accommodate v1.3. That is exactly what I am trying to do for FCO transition analysis here.

scopatz:
Also, on a code quality note, a stable code should never break down stream code on a minor or micro release.

scopatz:
Which is why this is needed *before* v1.3 and merging any later will necessitate a fork and lose any credibility towards stability.

rwcarlsen:
For recipe-based FCO work (not for bright-based), 1.3 will not have a
storage facility FWIW - so out of release code will have to be used
anyway.  Also, I am -1 on this without support for aliasing <item>.  Even
with it, however, I'm inclined to wait post 1.3.

gonuke:
I've been trying to follow the details around the issue/PR and I think I have a pretty good handle on it, but maybe I'm still missing something.

Are there examples of input files that worked with v1.2 and the FCO JSON workflow that stop working with v1.3-rc?  From what I understand, this would surprise me.

It would NOT surprise me if the FCO JSON workflow is unable to take advantage of new v1.3-rc archetypes, because of the implementation of new capability to support those archetypes.

Do I have this wrong?  If so, please help me understand.

gonuke:
In the meantime... I think we need to consider @scoptaz more subtle point.  If this PR doesn't go in 1.3, it can't go in until 2.0, because it will definitely break compatibility if we release 1.3 without it.

rwcarlsen:
@gonuke - that is only if we declare all cycamore archetypes as
non-experimental for the 1.3 release.  I wasn't aware of any plans to do
so, although I am not necessarily against it.  We have already changed
archetype schemas since 1.2.  I am okay with this being merged only if we
support aliasing the 'item' tag.

gonuke:
@rwcarlsen - what if someone develops an archetype without the 'item' tag based on 1.3 and then we require the 'item' tag in the next release.  The only way we could maintain support is to allow both with and w/o the item tag, which sounds even more awful! It's not just Cycamore compatibility we need to preserve.

scopatz:
I have now updated this PR to add item aliasing, as well as a PR that documents the new aliasing. 

> Do I have this wrong? If so, please help me understand.

You basically have it right. However, I would add a couple of points. The FCO workflow was developed in Feb and March and worked with a circa March 1 - 15th version of the code.  Given our nominal 3 month release cycle and that the majority of the work was happening in cycamore, I felt it reasonable to assume that what was available and valid then would be available and valid in v1.3.  Note that we could not have been based on v1.2 because we needed the extra features that supported cymetric and the time series. 

So this PR fixes an issue that was introduced between when I started using a standard version of cyclus in a psuedo-v1.3 mode and now.  Moreover, this PR is small compared to the whole size of the v1.3 effort to date and is *very* well tested, especially relative to other parts of the code.  This makes both the XML more consistent and the JSON workflow (which was around in early Feb) actually work.

Furthermore, we have often talked about having the other input formats, like we have other backends. I think that these are equally valuable. The JSON workflow was put in as a stop-gap measure to prevent a Python-based input file and linking against the Python interpreter in the short- to medium-term. There is basically no reason that this shouldn't work and/or be allowed to break. I think this is particularly true when the fix is small, well-contained, and exists, and supports our users and sponsors, like this one.

Finally, what the FCO transition analysis really needs is a stable release to be based on.  So unless there is going to be a v1.4 between now and the end of the month, the PR needs to go in, one needs to go in that fixes the same issues, or we have to fork cyclus.

rwcarlsen:
Looks good to me.  I'll defer to @gidden as the original reviewer and release manager.

rwcarlsen:
line 1556 should be:

```diff
elif t in ['std::list', 'std::set', 'std::vector']:
-    name = 'list' if isinstance(cpptype, STRING_TYPES) else ['list']
-    if names[0] is not None:
-        name = names[0]
+    name = 'list' if names[0] is None else names[0]
```

This is not a bug introduced by this PR, but I though it might be worth fixing if this is going in anyway.

cli/cycpp.py
```diff
@@ -1547,8 +1563,11 @@ def _buildschema(self, cpptype, schematype=None, uitype=None, names=None):
             impl += '</element>'
         elif t == 'std::map':
             name = 'map'
-            if names[0] is not None:
-                name = names[0]
+            if isinstance(names[0], STRING_TYPES):
+                names[0] = [names[0], None]
```
rwcarlsen: This needs an else clause added:

```python
        elif names[0] is None:
            names[0] = [name, None]
``` 
to handle cases where the map is not a top level state var (i.e. is nested) and the archetype dev hasn't provided a custom alias.

scopatz: you are right.

scopatz: put in a fix

gidden: Hi everyone,

The show stopper here for me is compatibility with the rest of the Cyclus
project, namely Cycic/Cyclist. I suggest we bring this discussion up on the
dev call today and do our best to figure out the right path forward for
everyone.

Cheers,
Matt


On Wed, May 6, 2015 at 9:45 AM, Anthony Scopatz <notifications@github.com>
wrote:

> In cli/cycpp.py
> <https://github.com/cyclus/cyclus/pull/1143#discussion_r29764732>:
>
> > @@ -1547,8 +1563,11 @@ def _buildschema(self, cpptype, schematype=None, uitype=None, names=None):
> >              impl += '</element>'
> >          elif t == 'std::map':
> >              name = 'map'
> > -            if names[0] is not None:
> > -                name = names[0]
> > +            if isinstance(names[0], STRING_TYPES):
> > +                names[0] = [names[0], None]
>
> put in a fix
>
> —
> <https://github.com/cyclus/cyclus/pull/1143/files#r29764732>.
>



-- 
Matthew Gidden, Ph.D.
Ph. 225.892.3192

scopatz:
> Hi everyone, The show stopper here for me is compatibility with the rest of the Cyclus project, namely Cycic/Cyclist. I suggest we bring this discussion up on the dev call today and do our best to figure out the right path forward for everyone. Cheers, Matt

If @FlanFlanagan could point me to a part of cycic that is failing or show me a traceback, I will work on it starting once Denia leaves this afternoon.  Also, I think it should be noted that this is much higher standard than the code that prompted this was held to.

scopatz:
Given the discussion on the call today, and @gonuke's email, is this going to go in?  cycic no longer seems like a roadblock.

gonuke:
Dumb question: should I be merging this to `develop` or `v1.3.0-release` or both?

mbmcgarry:
I second @gonuke 's question.  I assume we should at minimum merge into v1.3.0-release?

gidden:
You should definitely merge into v1.3.0-release. Per the specifics of CEP3,
the develop branch should be a superset of the release branch. However, I
have been merging the release branch into develop only when a new candidate
is released. The other (perhaps better) option wold have been opening PRs
into both branches simultaneously and merging both. However, we didn't
start this release that way (and I don't think we've ever done that
before), so I have found it cleaner to update develop when updating the
candidate.

On Fri, May 8, 2015 at 10:14 AM, mbmcgarry <notifications@github.com> wrote:

> I second @gonuke <https://github.com/gonuke> 's question. I assume we
> should at minimum merge into v1.3.0-release?
>
> —
> <https://github.com/cyclus/cyclus/pull/1143#issuecomment-100267747>.
>



-- 
Matthew Gidden, Ph.D.
Ph. 225.892.3192

scopatz:
Please merge this into this branch (v1.3.0-rc1) so that the PR says that it was merged rather than closed and then also merge it into the v1.3.0-release branch, since that is the proper release branch.  Also, when merging a PR into a release branch as a reviewer, I have almost always also merged that branch into develop so folks can start developing off of it, though it is fine to wait for another rc or the full release.

gonuke:
Thanks @scopatz!!  This was also merged manually against `v1.3.0-release` in e7b11e76c0f7861a8c583d26a28c622994f31770

scopatz:
Thanks @gonuke!

PR #1146, opened 2015-05-04
Fix trade record

rwcarlsen:
Previously, if a facility modified a received resource in its Accept[Matl/Prod]Trades function, DRE wouldn't record the traded resource until after AcceptTrades returned - resulting in the incorrect resource id and state being recorded as what was transacted.  This tweaks DRE to record the trades and resource info after trades are completely resolved but before the materials are sent to the requesters.  Very tricky bug to find!

### DIFF
--- a/src/trade_executor.h
+++ b/src/trade_executor.h
@@ -51,12 +51,12 @@ class TradeExecutor {
 
   /// @brief execute all trades, collecting responders from bidders and sending
   /// responses to requesters
-  /// @deprecated Use ExecuteTrades(Context*) instead.
+  ///
+  /// @deprecated Use ExecuteTrades(Context*) instead.  This function just
+  /// calls ExecuteTrades(NULL).
   void ExecuteTrades() {
     Warn<IO_WARNING>("this function does not record trades to the database");
-    GroupTradesBySupplier(trade_ctx_, trades_);
-    GetTradeResponses(trade_ctx_);
-    SendTradeResources(trade_ctx_);
+    ExecuteTrades(NULL);
   }
 
   /// @brief execute all trades, collecting responders from bidders and sending
--- a/src/trade_executor.h
+++ b/src/trade_executor.h
@@ -51,9 +51,22 @@ class TradeExecutor {
 
   /// @brief execute all trades, collecting responders from bidders and sending
   /// responses to requesters
+  /// @deprecated Use ExecuteTrades(Context*) instead.
   void ExecuteTrades() {
+    Warn<IO_WARNING>("this function does not record trades to the database");
     GroupTradesBySupplier(trade_ctx_, trades_);
     GetTradeResponses(trade_ctx_);
+    SendTradeResources(trade_ctx_);
+  }
+
+  /// @brief execute all trades, collecting responders from bidders and sending
+  /// responses to requesters
+  void ExecuteTrades(Context* ctx) {
+    GroupTradesBySupplier(trade_ctx_, trades_);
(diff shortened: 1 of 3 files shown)


gidden:
Thanks for the leg work, @rwcarlsen. Could you slap a deprecated warning in the documentation of the old API?

rwcarlsen:
done.

gidden:
Testing this would either require a full integration test (effectively what is in cyclus/cycamore#333) or a mock resource exchange (which we don't have, but I opened #1147) combined with a test context and mock back end. I'm okay with accepting the two PRs in conjunction as needed.

rwcarlsen:
Right.  I think the cycamore fab test will have to be good enough for now.
This is an important bug fix, and we unfortunately don't have the time now
for a better test.

gidden:
Hmm, I'm getting an ABI test error with
```
-cyclus::TradeExecutor<cyclus::Material>::ExecuteTrades()
-cyclus::TradeExecutor<cyclus::Product>::ExecuteTrades()
```

gidden:
So, this fails for me locally because the old `ExecuteTrades()` never is seen in a execution pathway, therefore may not be getting picked up as part of the symbols. If I add a fake pathway, e.g.,
```
    // execute trades!                                                                                                                                                                   
    TradeExecutor<T> exec(trades);
    if (trades.size() < 0) {
      exec.ExecuteTrades();
      exec.RecordTrades(ctx_);
    }
    exec.ExecuteTrades(ctx_);
```
then all tests pass.

gidden:
I can't think of anything that isn't a completely hacky work around. Can you just add both of those functions to the `api_blacklist` in `symbchk.py`?

rwcarlsen:
blacklisted.

src/trade_executor.h
```diff
@@ -51,12 +51,25 @@ class TradeExecutor {
 
   /// @brief execute all trades, collecting responders from bidders and sending
   /// responses to requesters
+  /// @deprecated Use ExecuteTrades(Context*) instead.
   void ExecuteTrades() {
+    Warn<IO_WARNING>("this function does not record trades to the database");
     GroupTradesBySupplier(trade_ctx_, trades_);
```
gidden: @rwcarlsen, it looks like this function is just implementing what is now `ExecuteTrades(NULL)`. Can you update this to reflect that?

src/trade_executor.h
```diff
@@ -51,12 +51,25 @@ class TradeExecutor {
 
   /// @brief execute all trades, collecting responders from bidders and sending
   /// responses to requesters
+  /// @deprecated Use ExecuteTrades(Context*) instead.
   void ExecuteTrades() {
+    Warn<IO_WARNING>("this function does not record trades to the database");
```
gidden: @rwcarlsen, it looks like this function is just implementing what is now ExecuteTrades(NULL). Can you update this to reflect that?

gidden:
ok, this passes now. final comment then ready to merge.

src/trade_executor.h
rwcarlsen: done.

gidden:
Ah sorry, I should have been clearer. I meant that the function seemed to be doing exactly what `ExecuteTrades(NULL)` would do. I think one of two things should happen (you choose):

- remove body of `ExecuteTrades()`
- update body of `ExecuteTrades()` to invoke `ExecuteTrades(NULL)`

rwcarlsen:
done.

gidden:
thanks @rwcarlsen

PR #1148, opened 2015-05-07
minor change to readme, for MacOSX safety. 

katyhuff:
The previous instructions were dangerous for macs. Mixing homebrew and macports is risky. People who use homebrew should stick to homebrew. The new instructions suggest the right thing - people using homebrew should replace port commands with brew commands.

I know you're trying to get a release out, so take or leave this. I'm just going over the docs and whatnot to try to be helpful with fresher eyes.


mbmcgarry:
Good point, thanks @katyhuff !

PR #1149, opened 2015-05-07
just for general community flavor, slight change to readme

katyhuff:
I feel that while the threat of flogging is funny, it's not totally appropriate and gives the wrong impression of what is actually a very kind, welcoming, culturally sensitive cyclus community.

Again, sorry. I know you're trying to get a release out. I was just too swamped earlier to get around to looking over these main documentation issues. take it or leave it.


scopatz:
I agree. That is wholly inappropriate.  Thanks @katyhuff

PR #1150, opened 2015-05-08
Canonizes and dumps alias, always.

scopatz:
This should be helpful to @FlanFlanagan.


gidden:
This builds and tests fine for me. Further, it looks to be working as advertised in the wild. Thanks @scopatz.

PR #1152, opened 2015-05-12
add enrichment time series

rwcarlsen:


gidden:
@rwcarlsen, can you add a test to `tests/toolkit/timeseries_tests.cc` please. it should be trivial.

gidden:
Also, do we have any sort of convention for what should be added as a kernel vs. "user" time series? For the record, I am +1 for these being kernel time series and supporting a variety of time series types in the kernel.

src/toolkit/timeseries.cc
```diff
@@ -8,6 +8,15 @@ void RecordTimeSeries<POWER>(cyclus::Agent* agent, double value) {
   RecordTimeSeries<double>("Power", agent, value);
 }
 
+template <>
+void RecordTimeSeries<ENRICH_SWU>(cyclus::Agent* agent, double value) {
```
gidden: @rwcarlsen, do you have a case for naming this `ENRICH_SWU` vs. `SWU`? I know it gives semantic meaning, but in general we have eschewed semantics in favor of explicit joining in order to discover agent/prototype type.

src/toolkit/timeseries.cc
```diff
@@ -8,6 +8,15 @@ void RecordTimeSeries<POWER>(cyclus::Agent* agent, double value) {
...
+  RecordTimeSeries<double>("EnrichmentSWU", agent, value);
+}
+
+template <>
+void RecordTimeSeries<ENRICH_FEED>(cyclus::Agent* agent, double value) {
```
gidden: Whereas here `FEED` would likely be an overloaded label.

src/toolkit/timeseries.cc
rwcarlsen: It isn't meant to be archetype specific - it is just meant to indicate that the swu is associated with some sort of generalized material enrichment.  And this way - people know to associate the feed and swu series with each other.

gidden: Ok, sounds good, thanks!

rwcarlsen:
tests added

gidden:
Thanks, this looks good. I'll merge when CI finishes.

gidden:
thanks @rwcarlsen

PR #1155, opened 2015-05-13
expand on Set API in material buy policy with defaults

gidden:
helpful for dev tut


mbmcgarry:
Manually merged into develop with commit 258c5f7..634097f
Thanks @gidden !

PR #1156, opened 2015-05-13
Emptyapi

gidden:
helpful for dev tut

### DIFF
--- a/src/toolkit/resource_buff.h
+++ b/src/toolkit/resource_buff.h
@@ -80,6 +80,11 @@ class ResourceBuff {
     return mats_.empty();
   }
 
+  /// Returns true if there are no mats in mats_
+  inline bool Empty() const {
+    return empty();
+  }
+
   /// PopQty pops the specified quantity of resources from the buffer.
   ///
   /// Resources are split if necessary in order to pop the exact quantity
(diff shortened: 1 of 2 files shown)


rwcarlsen:
ResourceBuff is deprecated.  And ResBuf already has an empty function.

gidden:
`ResBuf` is deprecated? The cycamore `Sink` still uses it.

gidden:
I'll open a separate PR. The issue at hand is not that it needs an `empty()` function, rather that the capitalization is not consistent between the `empty()` function and the `Pop()` and `Push()` functions.

rwcarlsen:
Well - ResBuf was meant to replace it as better, faster, and easier to use.  With ResourceBuff, you have to do all sorts of annoying casting.  Some of the functions were made lower case for two reasons:

* they are practically getters
* to be consistent with what is seen in the c++ std lib

gidden:
I don't disagree with using `ResBuff`. The `Sink` should have been updated when it was deprecated; however.

I know that they are practically getters/setters; however, it seems strange that some container-like functions are capitalized and others aren't.

gidden:
Let's move the discussion to #1157

PR #1157, opened 2015-05-13
Empty API for ResBuff

gidden:
now on the ResBuf

related to #1156

### DIFF
--- a/src/toolkit/res_buf.h
+++ b/src/toolkit/res_buf.h
@@ -102,8 +102,11 @@ class ResBuf {
   inline double space() const { return std::max(0.0, cap_ - qty_); }
 
   /// Returns true if there are no resources in the buffer.
+  /// @{
   inline bool empty() const { return rs_.empty(); }
-
+  inline bool Empty() const { return empty(); }
+  /// @}
+  
   /// Pops and returns the specified quantity from the buffer as a single
   /// resource object.
   /// Resources are split if necessary in order to pop the exact quantity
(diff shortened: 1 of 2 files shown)


rwcarlsen:
The cpp style guide says (https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Function_Names):

> You may also use lowercase letters for other very short inlined functions. For example if a function were so cheap you would not cache the value if you were calling it in a loop, then lowercase naming would be acceptable.

rwcarlsen:
And I'm not necessarily opposed to the renaming fundamentally, but I think it is worse to have 2 functions named also the same that do the same thing rather than have 1 function whose lack of capcase fails to match other API.  And I'd really rather not flat out rename the existing function - so I land on the side of maintaining the status quo as the lesser of 2 evils.

gidden:
ok

On Wed, May 13, 2015 at 3:54 PM, Robert Carlsen <notifications@github.com>
wrote:

> And I'm not necessarily opposed to the renaming fundamentally, but I think
> it is worse to have 2 functions named also the same that do the same thing
> rather than have 1 function whose lack of capcase fails to match other API.
> And I'd really rather not flat out rename the existing function - so I land
> on the side of maintaining the status quo as the lesser of 2 evils.
>
> —
> <https://github.com/cyclus/cyclus/pull/1157#issuecomment-101814222>.
>



-- 
Matthew Gidden, Ph.D.
Ph. 225.892.3192

gidden:
thanks for the discussion @rwcarlsen

PR #1158, opened 2015-05-13
make policies use resbufs

gidden:
hotfix for policies using resource buffers


mbmcgarry:
Manually merged into develop on commit 634097f..479e82e
Thanks @gidden !

rwcarlsen:
We should probably put in a Warn<DEPRECATION_WARNING>("ResourceBuff is deprecated in favor if ResBuf") in the ResourceBuff constructor.

PR #1159, opened 2015-05-13
hotfix for policies. throughputs of 0 should be possible

gidden:


opotowsky:
This passes and works for me, thanks @gidden!

PR #1160, opened 2015-05-14
some updates to stub facility test suite. no more clone (that comes a…

gidden:
…utomatically from pragmas). no more src_ prefix and _ suffix. grouped cyclus-specific code in an explicit block

useful for dev tut


mbmcgarry:
Looks good, thanks @gidden !
manually merged into develop branch: e06f3ad..c2b61b7

PR #1161, opened 2015-05-14
more consistent/better logging levels

gidden:
for the dev tut


mbmcgarry:
Looks good @gidden !
Merging into develop: c2b61b7..728b8e8

PR #1169, opened 2015-05-18
better docs for enums and allowing abbreviations and full name for fu…

gidden:
…nction types


gidden:
ok, this should be ready to be reviewed

gidden:
ping

mbmcgarry:
I'm looking at this now. I got a failure on nosetests, but it came from an initdb command, so I'm checking to see whether it's actually from another PR that was recently merged into v1.3.0-release.

mbmcgarry:
merged into develop: a61e86e..cf069e4
thanks @gidden !

PR #1183, opened 2015-06-07
added uilabels and reordered for Cycic

mbmcgarry:
Added uilabels to pragma variables, reordered variables to be logical in Cycic, and wrapped some lines.

### DIFF
--- a/agents/k_facility.h
+++ b/agents/k_facility.h
@@ -137,67 +137,92 @@ class KFacility : public cyclus::Facility {
 
  private:
   /// This facility has one output commodity and one input commodity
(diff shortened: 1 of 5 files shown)


mbmcgarry:
One thing that was very confusing: k-facility has in_capacity, out_capacity, and current_capacity, and they are all defined as being per timestep. I labeled in and out_capacities as throughputs, but I'm not sure what 'current_capacity' is.

gidden:
I suspect `current_capacity` is an "internal" variable...

gidden:
looks good, thanks @mbmcgarry!

Issue #1035, opened 2014-10-29
put gtest in cyclus-specific location


rwcarlsen:
currently, we install our gtest header and lib files into `[prefix]/include` and `[prefix]/lib` directly.  These install locations collide with distributions' gtest packages :(   We should either use distribution gtest dependency or move our gtest into cyclus subdirs.

gidden:
@rwcarlsen, can you give an update on this issue?

rwcarlsen:
no progress yet.

gidden:
to 1.4

PR #1188, opened 2015-06-29
move gtest install locations to be cyclus-specific

gidden:
this will allow us to have special knowledge of our gtest location, not overwrite another gtest install, and not have our gtest install be over-written.


rwcarlsen:
Also - This fixes #1035 - a longstanding issue - yay!

gidden:
yay!

PR #1192, opened 2015-07-09
archetype versioning for agents and db

rwcarlsen:
The cyclus agents report the same version as Cyclus.  The stubs also do, but I implemented a sample of how we expect others to do it - i.e. via a config step querying version control and filling in a template header.  Cycamore PR to come.


rwcarlsen:
Build and everything completed fine, but:

```
+conda install --use-local cyclus=0.0
Fetching package metadata: ......Error: Could not find Binstar user cyclus
```

Ideas?

gidden:
They've been having trouble all day; I'd suggest relaunching.

On Thu, Jul 9, 2015 at 12:44 PM, Robert Carlsen <notifications@github.com>
wrote:

> Build and everything completed fine, but:
>
> +conda install --use-local cyclus=0.0
> Fetching package metadata: ......Error: Could not find Binstar user cyclus
>
> Ideas?
>
> —
> <https://github.com/cyclus/cyclus/pull/1192#issuecomment-120083974>.
>



-- 
Matthew Gidden, Ph.D.
Ph. 225.892.3192

gidden:
Hi @rwcarlsen, I think there should probably be a default implementation of the `version()` call in case devs don't go through our build process *and* don't implement it. Otherwise we can see this:
```
$ cyclus --agent-version :cycamore:Reactor
pure virtual method called
terminate called without an active exception
Aborted (core dumped)
```

gidden:
Also, I get the following sqlite output 
```
$ cyclus null_sink.xml 
....
```
fo

rwcarlsen:
There *is* a default implementation of version - it returns ``'unspecified'``.  The problem you are seeing Occurs when we dynamically load archetypes that were compiled with old agent headers that didn't have the version function.  I don't think there is any way to avoid that - archetype libs will just have to be recompiled.

rwcarlsen:
There is a bug (which I'm working on) where the agentspec wasn't being recorded correctly in the db - good find.

gidden:
Gotcha, sorry I didn't catch that.

On Fri, Jul 10, 2015 at 12:20 PM, Robert Carlsen <notifications@github.com>
wrote:

> There *is* a default implementation of version - it returns 'unspecified'.
> The problem you are seeing Occurs when we dynamically load archetypes that
> were compiled with old agent headers that didn't have the version function.
> I don't think there is any way to avoid that - archetype libs will just
> have to be recompiled.
>
> —
> <https://github.com/cyclus/cyclus/pull/1192#issuecomment-120464303>.
>



-- 
Matthew Gidden, Ph.D.
Ph. 225.892.3192

rwcarlsen:
Bug fixed.  This should be ready for a final look once CI passes.

gidden:
so, do you only need a `foo_version.in` file if you're making a multi-archetype library? we'll likely need to update our hello cyclus and related documentation on the website to describe what's going on here.

gidden:
Can confirm DB bug fix

On Fri, Jul 10, 2015 at 12:33 PM, Robert Carlsen <notifications@github.com>
wrote:

> Bug fixed. This should be ready for a final look once CI passes.
>
> —
> <https://github.com/cyclus/cyclus/pull/1192#issuecomment-120469780>.
>



-- 
Matthew Gidden, Ph.D.
Ph. 225.892.3192

rwcarlsen:
Archetype devs are allowed to provide (or not) the version however they want.  The only requirement is that the ``version`` function returns the desired string.  For us in cycamore, we'll probably use a cmake config step to drop the tag/hash version into into a version header that defines a constant used by archetype version functions.  Some devs might opt for a makefile solution or even manual update.  We leave it up to them.  Although we can put whatever convention we want to recommend into the cycstub repo.

gidden:
I agree with all of this. I just want to make sure our Hello Cyclus
rename.py file will catch this and work as expected.

Also, how does this play if someone downloads the source tarball, and
therefore it is not in a git repo?

On Fri, Jul 10, 2015 at 1:32 PM, Robert Carlsen <notifications@github.com>
wrote:

> Archetype devs are allowed to provide (or not) the version however they
> want. The only requirement is that the version function returns the
> desired string. For us in cycamore, we'll probably use a cmake config step
> to drop the tag/hash version into into a version header that defines a
> constant used by archetype version functions. Some devs might opt for a
> makefile solution or even manual update. We leave it up to them. Although
> we can put whatever convention we want to recommend into the cycstub repo.
>
> —
> <https://github.com/cyclus/cyclus/pull/1192#issuecomment-120488654>.
>



-- 
Matthew Gidden, Ph.D.
Ph. 225.892.3192

rwcarlsen:
rename.py shouldn't have to do anything different should it?  The source
tarball won't work correctly - I'm not sure if that means it will compile
without the version or if it won't compile.  I should note that this issue
already existed in cyclus - we were already using a git command to
configure the version.cc.in file from a long time ago.

gidden:
can you please update the `UseCyclus.cmake` file to match the updates in cyclus/cycamore#382 after documentation has been added?

rwcarlsen:
UseCyclus has been updated and everything should be ready.

gidden:
thanks!

PR #1201, opened 2015-07-31
use python 3 in cyclus-deps

gidden:
simple update


rwcarlsen:
We also need to update the circle.yml file to call ``docker run ... nosetests3 -w ...`` instead of just plain nosetests.

rwcarlsen:
And the same change will need to go in cycamore too.

rwcarlsen:
Closed in favor of #1203.

PR #1205, opened 2015-08-25
added checking for double prototype name add, fixes #1204

### LINKED ISSUE #1204: multiple institutions with same name fails quietly
Consider the following case:

- two regions named R_A and R_B
- R_A has a DeployInst named Inst
- R_B has a NullInst named Inst
- each institution has initial facilities
- the DeployInst deploys facilities at some frequency

If both institutions have the same name, no deployment will occur.

My guess is that the nullinst prototype is overwriting the deployinst in the context's protos_ mapping.

gidden:


rwcarlsen:
I am -1 on this, see my comment in #1204.  I'd rather the check/error occur at the input file handling level rather than at the context level.

rwcarlsen:
We discussed adding 2 addprototype calls - one that throws on pre-existing key and one that overwrites pre-existing key.  I'll merge when that is in.

gidden:
updated

rwcarlsen:
Circle isn't picking up your most recent commit on this - could you rebase?

gidden:
done

rwcarlsen:
Good improvement.

PR #1207, opened 2015-08-27
call agents' get[matl/prod][requests/bids] in deterministic order

rwcarlsen:
This eliminates one of the most significant sources of non-determinism remaining.  In simulations I've run, this is enough to get ~exact reproducability.  There are other things like modifying the ``Arc::operator<`` function that could also increase determinism, and also using the ordered traders set for calling accept/get matl trades functions, but lets only do that as needed as it has a tendency to reduce performance, etc.

### DIFF
--- a/src/resource_exchange.h
+++ b/src/resource_exchange.h
@@ -73,7 +73,7 @@ class ResourceExchange {
 
   /// @brief queries traders and collects all requests for bids
   void AddAllRequests() {
-    std::set<Trader*> traders = ctx_->traders();
+    std::set<Trader*, trader_compare> traders = OrderedTraders();
     std::for_each(
         traders.begin(),
         traders.end(),
@@ -83,7 +83,7 @@ class ResourceExchange {
 
   /// @brief queries traders and collects all responses to requests for bids
   void AddAllBids() {
-    std::set<Trader*> traders = ctx_->traders();
+    std::set<Trader*, trader_compare> traders = OrderedTraders();
     std::for_each(
         traders.begin(),
         traders.end(),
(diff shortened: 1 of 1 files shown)


gidden:
This is a great start. `ResourceExchange` objects are managed by the `ExchangeManager` and are unique for each exchange. I think it makes a little more sense to create this set only once on instantiation and keep it as a member variable, given the existing structure and usage.

rwcarlsen:
Ya - I thought about that while I was coding it.  The only case that would break - which I'm not sure whether or not we want to support - is a trader that is unregistered for trading within Get[bla]Requests and so shouldn't be called during Get[bla]Bids.

gidden:
That's true. If we want to support it at some later time, we could add an exchange-specific `AddTrader` method.

rwcarlsen:
Or we could add a sorted traders list to the context that it maintains along side the current unsorted set - and effectively deprecate the current unsorted getter.  However, most of this would be a later thing.  For now, would you still like to add a member var to the exchange preventing 1 of the 2 sorts in the current implementation?

gidden:
It would be my preference, but you can argue it's a premature optimization. I'll merge as is if you don't want to add it. This also seems like something that would be hard to unit test.

rwcarlsen:
If the PR is still waiting around early next week, I'll make the change.  If not, I won't be sad :-)

rwcarlsen:
The tests add traders after the resource exchange is constructed - or do some other non-standard ordering thing that causes them to fail if I initialize the ordered trader list only in the constructor.  Instead, I ended up adding a lazy initialization in AddAllBids, AddAllRequests, AdjustAll functions.  This should be ready.

gidden:
looks and tests good to me, thanks @rwcarlsen

PR #1208, opened 2015-09-10
Implement CEP23 - custom time step duration

rwcarlsen:
Note that there are a couple of TODOs I added to the code to help remind us to update/remove the int casting when backends get uint64_t support.

### DIFF
--- a/src/composition.cc
+++ b/src/composition.cc
@@ -61,7 +61,7 @@ const CompMap& Composition::mass() {
   return mass_;
 }
 
-Composition::Ptr Composition::Decay(int delta) {
+Composition::Ptr Composition::Decay(int delta, uint64_t secs_per_timestep) {
   int tot_decay = prev_decay_ + delta;
   if (decay_line_->count(tot_decay) == 1) {
     // decay_line_ has cached, pre-computed result of this decay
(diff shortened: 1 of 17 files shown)


rwcarlsen:
Can someone try to build this on ubuntu 14? @mbmcgarry @gidden @scopatz?  CI fails because the compiler says uint64_t is an unrecognized type, but it works fine on my machine.  I just want to know if this is a general ubuntu issue or if it is a docker ubuntu image issue.

scopatz:
uint64_t wasn't an official type until C++11, I don't think.

rwcarlsen:
It became "more" official for c++11 - for some meaning of "more".  FWIW it compiles fine on my machine without any c++11 flags or warnings about using c++11 features.  I think it should work.

rwcarlsen:
I think I've got it - just need to explicitly include stdint.h for it to work in less bleeding edge environments.

rwcarlsen:
CI passes - this should be ready for a review.

src/composition.cc
```diff
@@ -121,7 +125,7 @@ Composition::Ptr Composition::NewDecay(int delta) {
   if (atom_.size() == 0)
     return decayed;
 
-  decayed->atom_ = pyne::decayers::decay(atom_, 2419200.0 * delta);
+  decayed->atom_ = pyne::decayers::decay(atom_, static_cast<double>(secs_per_timestep) * delta);
```
mbmcgarry: Should you also add a TODO here?

// TODO: when the backends get uint64_t support, the static_cast here should
+  // be removed.

rwcarlsen: No - this cast is different - it will still be necessary.  pyne::decayers::decay takes a double for the time argument - so I casted secs_per_timestep in order to promote the expression value to a double.

mbmcgarry:
Thanks @rwcarlsen !

PR #1211, opened 2015-10-13
Fixes #1209 and fixes #1210 allowing many traders per agent.

### LINKED ISSUE #1209: Multiple traders of a single manager/agent clobber each other
This bug was introduced in #1207.  In the DRE stage where traders are iterated over, our custom sorting order doesn't allow for multiple traders to come from a single agent/manager; they instead overwrite each other resulting in bad outcomes.

### LINKED ISSUE #1210: Segfaults when using BuyPolicy with no expliciti composition
The buy policy currently uses a NULL composition ptr when no other is specified, resulting in segfaults and other problems when bidders/traders try to query the composition's properties.

rwcarlsen:


rwcarlsen:
Note that this is a single commit and should be fast forward merged.

gidden:
Thanks for tackling, @rwcarlsen. Two questions:

- it appears this PR uses the "semi-deterministic" approach (perhaps the "intra-agent deterministic" approach). Can we prove (or make a reasonably strong statement) that this results in simulations that are "mostly deterministic"?
- we have a "NewBlankMaterial" [method](https://github.com/cyclus/cyclus/blob/develop/src/material.h#L166). This is not the first use case I've seen for wanting a similar "blank" composition. Would you mind adding such a method to the cyclus namespace and using it here?

rwcarlsen:
On Tue, Oct 13, 2015 at 3:27 PM, Matthew Gidden <notifications@github.com>
wrote:

> Thanks for tackling, @rwcarlsen <https://github.com/rwcarlsen>. Two
> questions:
>
>    - it appears this PR uses the "semi-deterministic" approach (perhaps
>    the "intra-agent deterministic" approach). Can we prove (or make a
>    reasonably strong statement) that this results in simulations that are
>    "mostly deterministic"?
>
> Correct.  I don't have any empirical evidence (yet - until cycamore
storage gets in:), and any more formal proof would be difficult and
something I'm not sure how to approach.  Any significant claims would
require meticulously analyzing all potential sources of non-determinism and
tracing them through the simulation.  We have focused here on resource
exchange trader method invocation order, but there are many others.  Ways
in which the current fix could be defeated:

* multiple traders per agent that use a shared random number generator
* multiple traders in an agent share state - i.e. a shared resource buffer
* (possibly?@gidden can confirm?) multiple traders in an agent requesting
same commod with same prefs.

There are probably others.  If determinism is important to any user, it is
pretty easy to avoid these things - all cycamore agents do.



>
>    - we have a "NewBlankMaterial" method
>    <https://github.com/cyclus/cyclus/blob/develop/src/material.h#L166>.
>    This is not the first use case I've seen for wanting a similar "blank"
>    composition. Would you mind adding such a method to the cyclus namespace
>    and using it here?
>
>
It seems to be a bit overkill - all it would do is {return
Composition::CreateFromMass(CompMap());}

gidden:
The style, etc., looks great. The only thing I'm hesitant about is adding a new test file named for a specific bug rather than adding these tests to another, already-existing file. My intuition is to name such a file something like "trade integration tests" or similar, as I would almost expect to see more as the code base gets any more complex. Perhaps "toolkit tests"? I'm asking for more of a dialogue here than making a proper suggestion..

rwcarlsen:
I've seen this pattern used in various other open source projects (tests
named for bugs), and I thought it was kind of neat.  I figured we could
give it a shot here - especially since now Cyclus is entering a more
maintenance heavy phase of its life.  The idea is to link the test to a
specific bug and associated context for provenance and also to avoid
repeating the already available history/discussion on the bug tracker.  I'm
not wedded to the idea, but I'd like to try it if others are okay with it.

rwcarlsen:
@gidden we'd like to get a final verdict from you on this.  If it's okay as is, then we can also have someone on our end merge.

gidden:
Looks good to me, merging.

gidden:
Thanks @rwcarlsen!

rwcarlsen:
[offtopic] Are you flying today?

gidden:
[offtopic] I am =)

Issue #1213, opened 2015-10-27
switch to c++11


rwcarlsen:
newest versions of libxml++ require c++ 11 - i.e. our build will be broken without the ``-std=c++11`` flag.

gidden:
There's a PR in the hopper for this:
https://github.com/cyclus/cyclus/pull/1118

rwcarlsen:
Looks like pyne could use a few explicit casts - it won't compile under c++11:

```bash
/home/r/src/cyclus/cyclus/src/pyne.cc:13435:29: error: non-constant-expression cannot be narrowed from type 'int' to 'hsize_t'
      (aka 'unsigned long long') in initializer list [-Wc++11-narrowing]
  hsize_t data_offset[1] = {row};
                            ^~~
/home/r/src/cyclus/cyclus/src/pyne.cc:13435:29: note: insert an explicit cast to silence this issue
  hsize_t data_offset[1] = {row};
                            ^~~
                            static_cast<hsize_t>( )
/home/r/src/cyclus/cyclus/src/pyne.cc:13459:26: error: non-constant-expression cannot be narrowed from type 'int' to 'hsize_t'
      (aka 'unsigned long long') in initializer list [-Wc++11-narrowing]
  hsize_t nuc_dims[1] = {nuc_size};
                         ^~~~~~~~
/home/r/src/cyclus/cyclus/src/pyne.cc:13459:26: note: insert an explicit cast to silence this issue
  hsize_t nuc_dims[1] = {nuc_size};
                         ^~~~~~~~
                         static_cast<hsize_t>( )
/home/r/src/cyclus/cyclus/src/pyne.cc:13711:29: error: non-constant-expression cannot be narrowed from type 'int' to 'hsize_t'
      (aka 'unsigned long long') in initializer list [-Wc++11-narrowing]
    hsize_t chunk_dims[1] ={chunksize};
                            ^~~~~~~~~
/home/r/src/cyclus/cyclus/src/pyne.cc:13711:29: note: insert an explicit cast to silence this issue
    hsize_t chunk_dims[1] ={chunksize};
                            ^~~~~~~~~
                            static_cast<hsize_t>( )
```

rwcarlsen:
fixed by #1216.

PR #1215, opened 2015-10-28
update amalgamated pyne to pyne/pyne:0cbc4a3

rwcarlsen:
prep for fixing #1213


gidden:
passes locally, thanks!

Issue #1199, opened 2015-07-29
Remove Third-Party Header File Leakage


gidden:
Identified during our Summer '15 hackathon, we can greatly reduce compile time (of Cyclus and archetypes through cycpp benefits) by only including third-party header files in implementation files. Furthermore, once these headers are no longer leaked, archetype devs' CMake includes will *only* require a find_package call for Cyclus (and whatever other libraries they depend upon), greatly simplifying the generation of the CMakeLists.txt file.

Big offenders are likely to be
- [ ] xml/xml++
- [ ] sqlite
- [ ] hdf5
- [ ] coin

The one issue is that coin-related objects are sometimes used in public interfaces, e.g., [here](https://github.com/cyclus/cyclus/blob/develop/src/prog_translator.h#L35). The locations where this is true are localized to the kernel, that is, archetype developer's have no access to these objects *during cyclus simulation run time*. However, it is still API breakage. There should be some consensus as to whether this type of API breakage is worth the non-trivial and possibly quite large compile-time speedups for both kernel and archetype devs.

Finally, the amalgamated pyne  also appears to leak at least HDF5 headers. This will likely have to be addressed as well in parallel.

rwcarlsen:
#1199 updates cyclus for all but hdf5.

rwcarlsen:
Just cycamore to go.  HDF5 will have to wait at least until we can see if hdf5 headers can be removed from pyne.h.

PR #1218, opened 2015-11-03
fix 3rd party dependency header leakage

rwcarlsen:
This buys us some time to decide how we want to deal with c++11 support for stubs, cycamore, etc. by not leaking the c++11 requiring headers (libxml++). And it makes compiling archetypes simpler anyway. Addresses a good chunk of #1199.

### DIFF
--- a/src/prog_translator.h
+++ b/src/prog_translator.h
@@ -12,6 +12,16 @@ namespace cyclus {
 class ExchangeGraph;
 class ExchangeNodeGroup;
 
+/// @brief struct to hold all problem instance state
+struct ProgTranslatorContext {
+  std::vector<double> obj_coeffs;
+  std::vector<double> row_ubs;
+  std::vector<double> row_lbs;
+  std::vector<double> col_ubs;
+  std::vector<double> col_lbs;
+  CoinPackedMatrix m;
+};
+
 /// a helper class to translate a product exchange into a mathematical
 /// program.
 ///
@@ -26,14 +36,7 @@ class ExchangeNodeGroup;
 class ProgTranslator {
  public:
   /// @brief struct to hold all problem instance state
-  struct Context {
-    std::vector<double> obj_coeffs;
-    std::vector<double> row_ubs;
-    std::vector<double> row_lbs;
-    std::vector<double> col_ubs;
-    std::vector<double> col_lbs;
-    CoinPackedMatrix m;
-  };
+  typedef ProgTranslatorContext Context;
 
   /// constructor
   ///
@@ -62,7 +65,7 @@ class ProgTranslator {
   /// @brief translates solution from iface back into graph matches
   void FromProg();
 
-  const ProgTranslator::Context& ctx() const { return ctx_; }
+  const Context& ctx() const { return ctx_; }
 
  private:
   void Init();
--- a/src/infile_tree.cc
+++ b/src/infile_tree.cc
@@ -2,6 +2,7 @@
 #include <iostream>
 #include <sstream>
 
+#include <libxml++/libxml++.h>
 #include <boost/lexical_cast.hpp>
 
 #include "error.h"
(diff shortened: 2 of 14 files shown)


rwcarlsen:
Ping - this is blocking c++11 PR which is blocking me from compiling cyclus and merging other PRs.

rwcarlsen:
Just kidding - not blocking c++11 PR, but it is simple enough to review anyway ;)

rwcarlsen:
Hold on this a sec - and do c++11 first please.

rwcarlsen:
Okay - this is ready.  And it fixes a nosetest failure introduced by the c11 PR due to the stubs cmakelists not explicitly enabling c11 support that was previously necessary due to leaked headers.

mbmcgarry:
Thanks @rwcarlsen !

src/prog_translator.h
```diff
@@ -12,6 +12,16 @@ namespace cyclus {
...
+/// @brief struct to hold all problem instance state
+struct ProgTranslatorContext {
+  std::vector<double> obj_coeffs;
+  std::vector<double> row_ubs;
+  std::vector<double> row_lbs;
+  std::vector<double> col_ubs;
+  std::vector<double> col_lbs;
+  CoinPackedMatrix m;
+};
+
```
Baaaaam: @rwcarlsen 
This specific change cause an ABI issues:

> -cyclus::ProgTranslator::Context::Context()
> -cyclus::ProgTranslator::Context::~Context()

I don't really understand the difference before and after the change. Could you explain why you did it ?
thx

scopatz: The relevant difference is actually farther  down where a struct was replaced by a typedef.

src/prog_translator.h
```diff
@@ -26,14 +36,7 @@ class ExchangeNodeGroup;
...
   /// @brief struct to hold all problem instance state
-  struct Context {
-    std::vector<double> obj_coeffs;
-    std::vector<double> row_ubs;
-    std::vector<double> row_lbs;
-    std::vector<double> col_ubs;
-    std::vector<double> col_lbs;
-    CoinPackedMatrix m;
-  };
+  typedef ProgTranslatorContext Context;
```
scopatz: Here. This may also qualify as an API change.

src/prog_translator.h
```diff
@@ -62,7 +65,7 @@ class ProgTranslator {
   /// @brief translates solution from iface back into graph matches
   void FromProg();
 
-  const ProgTranslator::Context& ctx() const { return ctx_; }
+  const Context& ctx() const { return ctx_; }
```
scopatz: This would also need to be changed back in all likelihood.

Baaaaam: I have a working branch which revert those change:). I did not PR it
because I wished to understand :) I can PR it anytime!

scopatz: Well, this changed the API by moving replacing a struct with a typedef. structs, like classes, have constructors and destructors implicitly added to them, if they are not given.  These member functions are then added to the symbol table.  Because this struct is nested inside of a class, the structs members become nested too.  The typedef just renames the type and - though it does add the new name to the symbol table - it doesn't do anything else since the original symbols live with their original name.

rwcarlsen: Although it is possible for users to depend on this code, it is considered part of our "internal" API.  And, off the top of my head I can't think of any code that a user could write that this would cause to not compile (i.e. this is not really an API change - although certainly is more ABI-breaking-ish).  And we not infrequently make changes to code that require users to recompile their code.  This change was *required* to eliminate some 3rd party dependency header leakage.  If somebody knows of users who are adversley affected by this change or knows of a better way to eliminate header leakage, than reverting would be no problem.  Otherwise, I think we would be shooting ourselves in the foot.

PR #1219, opened 2015-11-18
adds a private new decay function to composition

gidden:
I had a test failure on a fresh Ubuntu install from source complaining about the ABI test.

### DIFF
--- a/release/smbchk.py
+++ b/release/smbchk.py
@@ -40,6 +40,7 @@ api_blacklist = {
     'cyclus::SimInit::LoadComposition',
     'cyclus::TradeExecutor<cyclus::Material>::ExecuteTrades',
     'cyclus::TradeExecutor<cyclus::Product>::ExecuteTrades',
+    'cyclus::Composition::NewDecay',
 }
 
 def load(ns):


rwcarlsen:
One of the primary purposes of having private members is to maintain
flexibility to change them (i.e. implementation) without breaking API.  I
thought we decided to just whitelist false positives like this.

gidden:
you're right, forgot about this. how is this not failing on other systems,
tbw?

gidden:
updated

gonuke:
Can someone remind me when we take a new snapshot of the ABI?  And presumably we then cleanout the blacklist?

rwcarlsen:
@gonuke - probably better make an issue for that - otherwise we'll
certainly forget.

gonuke:
I was accidentally ambiguous with my grammar: can someone tell what me the current policy/practice is for the frequency of updating the ABI snapshot?  Is it documented and, if so, do we also clean out the blacklist on those occasions?

rwcarlsen:
http://fuelcycle.org/cep/cep3.html has some instructions.

gonuke:
I'm ready to merge this after a rebase

rwcarlsen:
Merged.  Thanks @gidden.

PR #1220, opened 2015-12-07
added Close() to FullBackend

scopatz:
This adds a Close() method to FullBackend to allow it to behave more like a true file object when used from libcyclus.  This was implemented because HDF5 keeps an internal cache of datasets, datatypes, and so on that is not cleared until the HDF5 object is itself closed.  Not clearing this cache was causing problems if you had:

1. run a simulation
2. opened a file with cymetric
3. left the interpreter open (via IPython, xonsh, etc)
4. run a new simulation in another process that used the same database
5. querying the new simulation data via cymetric would have partial, malformed data at best.

This PR allows cymetric to fully close the backend and the open an new handle if needed.

@hodger, can you please review.

### DIFF
--- a/src/query_backend.h
+++ b/src/query_backend.h
@@ -377,6 +377,13 @@ class QueryableBackend {
 class FullBackend: public QueryableBackend, public RecBackend {
  public:
   virtual ~FullBackend() {}
+
+  /// Closes the backend, if approriate.
+  virtual void Close() {}
+
+ protected:
+  /// Flag for whether the backend is closed or not.
+  bool closed_ = false;
 };
 
 /// Wrapper class for QueryableBackends that injects a set of Cond's into every
--- a/src/hdf5_back.cc
+++ b/src/hdf5_back.cc
@@ -37,7 +37,11 @@ Hdf5Back::Hdf5Back(std::string path) : path_(path) {
   vldts_[BLOB] = blob_type_;
 }
 
-Hdf5Back::~Hdf5Back() {
+void Hdf5Back::Close() {
+  // make sure we are still open
+  if (closed_)
+    return;
+
   // cleanup HDF5
   Flush();
   H5Fclose(file_);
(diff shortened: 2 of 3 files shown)


rwcarlsen:
Is there any reason that the destructor (i.e. RAII) doesn't work well for
this case?  The destructors are always called at the end of simulations (in
cli/cyclus.cc).

scopatz:
@rwcarlsen - this was in my original comment.  This is for libcyclus, not  cyclus-the-simulator.  Notably the way the Python garbage collector works, you have no way of forcing a deallocation to happen "now," for some value of now.  But in cymetric, there still needs to be a way of closing the database connection, which is a stronger operation than a flush, usually, since flushes may simply spawn out to another thread/process.

rwcarlsen:
From your original comment, it was unclear why the destructor (e.g. what is
done in cyclus the app) would not work for another task.  But the python/GC
issue makes sense.

src/query_backend.h
```diff
@@ -377,6 +377,13 @@ class QueryableBackend {
 class FullBackend: public QueryableBackend, public RecBackend {
  public:
   virtual ~FullBackend() {}
+
+  /// Closes the backend, if approriate.
+  virtual void Close() {}
+
+ protected:
+  /// Flag for whether the backend is closed or not.
+  bool closed_ = false;
```
rwcarlsen: I'd rather that we continue to maintain the precedent of these as abstract, stateless interfaces.  Implementations should decide whether to be idempotent or how they want to handle closing in general.

scopatz:
Yeah, sorry, I guess the Python interpreter is implicit in the reference to cymetric, though one would encounter this same issue in other libcyclus bindings to a garbage collected languages that don't have deep hooks into the GC itself (R, ruby, js, not sure about lua, etc.).

src/query_backend.h
```diff
@@ -377,6 +377,13 @@ class QueryableBackend {
 class FullBackend: public QueryableBackend, public RecBackend {
  public:
   virtual ~FullBackend() {}
+
+  /// Closes the backend, if approriate.
+  virtual void Close() {}
```
rwcarlsen: This would probably better to in the RecBackend class since it is mostly important for (all) cases where we are writing to the db.

hodger:
This looks good to me, thanks @scopatz! Merging

rwcarlsen:
@scopatz - no response to my inline comments? :(

scopatz:
Sorry I have been busy, and thought that @hodger might have had some questions about them.

PR #1225, opened 2015-12-24
Added 4 supported HDF5 types to dbtypes.json 

hodger:
All of these are maps containing a pair. They appear in hdf5_back.cc but did not previously exist at all in dbtypes.json. 

@scopatz could you please review?


scopatz:
Thanks @hodger!

PR #1226, opened 2015-12-29
[WIP] generating hdf5_back.cc code

hodger:
(Not ready to be merged, PR for collaboration.)

H5back_gen.py currently prints to console the case statement within the QueryResult method, using string templating based upon patterns in the hdf5_back.cc file.

@scopatz let me know what you think. Many updates to follow.


src/H5back_gen.py
```diff
@@ -0,0 +1,472 @@
+import json
```
scopatz: This file should be named in `lowercase_with_underscores.py` convention.

src/H5back_gen.py
```diff
@@ -0,0 +1,472 @@
...
+import textwrap
+
+with open('../share/dbtypes.json') as f:
+    rawtable = json.load(f)
+
+#table = set(tuple(row[:-3]) for row in rawtable[1:])
+v3_table = set(tuple(row) for row in rawtable[897:])
```
scopatz: PEP8: global (module-level scope) variables should be named with `ALL_UPPERCASE` convention.

src/H5back_gen.py
```diff
@@ -0,0 +1,472 @@
...
+
+#Takes dbtype as a parameter, returns list of dependecies' dbtypes,
+#first element of list is always original dbtype.
```
scopatz: This comment should be the function docstring.

src/H5back_gen.py
```diff
@@ -0,0 +1,472 @@
...
+            del pieces[i + 1]
+        elif pieces[i] in containers2:
+            pieces[i]+= "_" + pieces[i + 1] + "_" + pieces[i + 2]
+            del pieces[i + 1:i + 3]
+    pieces = [s] + pieces
+    return pieces
+
+dbtest = []
+db_to_cpp = {}
+indent = '    '
```
scopatz: these should be uppercase

src/H5back_gen.py
```diff
@@ -0,0 +1,472 @@
...
+
+for row in v3_table:
+    if row[6] == 1 and row[4] == "HDF5":
+        dbtest+=[row[1]]
+        db_to_cpp[row[1]] = row[2]
+
+class TypeStr(object):
```
scopatz: a docstring here would be nice.

src/H5back_gen.py
```diff
@@ -0,0 +1,472 @@
...
+    
+    @property
+    def cpptype(self):
+        return db_to_cpp[self.db]
+
+    @property
+    def dbtype(self):
+        return self.db
+
+case_template = """
```
scopatz: uppercase, and for the rest of these.

src/H5back_gen.py
scopatz: Also since this is meant to be a script, the first line is usually `#!/usr/bin/env python`.  This is called as shebang line, look up what it does.

src/H5back_gen.py
```diff
@@ -0,0 +1,472 @@
...
+for db in dbtest: 
+    current_type = TypeStr(db)
+    reader = readers[db]
+    ctx = {"t": current_type,
+           "no_close": no_close,
+           "H5Tclose": H5Tclose,
+           "H5Tclose_multi": H5Tclose_multi}
+    query_cases += case_template.format(t = current_type, read_x = textwrap.indent(reader.format(**ctx), indent))
+
+print(query_cases)
```
scopatz: This block should be in a `main()` function.  Then the following code should be added to the file:

```python
if __name__ == '__main__':
    main()
```

src/H5back_gen.py
```diff
@@ -0,0 +1,472 @@
...
+{no_close}\
+"""
+
+vl_reader = """\
+{t.cpptype} x = VLRead<{t.cpptype}, {t.dbtype}>(buf + offset);
+{no_close}\
+"""
+
+uuid_reader = """\
+{t.cpptype} x;' + '\nmemcpy(&x, buf+offset, 16);
```
scopatz: is this newline a typo?

scopatz:
Thanks a ton for putting this in @hodger, let me know when you have cleaned it up!

hodger:
@scopatz I believe I have made all the formatting changes you mentioned. Please let me know if I missed something! Thanks!

src/hdf5_back_gen.py
```diff
@@ -0,0 +1,496 @@
+#!/usr/bin/env python
+"""This module generates HDF5 backend code found in src/hdf5_back.cc"""
+import json
+import textwrap
+
+with open('../share/dbtypes.json') as f:
```
scopatz: This filename should be made a bit more general. Something like:

```python
with open(os.path.join(os.path.dirname(__file__), 'share', 'dbtypes.json'))
```

src/hdf5_back_gen.py
```diff
@@ -0,0 +1,496 @@
...
+    RAW_TABLE = json.load(f)
+
+V3_TABLE = set(tuple(row) for row in RAW_TABLE[897:])
```
scopatz: 897 seems like a magic number, this should be figured out dynamically rather than being hard coded.

src/hdf5_back_gen.py
```diff
@@ -0,0 +1,496 @@
...
+size_t nullpos;
+hsize_t fieldlen;
+H5Tget_array_dims2(field_type, &fieldlen);
+unsigned int itemsize = 2*CYCLUS_SHA1_SIZE;
+{t.cpptype} x;
+for(unsigned int k = 0; k < fieldlen; ++k) {{
+    x[VLRead<{t.sub[1].cpptype}, {t.sub[1].dbtype}>(buf + offset + itemsize*k)] = VLRead<{t.sub[2].cpptype}, {t.sub[2].dbtype}>(buf + offset + itemsize*k + CYCLUS_SHA1_SIZE);
+}}
+{H5TCLOSE}\
+"""
```
scopatz: Rather than doing the trailing `\` on the first and last line of all of these, you can instead end the quote with `""".strip()` rather than just `"""`

src/hdf5_back_gen.py
```diff
@@ -0,0 +1,496 @@
...
+with open('../share/dbtypes.json') as f:
+    RAW_TABLE = json.load(f)
+
+V3_TABLE = set(tuple(row) for row in RAW_TABLE[897:])
+
+def list_dep(orig):
+    """Returns a list of dependent dbtypes.
+    
+    Keyword arguments: 
+    orig -- original dbtype.
```
scopatz: This should be in numpydoc format.  Checkout pyne or xonsh or the numpydoc docs :)

src/hdf5_back_gen.py
```diff
@@ -0,0 +1,496 @@
...
+    
+    for i in range(len(pieces) - 1, -1, -1):
+        if pieces[i] in containers1:
+            pieces[i] += "_" + pieces[i + 1]
+            del pieces[i + 1]
+        elif pieces[i] in containers2:
+            pieces[i]+= "_" + pieces[i + 1] + "_" + pieces[i + 2]
+            del pieces[i + 1:i + 3]
+    pieces = [orig] + pieces
+    return pieces
```
scopatz: My concern with this function is that it relies on the dbtype *name* to infer things about the C++ type *structure*.  As a silly foil, what if there was a "volume-length units" primitive type that happened to be called "VL_UNITS". If I am reading this right, this function would interpret this as variable length. The "VL_" should be interpreted as a convention only, and not a strict rule.  

I think it is much better to rely on the C++ type to find the sub C++ types and lookup all the potential dbtypes that share the sub-C++type.  You might loose the VL information only doing this, but I am not sure that is an issue yet.

scopatz:
Thanks @hodger! The refactor looks great!  I just have the one main concern about the `list_deps()` function.

At this point I think it is a good idea to try to get this integrated and actually generated the code.  The sooner we start compiling the generated code, the better :)  This has a few important steps I'll post in a minute.

scopatz:
**Steps to integrate into cyclus:**

1. move the current h5back.cc file into a template file, probably called `h5back.cc.in` or something.
2. remove the part of this file that we are generating, namely the Query() function body, replace it with an easily discoverable and unique token.  Maybe `#pragma cyclus H5Back::Query` is a good idea.
3. Instead of the main function printing out the generated code, have it read in `h5back.cc.in`, do a simple string replacement of `#pragma cyclus H5Back::Query` with the generated code. Write the filled in string back out as `h5back.cc`.
4. Modify the build system (all the CMakeLists.txt files) to automatically run the script as a subprocess each time cyclus is built.  There are other examples of running scripts in cyclus, so you'll might want to look around for how it is done.  CMake can be a pain, so if you get stuck, ask for help :)

src/hdf5_back_gen.py
hodger: The thing that confuses me is when we have the following:
```python
MAP_VL_STRING_X_READER =  """\
unsigned int itemsize = sizeof({t.sub[2].cpptype}) + CYCLUS_SHA1_SIZE;
jlen = col_sizes_[table][j] / itemsize;
{t.cpptype} x;
for (unsigned int k = 0; k < jlen; ++k) {{
    x[VLRead<{t.sub[1].cpptype}, {t.sub[1].dbtype}>(buf + offset + itemsize*k)] = *reinterpret_cast<{t.sub[2].cpptype}*>(buf + offset + itemsize*k + CYCLUS_SHA1_SIZE);
}}
{NO_CLOSE}\
"""
```
If we are generating sub types based upon C++ type, we know that {t.sub[1].dbtype} won't be accurate (since its dbtype in this case needs to be VL_STRING). VLRead would be taking in STRING, instead. Doesn't that pose a problem, since STRING is a completely different enum value?

scopatz: Ahh yes, it does.  Sorry. I think that the best solution then is to add a column to the dbtypes.json file. This should be similar to the canonical form, but use the dbtype name, instead. Here are some examples:

```
INT -> INT
VL_STRING -> VL_STRING
VECTOR_INT -> [VECTOR INT]
VECTOR_VL_STRING -> [VECTOR, VL_STRING]
VL_MAP_INT_FLOAT -> [VL_MAP, INT, FLOAT]
MAP_INT_VECTOR_FLOAT = [MAP, INT, [VECTOR, FLOAT]]
```

Note that with the template types, the first element of the list is not necessarily a valid type.  This will allow you to look up the appropriate subdbtype. In fact, this might make a better canonical form overall...

hodger: Okay, I will work on implementing this with the other changes. Thanks!

scopatz: Thanks!

rwcarlsen:
I'm concerned that this isn't actually recursively generating read/write code.  Adding a new data type still requires writing all the same c++ code as before - it just gets put in a python file as strings instead of being directly in a c++ file.  Proper recursive template handling read/write code generation could replace/eliminate a lot of the code here.  While I definitely support moving in that direction, I would prefer we wait until we can provide that additional robustness/utility before adding complexity to the build system.

@scopatz, @hodger Do you have a timeline for churning out a more generalized code-gen algo?

hodger:
@rwcarlsen I definitely agree that a more generalized algorithm is preferable, and that is something that I expect to be working on soon. Part of the goal of this particular project was to help me become familiar with Python and the Cyclus codebase, so @scopatz has been helping me in taking it one step at a time.

hodger:
@scopatz this code generates correctly in the hdf5_back.cc file, and the unit tests pass. next steps?

scopatz:
Woohoo! Thanks @hodger!

PR #1227, opened 2016-01-11
update gtest amalgamation

rwcarlsen:
This fixes this error:

```
cyclus_unit_tests: symbol lookup error: cyclus_unit_tests: undefined symbol: _ZNK7testing7Message9GetStringB5cxx11Ev
```

that I have been getting just recently when I try to run the unit tests after a successful compile.  Strange.  I've looked at the so's with nm and have also confirmed proper linking with ldd.  It must be some strange incompatibility with the c++ tricks gtest was doing with new c11 compiler details.  In any case, the new gtest amalg does not exhibit the problem.


opotowsky:
Thanks for the update, @rwcarlsen

PR #1228, opened 2016-01-12
Add optional explicit agent inventory tables

### LINKED ISSUE #1223: Store inventories at run time
see https://groups.google.com/d/topic/cyclus-dev/RxiOb3jUSbM/discussion

rwcarlsen:
Built on #1227 - so this will need a rebase after that goes in.  This is still a work in progress and needs tests.  Closes #1223.

Input file now has two new optional config params in control section:

```xml
<explicit_inventory>[bool]</explicit_inventory>
<explicit_inventory_compact>[bool]</explicit_inventory_compact>
```
which each generate new correspondingly named tables in the database.  There is one main design decision we still need to make:

* We have the ability to record inventories separately by buffer name - i.e. facilities with multiple buffers can have their different buffer inventories recorded separately (e.g. fresh_fuel, in_core, spent_fuel).  This is how I've implemented it here.  The alternative is to just squash all different buffers' materials together and record a single composition for the agent for the time step.  What do you all think?

### DIFF
--- a/share/cyclus-flat.rng.in
+++ b/share/cyclus-flat.rng.in
@@ -22,6 +22,12 @@ datatypeLibrary="http://www.w3.org/2001/XMLSchema-datatypes">
         <element name="dt"><data type="nonNegativeInteger"/></element> 
       </optional>
       <optional>
+        <element name="explicit_inventory"> <data type="boolean"/> </element>
+      </optional>
+      <optional>
+        <element name="explicit_inventory_compact"> <data type="boolean"/> </element>
+      </optional>
+      <optional>
         <element name="solver"> 
           <interleave>
             <optional><element name="config">
--- a/share/cyclus.rng.in
+++ b/share/cyclus.rng.in
@@ -23,6 +23,12 @@ datatypeLibrary="http://www.w3.org/2001/XMLSchema-datatypes">
         <element name="dt"><data type="nonNegativeInteger"/></element> 
       </optional>
       <optional>
+        <element name="explicit_inventory"> <data type="boolean"/> </element>
+      </optional>
+      <optional>
+        <element name="explicit_inventory_compact"> <data type="boolean"/> </element>
+      </optional>
+      <optional>
         <element name="solver"> 
           <interleave>
             <optional><element name="config">
--- a/src/context.cc
+++ b/src/context.cc
@@ -19,6 +19,8 @@ SimInfo::SimInfo()
       dt(kDefaultTimeStepDur),
       decay("manual"),
       branch_time(-1),
+      explicit_inventory(false),
+      explicit_inventory_compact(false),
       parent_sim(boost::uuids::nil_uuid()),
       parent_type("init") {}
 
@@ -30,6 +32,8 @@ SimInfo::SimInfo(int dur, int y0, int m0, std::string handle)
       decay("manual"),
       branch_time(-1),
       handle(handle),
+      explicit_inventory(false),
+      explicit_inventory_compact(false),
       parent_sim(boost::uuids::nil_uuid()),
       parent_type("init") {}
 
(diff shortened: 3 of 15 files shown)


rwcarlsen:
This has been rebased.  Just needs tests now.  I'll be working on other things for a while - I think @opotowsky was planning on picking up the test writing.

opotowsky:
Regarding whether to keep @rwcarlsen's implementation or to only record a single composition, I vote that we keep the current method of recording the inventories separately. Squashing can probably be a post-processing step.

opotowsky:
@rwcarlsen and @scopatz This follows the discussion on the issue https://github.com/cyclus/cyclus/issues/1223, but I just ran an EG23 sim on both versions of the inventories tables. The db with the compact table is 2GB and the db with the non-compact table is 0.5GB. For all iterations, I have:
No Inv: 248 MB
Compact Inv: 2.1 GB
Non-compact Inv: 523 MB

scopatz:
Maybe I am not understanding the usage of the word "compact" here...

rwcarlsen:
Compact just means the composition map (map<int,double>) is just dumped
outright as a single cell to the backend db to be handles however the
backend handles it (as per your preference for the inventory table).  It
just so happens that the boost-serialization that the sqlite backend uses
for putting arbitrary container types as an entry into a single row cell is
not particularly efficient (surprise :-).  Compact means db schema - not
necessarily data size.  I expect that HDF5 will be the reverse - with a
smaller compact table and a larger non-compact table.

There is a question we'd like your feedback on @scopatz - right now, as
implemented, the inventories of each resbuf are recorded separately.  I
(and others) have decided that this is probably useful enough to keep the
implementation this way.  Do you like it, or would you rather just have all
inventories for a particular agent combined into a single entry each time
step?

scopatz:
Just to clarify, the current implementation aggregates each resource buffer individually but does not aggregate all of the resources buffers together.  Thus each agent has N maps stored, where N is the number of resbufs, not 1 map stored.  The current implementation does not store each resource/material separately. Is this correct?

If so, I think that this is fine. How are the resource buffers for an agent disambiguate? Is there a column that also lists the name of the statevar for the resource buf?

rwcarlsen:
On Thu, Feb 4, 2016 at 11:00 AM, Anthony Scopatz <notifications@github.com>
wrote:

> Just to clarify, the current implementation aggregates each resource
> buffer individually but does not aggregate all of the resources buffers
> together. Thus each agent has N maps stored, where N is the number of
> resbufs, not 1 map stored. The current implementation does not store each
> resource/material separately. Is this correct?
>
Correct.

> If so, I think that this is fine. How are the resource buffers for an
> agent disambiguate? Is there a column that also lists the name of the
> statevar for the resource buf?
>
Yes - there is a string column that provides the name of each resbuf as
reported by the map<string,resbuf> returned from the SnapshotInv function -
so we get nice names meaningful to humans like "fiss", "spentfuel",
"freshfuel", etc.

scopatz:
Alright, this seems totally reasonable to me. +1

opotowsky:
I forgot to mention that I was using all sqlite dbs because I'm having hdf5 issues with cyclus again....

rwcarlsen:
The cycamore separations facility probabiliy still doesn't work with hdf5
(it didn't before).  Hopefully the code-gen stuff Anthony is overseeing
will change that soon.

rwcarlsen:
Tests merged in.  This should be ready for a final look by someone.

mbmcgarry:
Nosetests fails:
````ERROR: Failure: NameError (global name 'sqlite3' is not defined)
----------------------------------------------------------------------
  File "/Library/Python/2.7/site-packages/nose/loader.py", line 251, in generate
    for test in g():
  File "/Users/mbmcgarry/git/cyclus/tests/test_inventories.py", line 69, in test_inventories
    conn = sqlite3.connect(outfile)
NameError: global name 'sqlite3' is not defined

----------------------------------------------------------------------
`````

rwcarlsen:
I updated - can you try again?

mbmcgarry:
Looks good now @rwcarlsen !

PR #1230, opened 2016-02-02
Fix parametrized FacilityTests

rwcarlsen:
The FacilityTests initialization does not honor the normal initialization
guarantees for archetypes.  And so any type of remotely sophisticated
configuration validation of archetypes will be triggered to fail if we try to
call methods like e.g. Tick, Tock, Build, etc.  Remove these calls.  We
already know the agents satisfy the Agent class interface implicitly by being
able to load them and cast/assign them to Agent*/Facility*/etc.  From now on,
only call state-reporting functions with no side-effects in these tests.

### DIFF
--- a/tests/agent_tests/agent_tests.cc
+++ b/tests/agent_tests/agent_tests.cc
@@ -18,13 +18,11 @@ TEST_P(AgentTests, Clone) {
   delete clone;
 }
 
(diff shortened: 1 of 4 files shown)


mbmcgarry:
Thanks @rwcarlsen

PR #1231, opened 2016-02-09
fix mocksim to initialize timer via context InitSim

rwcarlsen:
so the context has the correct sim_info struct for agents to query.


mbmcgarry:
Thanks @rwcarlsen !

PR #1233, opened 2016-02-23
make sure db-dumped input file includes xinclude substitutions

rwcarlsen:


Baaaaam:
seems to be also good :)
Merging!!

PR #1234, opened 2016-02-23
be careful with exclusivity float comparisons in DRE

rwcarlsen:
Previously, some exclusivity arcs were being terminated (set to zero) because
floating point comparisons failed that should have succeeded such as:

    29565 <= 29564.999999999994

This uses ULP comparisons to determine closeness of floats more robustly
handling cases of numbers that should be treated as ~identical.

This resolves some issues Baptiste has been having in his FCO work.


Baaaaam:
seems to be ok!

merging

Issue #1236, opened 2016-03-11
Errors in hdf5_back_gen.py


gonuke:
I am getting the following error when I try to build cyclus on a clean Ubuntu system....  I know there are a lot of issues and discussion around HDF5 right now, but I didn't see anything about this...  @scopatz ? @hodger ?

```
-- Found '/root/cyclus/share/cyclus_nuc_data.h5'
Traceback (most recent call last):
  File "/root/cyclus/src/hdf5_back_gen.py", line 32, in <module>
    CANON_SET.add(convert_canonical(row[7]))
  File "/root/cyclus/src/hdf5_back_gen.py", line 28, in convert_canonical
    return tuple(convert_canonical(x) for x in raw_list)
  File "/root/cyclus/src/hdf5_back_gen.py", line 28, in <genexpr>
...
  File "/root/cyclus/src/hdf5_back_gen.py", line 28, in convert_canonical
    return tuple(convert_canonical(x) for x in raw_list)
  File "/root/cyclus/src/hdf5_back_gen.py", line 28, in <genexpr>
    return tuple(convert_canonical(x) for x in raw_list)
RuntimeError: maximum recursion depth exceeded
```

scopatz:
@hodger, can you take a look please?  Also @gonuke, I still cannot compile / link Cyclus on my main machine.  See #1232.

gonuke:
Note, these errors don't necessarily prevent the build from completing, but I wonder if they are related.  They happen early in the CMake process and may not have been noticed by others???

hodger:
I am going over this, but I can't seem to duplicate that error. When I attempt to install from develop, the hdf5_back.cc seems to build fine, but I get errors from the Google Test.

```
/home/ryan/cyclus/cyclus/tests/GoogleTest/gtest/gtest-all.cc:7833:1: error: ‘GTEST_ATTRIBUTE_NO_SANITIZE_ADDRESS_’ does not name a type
 GTEST_ATTRIBUTE_NO_SANITIZE_ADDRESS_
 ^
/home/ryan/cyclus/cyclus/tests/GoogleTest/gtest/gtest-all.cc: In function ‘pid_t testing::internal::ExecDeathTestSpawnChild(char* const*, int)’:
/home/ryan/cyclus/cyclus/tests/GoogleTest/gtest/gtest-all.cc:7900:57: error: ‘StackGrowsDown’ was not declared in this scope
     static const bool stack_grows_down = StackGrowsDown();
                                                         ^
[...]

PR #1237, opened 2016-03-23
Fix for hdf5_back_gen.py compatibility issues.

hodger:
This is in reference to #1236.

On my clean install of 14.04, this fix allows to cyclus to install successfully with no hdf5_back recursion errors. I sorted out unicode characters at the point when we pull in data from dbtypes.json. 

Additionally, this allows almost all unit tests to pass, except the following 3: 

```[----------] Global test environment tear-down
[==========] 553 tests from 109 test cases ran. (30570 ms total)
[  PASSED  ] 550 tests.```
```[  FAILED  ] 3 tests, listed below:```
```[  FAILED  ] ProgTranslatorTests.translation```
```[  FAILED  ] Hdf5BackTests.ReadWriteDouble```
```[  FAILED  ] Hdf5BackTest.ReadWriteAll```

In terms of the ReadWriteDouble, the generated double code looks to be correct.
```
case DOUBLE: {
        double x = *reinterpret_cast<double*>(buf + offset);
        is_row_selected = CmpConds<double>(&x, &(field_conds[qr.fields[j]]));
        if (is_row_selected)
            row[j] = x;
        break;
}
```

So I'm not sure what's going on there.

### DIFF
--- a/src/hdf5_back_gen.py
+++ b/src/hdf5_back_gen.py
@@ -1,12 +1,27 @@
-#!/usr/bin/python3
+#!/usr/bin/env python
 """This module generates HDF5 backend code found in src/hdf5_back.cc"""
 import os
 import json
-import textwrap
 from ast import literal_eval
 
+def resolve_unicode(item):
+    if isinstance(item, str):
+        return item.encode('utf-8')	
+    elif isinstance(item, tuple):
+        return tuple([resolve_unicode(i) for i in item])
+    elif isinstance(item, list):
+        return [resolve_unicode(i) for i in item]
+    else: 
+        #print("Reading in ", type(item))
+        try:
+            return item.encode('utf-8')
+        except Exception:
+            #print('could not encode ',type(item))
+            pass
+        return item
+
 with open(os.path.join(os.path.dirname(__file__), '..', 'share', 'dbtypes.json')) as f:
-    RAW_TABLE = json.load(f)
+    RAW_TABLE = resolve_unicode(json.load(f))
 
 TABLE_START = 0
 for row in range(0, len(RAW_TABLE)):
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by hodger** [comment]
@scopatz @gonuke, please take a look when you have a chance

**Comment by gonuke** [comment]
This works for me! Thanks @hodger 

I'm not sure how to interpret the first CI error, though? Is this something we've seen before @rwcarlsen and if so, why can't we make it go away?

**Comment by scopatz** [comment]
Thanks @hodger! Can you print the actual failure message for `Hdf5BackTests.ReadWriteDouble`.  It is farther up in stdout it you scroll up?

**Comment by gonuke** [comment]
I had all tests passing with this.  (although all the HDF5 tests fail on Ubuntu 15.10)

File: src/hdf5_back_gen.py [hunk 1]
```diff
@@ -23,12 +39,12 @@
 INDENT = '    '
 
 def convert_canonical(raw_list):
-    if isinstance(raw_list, str):
+    if type(raw_list) is str:
```
scopatz: I think that is instance is better in general.  Or was that causing the problem?  unicode is not a subtype of str, so the original line is fine.

src/hdf5_back_gen.py
```diff
@@ -485,6 +501,22 @@ def __init__(self, canon):
            'MAP_PAIR_INT_VL_STRING_DOUBLE': MAP_PAIR_INT_VL_STRING_DOUBLE_READER,
            'VL_MAP_PAIR_INT_VL_STRING_DOUBLE': VL_READER}
 
+def indent(text, prefix, predicate=None):
+    """Adds 'prefix' to the beginning of selected lines in 'text'.
+    If 'predicate' is provided, 'prefix' will only be added to the lines
+    where 'predicate(line)' is True. If 'predicate' is not provided,
+    it will default to adding 'prefix' to all non-empty lines that do not
+    consist solely of whitespace characters.
```
scopatz: Did you copy this from somewhere? if so you should make a note in the docstring.

hodger: Oh yes, sorry, this is the code from textwrap because indent was not implemented in Python2.7. My mistake

hodger:
@gonuke I got 33 failed tests after (successfully) installing on a clean 15.10. 
the gist: https://gist.github.com/hodger/3f31ef28a2ffd786aa00
I don't yet understand what is causing the failures. The generated code is the same as it was.

hodger:
However, (@scopatz) the error with the HDF5 Double is the same one I got in my 14 installation, so if you're interested, that entire error is in the gist above.

scopatz:
Ok that is a legit error.

hodger:
Update: so @scopatz and I looked into all of this today and came to the conclusion that the Read-Write errors have to do with the HDF5 version used, not the code generated by hdf5_back_gen.py. The Double error is caused most likely by the use of a 32 bit machine, and stopped appearing when I moved to a clean 64 bit machine. 

On a fresh install of 15.10, with HDF5 version 1.8.13, (with this fix) Cyclus installs successfully, and all tests pass. We will need to look into whether 1.8.16 will allow the tests to pass as well.

scopatz:
Also 30 of the errors are clearly coming from hdf5 v1.8.15-patch1.

hodger:
@scopatz, @gonuke thoughts on merging this in since it fixes the python problem (and since hdf5_back_gen is not the source of our lovely hdf5 issues)?

scopatz:
I am happy to merge this, but I don't understand where this failing check is coming from. Any ideas @gonuke @Baaaaam @mbmcgarry @rwcarlsen @gidden or @katyhuff?

rwcarlsen:
The failure is from the old polyphemus CI hook that keeps resurecting itself.  I've removed/deleted the hook from this repository like 3 times in the past already.  How do you kill the undead?

scopatz:
Uggh.  Well I just tried removing it again. Thanks for confirming @rwcarlsen

scopatz:
Thanks @hodger!

PR #1238, opened 2016-04-07
read CYCLUS_NO_EXCEPTION env variable to know if catching exception is require

Baaaaam:
As @rwcarlsen asked, with this modification, one can now remove the catch of exception  by setting CYCLUS_NO_EXCEPTION environment variable to anything but 0.

Not sure the choose of the variable name is the best... I let you guys suggest something better

### DIFF
--- a/cli/cyclus.cc
+++ b/cli/cyclus.cc
@@ -196,17 +196,17 @@ int main(int argc, char* argv[]) {
   }
 
   
-  char* CYCLUS_NO_EXCEPTION = getenv("CYCLUS_NO_EXCEPTION");
-  if( CYCLUS_NO_EXCEPTION !=NULL && CYCLUS_NO_EXCEPTION != "0" ){
+  char* CYCLUS_NO_CATCH = getenv("CYCLUS_NO_CATCH");
+  if( CYCLUS_NO_CATCH !=NULL && CYCLUS_NO_CATCH != "0" ){
     si.timer()->RunSim();
-  }
-  else {
+  }   else {
     try {
       si.timer()->RunSim();
     } catch (cyclus::Error err) {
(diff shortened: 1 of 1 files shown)


cli/cyclus.cc
```diff
@@ -195,13 +195,19 @@ int main(int argc, char* argv[]) {
     si.recorder()->RegisterBackend(fback);
   }
 
-  try {
+  
+  char* CYCLUS_NO_EXCEPTION = getenv("CYCLUS_NO_EXCEPTION");
```
rwcarlsen: Rename env var to `CYCLUS_NO_CATCH_EXCEPTION` or maybe just `CYCLUS_NO_CATCH`.

cli/cyclus.cc
```diff
@@ -195,13 +195,19 @@ int main(int argc, char* argv[]) {
...
+  if( CYCLUS_NO_EXCEPTION !=NULL && CYCLUS_NO_EXCEPTION != "0" ){
     si.timer()->RunSim();
-  } catch (cyclus::Error err) {
-    std::cerr << err.what() << "\n";
-    return 1;
   }
-
+  else {
```
rwcarlsen: remove line break - i.e. should say `} else {`

cli/cyclus.cc
```diff
@@ -195,13 +195,19 @@ int main(int argc, char* argv[]) {
...
+    try {
+      si.timer()->RunSim();
+    } catch (cyclus::Error err) {
+      std::cerr << err.what() << "\n";
+      return 1;
+    }  }
```
rwcarlsen: insert line break (keeping proper nesting indentation).

PR #1253, opened 2016-05-17
Fix for anonymous namespaces, triggered by boost

scopatz:
On boost v1.60.0, there are some terrible anonymous namespaces that are hidden by a macro call stack. In particular in `boost/type_traits/detail/has_postfix_operator.hpp`, we have effectively, 

```C++
namespace boost {
namespace detail {

// This namespace ensures that argument-dependent name lookup does not mess things up.
namespace BOOST_JOIN(BOOST_TT_TRAIT_NAME,_impl) {

struct no_operator { };
}
}
}
```

Where the `BOOST_JOIN()` macro is defined as, 

```C++
#define BOOST_JOIN( X, Y ) BOOST_DO_JOIN( X, Y )
#define BOOST_DO_JOIN( X, Y ) BOOST_DO_JOIN2(X,Y)
#define BOOST_DO_JOIN2( X, Y ) X##Y
```

This was causing cycpp to fail with the following error:

```console
  File "/home/scopatz/miniconda3/conda-bld/work/cyclus-cd572972b3dbf4b6d24a8e500ff270a42e8b2c7d/cli/cycpp.py", line 2317, in <module>
    main()
  File "/home/scopatz/miniconda3/conda-bld/work/cyclus-cd572972b3dbf4b6d24a8e500ff270a42e8b2c7d/cli/cycpp.py", line 2301, in main
    context, superclasses = accumulate_state(canon)   # pass 2
  File "/home/scopatz/miniconda3/conda-bld/work/cyclus-cd572972b3dbf4b6d24a8e500ff270a42e8b2c7d/cli/cycpp.py", line 861, in accumulate_state
    state.accumulate(statement, sep)
  File "/home/scopatz/miniconda3/conda-bld/work/cyclus-cd572972b3dbf4b6d24a8e500ff270a42e8b2c7d/cli/cycpp.py", line 713, in accumulate
    filter.transform(statement, sep)
  File "/home/scopatz/miniconda3/conda-bld/work/cyclus-cd572972b3dbf4b6d24a8e500ff270a42e8b2c7d/cli/cycpp.py", line 367, in transform
    classname = state.classname()
  File "/home/scopatz/miniconda3/conda-bld/work/cyclus-cd572972b3dbf4b6d24a8e500ff270a42e8b2c7d/cli/cycpp.py", line 685, in classname
    return "::".join(names)
TypeError: sequence item 2: expected str instance, NoneType found
```

because the particular `BOOST_JOIN` happened to evaluate to an empty string, creating an anonymous namespace, and thus the cycpp `self.namespaces == ['boost', 'detail', None, 'no_operator']`.

This PR replaces the None with `'<anonymous>'`. This will allow cycpp to skip such crazy cases, while also allowing it to fail at compile time if anyone tries to code generate to a class in an anonymous namespace (which is not allowed).

@hodger or @FlanFlanagan, would you mind taking a look at this and merging it.  I need this to proceed with conda-forge packages.

### DIFF
--- a/cli/cycpp.py
+++ b/cli/cycpp.py
@@ -111,7 +111,7 @@ WRANGLERS = {
     'mi6::Spy',  # for testing!!
     }
 
-ENTITIES = [('cyclus::Region', 'region'), ('cyclus::Institution', 'institution'), 
+ENTITIES = [('cyclus::Region', 'region'), ('cyclus::Institution', 'institution'),
             ('cyclus::Facility', 'facility'), ('cyclus::Agent', 'archetype')]
 
 def escape_xml(s, ind='    '):
@@ -217,12 +217,12 @@ class Filter(object):
         elif varname is None:
             s = ("The {0!r} class in the {1} machine has the following "
                  "annotations:\n{2}")
-            s = s.format(classname, mc.__class__.__name__, 
+            s = s.format(classname, mc.__class__.__name__,
                          pformat(mc.context[classname]))
         else:
             s = ("The {0!r} state variable on the {1!r} class in the {2} machine "
                  "has the following annotations:\n\n{3}\n")
-            s = s.format(varname, classname, mc.__class__.__name__, 
+            s = s.format(varname, classname, mc.__class__.__name__,
                          pformat(mc.context[classname]['vars'][varname]))
         return s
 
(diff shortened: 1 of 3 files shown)


scopatz:
Also, this PR makes it so that cmake will fail immediately if cycpp fails. Probably long overdue.

scopatz:
Fixed test failure

rwcarlsen:
Awesome work @scopatz.  The cmake immediate fail in particular will be very helpful.

PR #1256, opened 2016-05-26
fail fast on more cmake EXECUTE_PROCESS calls

### LINKED ISSUE #1249: No error if subproc fails during build
Commands like `EXECUTE_PROCESS()`, such as used [here](https://github.com/cyclus/cyclus/blob/develop/src/CMakeLists.txt#L8) do not cause CMake to fail, thus causing extraneous errors in testing or building down the road.  These should fail ASAP.

rwcarlsen:
fixes #1249


Baaaaam:
looks good

PR #1258, opened 2016-06-03
Map string map string int

mbmcgarry:
This datatype is needed for some of my mbmore archetypes


scopatz:
Do you have any tests for this, @mbmcgarry?

scopatz:
Thanks for putting this in!

rwcarlsen:
I think we should decide some conventions for db type tests moving forward. - Since we use code generation to support these types (and boost serialization for sqlite), I think it probably shouldn't be necessary to keep testing every new type - because that just continues to retest the same proven code-generation from all the other tested container types.  I think we have or will soon reach type nesting level that demonstrates the code-generation for backends is working well enough.

scopatz:
Code generation can fail, just like any other piece of code.  I think the best way to ensure that there are not errors with a given data type and backend is to have tests for that type and backend.  This also ensures that if the code generator needs to change, the data types will still work.

rwcarlsen:
Fair enough - but I think it would be sufficient (at least for the sqlite side) to just check storing the type and retrieving it - no need to check for explicit values.  If code generation didn't work correctly, you would get an exception during store or retrieve.  The only way I can imagine other failures would be if boost serialization wasn't working right, and we don't need to be testing dependencies.

So maybe a test like this:

```c++
TEST_F(SqliteBackTests, MapStrDouble) {
     typedef std::map<std::string, std::map<int, std::vector<int> > > Foo;
     
     Foo f;
     r.NewDatum("monty")
         ->AddVal("python", f)
         ->Record();
     r.Close();
     cyclus::QueryResult qr = b->Query("monty", NULL);
     EXPECT_NO_EXCEPTION(f = qr.GetVal<Foo>("python"));
}

```

rwcarlsen:
And @mbmcgarry - for the future, we could maybe add a comment like this to the sqlitebacktests file:

```c++
// For future types, since the code generation and serialization has been
// sufficiently proven with existing tests, just use the following template -
// only adjusting the typedef at the top:
//
//     TEST_F(SqliteBackTests, MapStrDouble) {
//       typedef std::map<std::string, std::map<int, std::vector<int> > > Foo;
//     
//       Foo f;
//       r.NewDatum("monty")
//           ->AddVal("python", f)
//           ->Record();
//       r.Close();
//       cyclus::QueryResult qr = b->Query("monty", NULL);
//       EXPECT_NO_EXCEPTION(f = qr.GetVal<Foo>("python"));
//     }

```

PR #1261, opened 2016-07-15
Correct the way sell_mat_policy (SMP) was handling the Pop() on ResBuf

Baaaaam:
Before SMP was Poping mat from ResBuf using Pop(qty,eps) but using a relative eps of 1e-12.

As [discussed](https://groups.google.com/forum/#!topic/cyclus-dev/d25nGI9dYH0), I change it to absolute value for eps: `cyclus:eps_rsrc()`


scopatz:
Looks good to me! Thanks!

PR #1262, opened 2016-07-20
Cbc 2.9.x interface, prevents segfault

scopatz:
Updating the OsiCbc interface to the latest version prevents segfaults in the unit tests on my laptop.  This took me months (off and on) to figure out.

PR #1264, opened 2016-07-22
Integration test fixes

scopatz:
This fixes the integration tests on my laptop back to the point where only the symbol check gives the normal failure (and not an error).  This builds on #1262, which is also needed for me.

It is really just a minor Python2/3 bytes issue and skipping the stub tests when they assume too much about the host system (like default args on the install script, or that it has a compiler).

### DIFF
--- a/release/smbchk.py
+++ b/release/smbchk.py
@@ -1,5 +1,5 @@
 #! /usr/bin/env python
-"""Collects & diffs public symobls in libcyclus.so. Used to ensure stability 
+"""Collects & diffs public symobls in libcyclus.so. Used to ensure stability
 between versions. Now with 100% fewer vowels!
 
 The following tasks may be useful:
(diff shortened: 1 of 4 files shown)


scopatz:
OK, now that #1262 has been merged, it would be great if this could be reviewed! Thanks!

scopatz:
ping @FlanFlanagan can you please take a look?

scopatz:
ping anyone...

gonuke:
I'm reviewing now - it builds fine on my system.  Any thoughts on how to force it to fail in a way that exposes the `SkipTest` addition?

scopatz:
Thanks @gonuke!  If you remove the compiler from your PATH that should force the test to skip

gonuke:
Removing the compiler caused other problems... 

I mangled the build_dir in the install cmd to make it fail

Thanks @scopatz

scopatz:
Thanks @gonuke!

Issue #419, opened 2012-10-29
Add Units to @param definitions in Enrichment.h


katyhuff:
Leftover documentation issues in pull request #415 were not dealt with in the interest of expediency. However, in the interest of well informed user/developers down the road, many of the `@param` and `@return` directives in the documentation of the Enrichment.h utility are lacking units where units may be necessary. That is, directives such as the following all deserve units to be specified where applicable (atom percent, mass percent, kg, g, etc.) : 

```
@param product_qty the amount of product Uranium
@param assays the assay of product, feed, and tails
@return the quantity of tails resulting from enriching the product
```

PR #1268, opened 2016-08-11
[Fix proposition 1/2] fix proposition cycamore issue #419

Baaaaam:
This is a soft proposition to solve `cycamore` issue [#419](https://github.com/cyclus/cycamore/issues/419#issuecomment-239237669).

This just change the `cyclus` in the INSTALL header path by `${lib_name}`, which is the name used actually to install other cyclus modules headers.

The other solution is simply a suppression of the HEADER installation instruction from the `UseCyclus.cmake` file, as it is also present in the src/CMakefile.txt of Cycamore or the other cystub-ed modules....
Please see for the PR #1269 for the second fixing proposition


Baaaaam:
As @gonuke suggest the third solution I close this one (which can be re-open if require)

PR #1269, opened 2016-08-11
[Fix proposition 2/2] fix proposition cycamore issue #419 

Baaaaam:
This is a "hard" proposition to solve cycamore issue [#419](https://github.com/cyclus/cycamore/issues/419#issuecomment-239237669).

This solution is simply a suppression of the HEADER installation instruction from the UseCyclus.cmake file, as it is also present in the src/CMakefile.txt of Cycamore or the other cystub-ed modules....

The other proposition change the cyclus in the INSTALL header path by ${lib_name}, which is the name used actually to install other cyclus modules headers.

Please see for the PR #1268 for the first fixing proposition


Baaaaam:
As @gonuke suggest the third solution I close this one (which can be re-open if require)

PR #1270, opened 2016-08-12
Fix header install path: addressing cycamore issue #419

Baaaaam:
This is an alternative solution to cycamore issue [#419](https://github.com/cyclus/cycamore/issues/419#issuecomment-239237669) from #1268 and #1269.

This solution is simply a suppression of the HEADER installation instruction from the UseCyclus.cmake file, as well as the src/CMakefile.txt of the cystub....

Please see for the PR #1268 and #1269 for the other fixing proposition


Baaaaam:
anyone ?

mbmcgarry:
thanks @Baaaaam !

PR #1271, opened 2016-08-18
Add variable for new cycamore::Mixer api

Baaaaam:
this add a `std::vector<std::pair<std::pair<double, double>, std::vector<std::pair<std::string, double> > > >` type into cyclus.

this is require for the new cycamore::mixer API in order to fix: cycamore issue [#421](https://github.com/cyclus/cycamore/issues/421)

this will be followed shortly by a PR on Cycamore.


tests/sqlite_back_tests.cc
```diff
@@ -41,6 +41,17 @@ class SqliteBackTests : public ::testing::Test {
 //       EXPECT_NO_THROW(f = qr.GetVal<Foo>("python"));
 //     }
 
+TEST_F(SqliteBackTests, VecPairPairDoubleDoubleVecPairStringDouble) {
```
mbmcgarry: Shouldn't this test be named 
VecPairPairDoubleDoubleMapStringDouble ?

Baaaaam: yes you are right.
it is corrected.

mbmcgarry:
thanks @Baaaaam !

PR #1272, opened 2016-08-18
small fix on query_backend

Baaaaam:
If I understood correctly how it should be implemented those corrections were needed...

I don't know if it matter or not....


gonuke:
Can you point me to the documentation that supports this change?  I can't find it.

Baaaaam:
I did not find any documentation, I just try to understand how to do it, accordingly to the other similar definition.
As far as I understand, the number in comment should be incremented each time a `vector`, a `pair`, a `map`, a `string`, a `list` appear in the type definition.
But I don't know why.

gonuke:
@scopatz @rwcarlsen @gidden - are these comments parsed by `cycpp`?  Or something else? Is there documentation for how to report the `shape` correctly?

scopatz:
That number is not the shape, but the shape rank documented here: http://fuelcycle.org/arche/dbtypes.html

It is the maximum length of the shape, documented here: http://fuelcycle.org/arche/cycpp.html#id1

scopatz:
Also, these comments are parsed by the tool that creates the json types file.  @hodger wrote that, I believe.

gonuke:
Since the documentation is not entirely clear to me with regards to maps, I think this explains why this change is correct

I think the following is true for `MAP_STRING_MAP_STRING_INT
* a MAP_A_B has a rank of 1 for simple types A & B
* a MAP_A_STRING  has a rank of 2 for simple type A; the variable length of strings adds a rank
* a MAP_A_MAP_B_C has a rank of 2 for for simple types A, B, C because MAP_B_C has one dimension and MAP_A_(MAP...) has another dimension
* a MAP_STRING_MAP_B_C has a rank of 3 for simple types B,C: 1 for the first MAP, 1 for the STRING and 1 for the 2nd MAP
* a MAP_STRING_MAP_STRING_C has a ranks of 4: 1 for each of the MAPs and 1 for each of the STRINGS

Do we have a test that would indicate failure when this is wrong (as it is currently)?

gonuke:
I guess that the only thing that relies on this is `cymetric` (??) so that's where we'd first see failure?

PR #1273, opened 2016-08-24
More Horrifying Fixes

scopatz:
After more all-consuming weeks of my life wasted #notbitter, I finally have a version of Cyclus that is building & testing on conda-forge on Linux and Mac.  See the end of conda-forge/staged-recipes#678 for proof.  

The problems this PR fixes are  mostly, but not entirely, Coin's fault. The issues are as follows:

1) CoinClp was not being linked in explicitly at all, previously, even though it was required.

2) OsiCbc was again not being linked in, even though this is the primary interface we use.

3) Coin libraries were not being linked in in strict dependency order. This could cause undefined symbol errors at link time on more pedantic systems (ie clang)

4) SQLite headers could be found that were separate and associated with the SQLite shared library that was found. This was likely due to some subtlety or bug or difference between CMake's `find_path()` (used for header) and `find_library()`.  Now, the header is forced to match the library.

5) Even though CMake would *build* out of core, the Cyclus setup was not capable of installing fully out of core.  That is, libcyclus.so would always link to some files in the core OS file system somewhere. The exceptions to this were the few packages (Coin, HDF5, and Boost) that we had specific 'root' targets for.  This made it impossible to build Cyclus in conda in a way that would be relocatable to other systems.  The fix that this PR implements is a `$DEPS_ROOT_DIR` environment variable that acts as a base hint for all dependencies. This was then applied through out the code base. The Coin, HDF5, Boost, etc roots will override the DEPS value for that package alone if both are provided.

6) Integer programs are just broken. At least on CoinCbc v2.9 with our OsiCbcInterface wrapper. However, outside of some tests, integer programs are not actually used in the core explicitly by default.  It is not clear that users can turn them on, even.  

The **worst** part about this isn't just that Coin gives the wrong answer.  No.  The worst part is that when Coin fails because a program simply has an integer variable it executes an assert statement.  Asserts in C++ cannot be caught by exception handling and issue a SIGABRT, terminating the program. I am pretty sure this kind of destructive behaviour in a library is against the style guide - though maybe since we are programmatically  calling a CLI, we were asking for this kind of trouble.

The solution implemented here adds an environment variable `$ALLOW_MILPS` that declares whether integer-containing programs may be used.  A default may be selected at compile time, depending on whether you know it will work or not.  The `Env::allow_milps()` member function is then consulted in the SIGABRT prone areas as a substitute for exception handling.

7) The OsiCbcSolverInterface is missing some of its members in the upstream Mac package. Unfortunately, these members are declared pure virtual on the OsiSolverInterface base class and must be provided by link time. Most of these have trivial implementations - dropping a name argument - which we must now provide. 

@FlanFlanagan - and anyone else who wants - can you please take a look?


gidden:
Hey @scopatz, good work slaying the beast that is the coin suite. A few notes from what I remember (or don't):

- I am not sure why the Clp library was not being explicitly linked, though in all our previous CI and local builds, this did not seem to be an issue

- I am mostly certain that OsiCbc is [comprised of only the header/cpp file](https://github.com/coin-or/Cbc/tree/master/Cbc/src/OsiCbc) which we build as part of cycluscore, thus was not needed to be explicitly linked in the coin cmake. Of course my memory could have faltered here -- you likely know better than I what is required.

- I vaguely remember the Osi interface missing members previously -- probably when we were still on polyphemus and had mac access, which is why I think I copied it into the code base in the first place... it was so long ago, apologies for not remembering earlier.

- Integer programs do work with the CI as currently set up -- I doubt CI is on bleeding-edge coin, though. I am on 2.8.7 on my system, for example, and have extensively tested/used cbc on it. For posterity, using COIN solvers for the DRE is documented [here](http://fuelcycle.org/user/input_specs/control.html).

Also it looks like this is failing because of a missing cmake file for `libxml2`..

gidden:
If you are having really bad problems with 2.9, perhaps we should freeze the dependency at the end of the 2.8.7 line? Until someone has time and energy to debug more thoroughly? Separately, `ALLOW_MIPS` could be set to 0 by default if 2.9 is encountered with a loud warning during configure time? Just some thoughts..

Otherwise, this looks good to me, though I will defer to @FlanFlanagan

FlanFlanagan:
I'm having the libxml2-config.cmake issue for sure. Trying to find a work around for that now.

scopatz:
Hey @gidden, 

> I am not sure why the Clp library was not being explicitly linked, though in all our previous CI and local builds, this did not seem to be an issue

I think it isn't always available. At least not always on Ubuntu. This is why it is skipped.  However, in the conda package it is present and I think it is needed or desirable to link in.

> I am mostly certain that OsiCbc is comprised of only the header/cpp file which we build as part of cycluscore, thus was not needed to be explicitly linked in the coin cmake. Of course my memory could have faltered here -- you likely know better than I what is required.

My "knowing" here is mostly empirical. I believe that it wasn't working on Mac without it.

> I vaguely remember the Osi interface missing members previously -- probably when we were still on polyphemus and had mac access, which is why I think I copied it into the code base in the first place... it was so long ago, apologies for not remembering earlier.

No worries! The fault clearly lies with the Coin devs here since they are shipping an incomplete type.

> Integer programs do work with the CI as currently set up -- I doubt CI is on bleeding-edge coin, though. I am on 2.8.7 on my system, for example, and have extensively tested/used cbc on it. For posterity, using COIN solvers for the DRE is documented here.

Right, they don't always fail.  Just on v2.9.x for whatever reason.

> If you are having really bad problems with 2.9, perhaps we should freeze the dependency at the end of the 2.8.7 line? Until someone has time and energy to debug more thoroughly? Separately, ALLOW_MIPS could be set to 0 by default if 2.9 is encountered with a loud warning during configure time? Just some thoughts..

I think freezing at 2.8.7 would be a mistake - which is why I have poured so much time and effort into getting *something* working.  Future releases hold three basic possibilities:

1. This issue will be fixed, in which case freezing did us no good.
2. The issue will remain broken, in which case freezing is indefinite. In this scenario I believe it would be much more likely that Coin wold eventually be removed as a dependency than anyone would update it and investigate what is wrong. In this case, why wait on removing it?  Chatting with @gonuke in Paris, it seemed like removal was an unacceptable option in the present, even though it would have been the easiest.
3. Other parts of Coin will break.  Freezing will then simply compound and confound any attempts to update later on.

Freezing dependencies (if you'll allow the editorialization) is one of the bigger problems with scientific applications. It is a fast track to turning a package into "legacy code." I don't think that Cyclus can afford to be legacy at this point.

> Also it looks like this is failing because of a missing cmake file for libxml2..

Yep. I figured this out and it looks like CI here is now passing.  @FlanFlanagan, mind taking another look.

PR #1277, opened 2016-08-25
Ensure that cyclus and cycamore install.py files are the same

scopatz:
Relatively minor change so that the input files are exactly the same.

### DIFF
--- a/install.py
+++ b/install.py
@@ -72,6 +72,8 @@ def install_cyclus(args):
             cmake_cmd += ['-DCOIN_ROOT_DIR=' + absexpanduser(args.coin_root)]
         if args.boost_root:
             cmake_cmd += ['-DBOOST_ROOT=' + absexpanduser(args.boost_root)]
+        if args.cyclus_root:
+            cmake_cmd += ['-DCYCLUS_ROOT_DIR='+absexpanduser(args.cyclus_root)]
         if args.hdf5_root:
             h5root = absexpanduser(args.hdf5_root)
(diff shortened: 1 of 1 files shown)


FlanFlanagan:
Worked for me. I'll merge in the morning in case anyone else wants to take a look.

scopatz:
Thanks @FlanFlanagan

Baaaaam:
I'll take a look this morning...

install.py
```diff
@@ -171,6 +173,9 @@ def main():
     hdf5 = "the path to the HDF5 libraries directory"
     parser.add_argument('--hdf5_root', help=hdf5)
 
+    cyclus = "the relative path to Cyclus installation directory"
+    parser.add_argument('--cyclus-root', '--cyclus_root', help=cyclus)
```
Baaaaam: Why do we want to provide `cyclus` relative path to `cyclus` installer ?

scopatz: That is ancillary. We just want the install.py files to be the same everywhere. It is too much work to maintain multiple versions, and so the version we get is the superset of all required features.

FlanFlanagan:
I'm going to let @Baaaaam merge this one. Not sure if we wants to fix the relative path thing or not.

gonuke:
If these files are identical across cyclus & cycamore repos, then it will be great if we can figure out how to maintain a single copy! (for another time)..... merging.

Thanks @scopatz

scopatz:
Thanks @gonuke!

PR #1278, opened 2016-09-07
Fix abi

Baaaaam:
fix the 2 ABI issue not induced by the c++11 change


scopatz:
Seems reasonable to me!  Thanks @Baaaaam!

scopatz:
Thanks @Baaaaam!

PR #1280, opened 2016-09-09
revert the change in prog_translator.h

Baaaaam:
I don't know what will be the final decision, but at least this is ready if require.

### DIFF
--- a/src/prog_translator.h
+++ b/src/prog_translator.h
@@ -12,6 +12,15 @@ namespace cyclus {
 class ExchangeGraph;
 class ExchangeNodeGroup;
 
+/// @brief struct to hold all problem instance state
+struct ProgTranslatorContext {
+  std::vector<double> obj_coeffs;
+  std::vector<double> row_ubs;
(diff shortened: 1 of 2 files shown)


src/prog_translator.h
```diff
@@ -27,14 +36,7 @@ class ExchangeNodeGroup;
...
   /// @brief struct to hold all problem instance state
-  struct Context {
-    std::vector<double> obj_coeffs;
-    std::vector<double> row_ubs;
-    std::vector<double> row_lbs;
-    std::vector<double> col_ubs;
-    std::vector<double> col_lbs;
-    CoinPackedMatrix m;
-  };
+  typedef ProgTranslatorContext Context;
```
scopatz: For what it is worth, it is this typedef that is incorrect and should be a struct whose constructor throws an error.

scopatz: Note that this typdef, with the rest of this changeset is not used anywhere else in the code base.

Baaaaam:
closing this, as #1281 do it better.

PR #1284, opened 2016-09-15
checking in cystubs for cpp11 and enable it

Baaaaam:


stubs/CMakeLists.txt
```diff
@@ -14,6 +14,16 @@ ENDIF()
 
 PROJECT(@lib_name@)
 
+# check for and enable c++11 support
+INCLUDE(CheckCXXCompilerFlag)
+CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
+IF(COMPILER_SUPPORTS_CXX11)
+  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
```
scopatz: I don't believe that this is fully valid on Mac where the switch is `-stdlib=libc++`.  See what conda-forge does for more details: https://github.com/conda-forge/toolchain-feedstock/blob/master/recipe/activate.sh

Baaaaam: > Getting Started
> 
> If you already have libc++ installed you can use it with clang.
> 
> $ clang++ -stdlib=libc++ test.cpp
> $ clang++ -std=c++11 -stdlib=libc++ test.cpp
> On OS X and FreeBSD libc++ is the default standard library and the -stdlib=libc++ is not required.


from: http://libcxx.llvm.org/docs/UsingLibcxx.html, this is not true anymore.

scopatz: Yes, I know that you may use the switch that you specify here on Mac.  That isn't what I was saying though. 

My claim is that when everything else that you try to link against uses `-stdlib=libc++` but an object was compiled with `-std=c++11`, the linking will fail. The ABIs are actually different for these two and they are not interchangeable. 

Conda-forge has decided that `-stdlib=libc++` is the standard (see the toolchain link) across all packages. Going against this will likely break the conda package.  If you would like to show that it doesn't, please open up a PR against conda-forge/cyclus-feedstock that pulls in this changeset to show that it doesn't.  Otherwise, I feel it is far too risky to merge this change.

Baaaaam: OK i'll do it, but it should be ok, as the `CMakeList.txt` of `Cyclus` already include this. (this is the `cycstub` one)

Baaaaam: @scopatz I am willing to PR this changeset into `conda-forge/cyclus-feedstock`. But I am not sure to understand how :s.

Secondly, What is was trying to highlight with my previous quote is that` OS X use by default libc++ and the  -stdlib=libc++ is not required`. (There is maybe something I did not get from you reply after)

gonuke: does `-stdlib=libc++` guarantee c++11?

Baaaaam: what is the status there ?
@scopatz do you still want me to this changeset into `conda-forge/cyclus-feedstock` ? If so, please tell me more, I don't understand what I should change, and PR.


btw, I am not sure this is relevant as `-stdlib=libc++` is set by default on MAC when using apple clang ):
> On OS X and FreeBSD libc++ is the default standard library and the -stdlib=libc++ is not required.
from: http://libcxx.llvm.org/docs/UsingLibcxx.html.

so this should fulfill the condo-forge requirement even without the `-stdlib=libc++` flag

Baaaaam:
btw this I think should be PR into the `release` branch and not the `develop` branch.
So when we agree to this change I can close it and reopen it into `release` to be merged...

(I know it does not matter a lot, but it will prevent the develop branch to be "on hold"

mbmcgarry:
On the dev call we resolved the review discussion. Thanks @Baaaaam !

PR #1285, opened 2016-09-20
more robust core version handling

scopatz:
This extends the CMake version handling to extend beyond the case where the full git repo and git binary are available, such as in certain conda recipes or when building directly from a tarball.  This adds the `CORE_VERSION` variable to the build system and allows it to be set via `install.py --core-version=X.X.X`

Proof that this works is available in https://github.com/conda-forge/cyclus-feedstock/pull/3, which was failing on the 1.4.0-rc3 tag but works with my branch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,12 +271,19 @@ IF(NOT CYCLUS_DOC_ONLY)
         "${Boost_INCLUDE_DIR}"
         "${COIN_INCLUDE_DIRS}")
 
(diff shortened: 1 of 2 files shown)


gonuke:
Thanks @scopatz

gonuke:
Should this be PR'ed into the release branch?  Or should I just merge to both?

scopatz:
Hi @gonuke, I think it should be merged to both.

gonuke:
Thanks @scopatz . I'll now merge to release as well

PR #1286, opened 2016-09-21
Fix for python2

gonuke:
I encountered errors with incorrect syntax for python2 here.

```
  File "/home/wilsonp/UW/research/software/cyclus/cyclus/src/hdf5_back_gen.py", line 1093, in <module>
    main()
  File "/home/wilsonp/UW/research/software/cyclus/cyclus/src/hdf5_back_gen.py", line 1082, in main
    CPPGEN = CppGen()
  File "/home/wilsonp/UW/research/software/cyclus/cyclus/src/hdf5_back_gen.py", line 146, in __init__
    super().__init__(tree=tree)
TypeError: super() takes at least 1 argument (0 given)
```

I'm not sure why we haven't noticed before, but it scrolls past quickly and doesn't cause failure.  Perhaps our build process should fail when this step fails - I think errors at this step has been an issue before.

I need someone else to test on Python3 ( @hodger , @scopatz )

### DIFF
--- a/src/hdf5_back_gen.py
+++ b/src/hdf5_back_gen.py
@@ -121,7 +121,7 @@ class PrettyFormatter(Visitor):
     """Formats a tree of nodes into a pretty string"""
 
     def __init__(self, tree=None, indent=' '):
-        super().__init__(tree=tree)
+        super(PrettyFormatter, self).__init__(tree=tree)
         self.level = 0
         self.indent = indent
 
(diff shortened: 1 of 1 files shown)


scopatz:
Yes our build system should fail on error of `hdf5_back_gen` and it does not.  This looks correct but I will test it now.

scopatz:
Thanks @gonuke! And for what it is worth, I think we should should move away from Python 2 sometime soon.

gonuke:
I agree about moving away from Python2, but many systems still have it as default and maintaining 2 python versions frightens me :)

PR #1289, opened 2016-10-04
Added cost function pointer to request.

scopatz:
Thanks @FlanFlanagan!

### DIFF
--- a/src/request.h
+++ b/src/request.h
@@ -28,7 +28,7 @@ template <class T> class RequestPortfolio;
 template <class T>
 class Request {
  public:
-  typedef double (*cost_function_t)(boost::shared_ptr<Material>);
+  typedef double (*cost_function_t)(boost::shared_ptr<T>);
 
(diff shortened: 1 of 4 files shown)


scopatz:
ping!

rwcarlsen:
Also - since Cyclus is c++11 - why not use std::function and lambdas? -
that would make it easier to include facility time-step specific state in
the cost function (i.e. with closed over vars).

src/request.h
```diff
@@ -26,6 +28,8 @@ template <class T> class RequestPortfolio;
 template <class T>
 class Request {
  public:
+  typedef double (*cost_function_t)(boost::shared_ptr<Material>);
```
scopatz: Good call, I have added this now.

scopatz:
Now uses std::function, thanks for catching that @rwcarlsen

scopatz:
Ping again!

PR #1290, opened 2016-10-05
adding a decay heat calculation to material class

FlanFlanagan:


FlanFlanagan:
@scopatz

src/material.cc
```diff
@@ -193,6 +193,16 @@ void Material::Decay(int curr_time) {
   Transmute(decayed);
 }
 
+double Material::DecayHeat() {
+    double decay_heat = 0.;    
+    pyne::Material p_map = pyne::Material(comp_->atom(), 15.);
```
scopatz: shouldn't this be called `p_mat`? Also what is with the 15?

scopatz: Also, PyNE uses mass weights not atom weights.

src/material.cc
```diff
@@ -193,6 +193,16 @@ void Material::Decay(int curr_time) {
   Transmute(decayed);
 }
 
+double Material::DecayHeat() {
+    double decay_heat = 0.;    
+    pyne::Material p_map = pyne::Material(comp_->mass(), 1.);
```
FlanFlanagan: Ahh I see now

scopatz:
Let me know when you think it is GTG. LGTM Boss

scopatz:
Thanks!

PR #1293, opened 2016-10-11
Bidpref

FlanFlanagan:
Added a preference to bids.

### DIFF
--- a/src/cyclus.h
+++ b/src/cyclus.h
@@ -59,4 +59,9 @@
 #include "toolkit/symbolic_functions.h"
 #include "toolkit/timeseries.h"
 
+// Undefines isnan from pyne
+#ifdef isnan
+  #undef isnan
+#endif
+
 #endif  // CYCLUS_SRC_CYCLUS_H_
--- a/src/exchange_context.h
+++ b/src/exchange_context.h
@@ -3,6 +3,7 @@
 
 #include <assert.h>
 #include <map>
+#include <math.h>
 #include <string>
 #include <utility>
 #include <vector>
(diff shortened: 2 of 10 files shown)


scopatz:
LGTM!

src/exchange_context.h
```diff
@@ -79,8 +80,9 @@ struct ExchangeContext {
 
     bids_by_request[pb->request()].push_back(pb);
 
+    double bid_pref = pb->preference();
     trader_prefs[pb->request()->requester()][pb->request()].insert(
-        std::make_pair(pb, pb->request()->preference()));
+        std::make_pair(pb, std::isnan(bid_pref) ? pb->request()->preference() : bid_pref));
```
gidden: the line in question

gidden:
Hey @FlanFlanagan and @scopatz, I'm very happy you guys are making enhancements to the DRE infrastructure (it's nice to finally get another set of eyes on it, too!). I'm happy to try to look over these PRs once, though I admit I will not always be timely. 12 hours from submission to merge is a bit tight, though.

The way this is implemented is actually a pretty huge change to the methodology of the DRE. Throughout my work, I explicitly noted that it is "requester-based". The fact that this is a limitation is not surprising, but it is what is advertised so far. This PR reverses that, and makes the DRE now "bidder-based". I highlight above the line in question.

I would argue that this kind of change should at minimum be discussed on the list first (please show me if I missed it!). Might I suggest that this is either reverted or a follow up PR is provided after a discussion on the list?

src/cyclus.h
```diff
@@ -59,4 +59,9 @@
 #include "toolkit/symbolic_functions.h"
 #include "toolkit/timeseries.h"
 
+// Undefines isnan from pyne
+#ifdef isnan
```
scopatz: The right thing to do would be to make PyNE C++11 only and then remove the ifdef there.  Until that time, though...

PR #1294, opened 2016-10-13
pyne update

FlanFlanagan:
Updated pyne.cc and pyne.h

### DIFF
--- a/src/pyne.cc
+++ b/src/pyne.cc
@@ -17561,7 +17561,7 @@ std::string pyne::Material::mcnp(std::string frac_type) {
       ss.str("");
       ss << pyne::nucname::mcnp(i->first);
       nucmcnp = ss.str();
-      
+
       int mcnp_id;
       mcnp_id = pyne::nucname::mcnp(i->first);
       // Spaces are important for tests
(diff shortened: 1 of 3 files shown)


src/pyne.cc
```diff
@@ -10339,7 +10339,7 @@ const double pyne::barns_per_cm2 = 1e24;
 const double pyne::cm2_per_barn = 1e-24;
 const double pyne::sec_per_day = 24.0 * 3600.0;
 const double pyne::MeV_per_K = 8.617343e-11;
-const double pyne::MeV_per_MJ = 1.60217657e-19;
+const double pyne::MeV_per_MJ = 6.2415096471204E+18;
```
Baaaaam: I have 2 questions:
- do we want so much precision ?
- how did you get it ? when I take `e` from [NIST](http://physics.nist.gov/cuu/Constants/): `1.602 176 6208 x 10-19 C` I get : `1/e = 6.241 509 1259e+18`

scopatz: 1. This is what is in PyNE, and 
2. Here is where the number comes from, http://www.translatorscafe.com/cafe/EN/units-converter/energy/9-65/megaelectron-volt-kelvin/

And note that MeV << MJ, so for x MeV /  1 MJ, x will be big.

Baaaaam: I don't want to be picky but on translatorscafe.com I got: 6.241506363094e+18.

it is probably better to be consistant with PyNE...

scopatz: Thanks for reviewing @Baaaaam!

PR #1313, opened 2016-11-16
updated deps

scopatz:


DEPENDENCIES.rst
```diff
@@ -106,6 +114,14 @@ and (optionally):
...
 #. liblapack-dev
 #. libgoogle-perftools-dev
+#. python3
+#. python3-dev
+#. python3-tables
+#. python3-pandas
+#. python3-numpy
+#. python3-nose
+#. python3-jinja2
+#. cython3
```
Baaaaam: shall we refer also to python2-xxx ?

Baaaaam:
thx @scopatz

DEPENDENCIES.rst
Baaaaam: for some not careful users, it might mess up their installation....

scopatz: I'd rather not. This is a fully valid install and apt should not mess up anyone's system with this.  Also, I am pretty sure that on newer versions of ubuntu python3 is the default, not python2

gonuke: Ubuntu 16 has python2 as the default, so this set of dependencies will not work.  I have demonstrated that it will not work.

If we remove the `3`, it might (??) work on all systems because it will get the appropriate version.

gonuke:
Sorry that I missed this in review, but can we also add the optional entries to the `apt-get` example, either in the existing `apt-get` example, or by adding a second one that includes all the optional dependencies.  I think @Baaaaam will PR this.

DEPENDENCIES.rst
scopatz: I guess the default version depends on whether you are on ubuntu server or not: https://wiki.ubuntu.com/XenialXerus/ReleaseNotes

Removing the 3 will go to the Python 2 version, but on some level it is up to the user to know whether they want the binding to 2 or 3. If you are using the 3 interpreter, the 2 packages won't help. Maybe they could be listed as "python3-numpy or python-numpy"

PR #1314, opened 2016-11-18
update dependencies

Baaaaam:
this change:
- the existing apt-get line to install only require package
- add a new apt-get line to install all Cyclus dependencies (require and optional)
- change de python dependencies to pyhon-xx only (I remove the python3 line in optional as python is the require dependencies)

**Question:** python is a require dependency by not in the apt-get **require** package ?! Shall I change that to ?


Baaaaam:
I close this because there is no ci-hook on release branch...
I PR it back on develop

PR #1316, opened 2016-11-21
Fix smbch

Baaaaam:
allows `smbchk.py` to deal with some c++11 symbols...

not sure if it is the right way to deal with this issues....

### DIFF
--- a/release/smbchk.py
+++ b/release/smbchk.py
@@ -48,7 +48,7 @@ API_BLACKLIST = {
 SYMBOLS_REPLACEMENTS = [
     ("std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >", "std::string"),
     ("std::__cxx11::list", "std::list"),
-    ("[abi:cxx11]", "")
+    ("[abi:cxx11]", ""),
     ("string >", "string>") # fix string replacement when a '>' follows the string
 ]
 
--- a/release/smbchk.py
+++ b/release/smbchk.py
@@ -89,11 +89,11 @@ def nm(ns):
             continue
         if typ not in ok_types:
             continue
-#        if ' ' in name:
-#            # handle funny private pointer cases
-#            pre, post = name.split(' ', 1)
-#            if pre.endswith('*') or post.startswith('std::__') or post.startswith('const* std::__'):
-#                continue
+        if ' ' in name:
+            # handle funny private pointer cases
+            pre, post = name.split(' ', 1)
+            if pre.endswith('*') or post.startswith('std::__') or post.startswith('const* std::__'):
+                continue
         # use trailing underscore naming convention to skip private variables
         m = NAME_RE.match(name)
         if m is None or m.group(1).endswith('_'):
--- a/release/smbchk.py
+++ b/release/smbchk.py
@@ -45,12 +45,13 @@ api_blacklist = {
     'cyclus::Composition::NewDecay',
 }
 
-cpp11_symbols = [
+CPP11_SYMBOLS_REPLACEMENTS = {
     ["std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >", "std::string"],
     ["std::__cxx11::list", "std::list"],
     ["std::basic_stringstream", "std::basic_stringstream"],
     ["[abi:cxx11]", ""]
-]
+}
+
 
 def load(ns):
     """Loads a database of symbols or returns an empty list."""
@@ -60,10 +61,12 @@ def load(ns):
         db = json.load(f)
     return db
 
+
 def save(db, ns):
     """Saves a database of symbols."""
     with io.open(ns.filename, 'wb') as f:
         json.dump(db, f, indent=1, separators=(',', ': '))
+
 
 def nm(ns):
     """Obtains the latest symbols as a sorted list by running and parsing the
@@ -85,10 +88,10 @@ def nm(ns):
     ok_types = {'B', 'b', 'D', 'd', 'R', 'r',
                 'S', 's', 'T', 't', 'W', 'w', 'u'}
     for line in stdout.splitlines():
-        # replace c++11 symbol by standart one
-        for symbol in cpp11_symbols:
-            line  =line.replace(symbol[0],symbol[1])
-        #fix string replament when a '>' follows the string
+        # replace c++11 symbol by standard one
+        for symbol in CPP11_SYMBOLS_REPLACEMENTS:
+            line = line.replace(symbol[0], symbol[1])
+        # fix string replacement when a '>' follows the string
         line = line.replace("string >", "string>")
 
         line = line.strip().decode()
@@ -111,15 +114,19 @@ def nm(ns):
         names.add(name)
     return sorted(names)
 
+
 def git_log():
     """Returns git SHA, date, and timestamp from log."""
-    stdout = subprocess.check_output(['git', 'log', '--pretty=format:%H/%ci/%ct', '-n1'])
+    stdout = subprocess.check_output(
+        ['git', 'log', '--pretty=format:%H/%ci/%ct', '-n1'])
     return stdout.decode().strip().split('/')
+
 
 def core_version():
     stdout = subprocess.check_output(['cyclus', '--version'],
                                      universal_newlines=True)
     return stdout.splitlines()[0].strip()
+
 
 def update(db, ns):
     """Updates a symbol database with the latest values."""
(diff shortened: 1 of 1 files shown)


release/smbchk.py
```diff
@@ -76,6 +76,11 @@ def nm(ns):
     names = set()
     ok_types = {'B', 'b', 'D', 'd', 'R', 'r', 'S', 's', 'T', 't', 'W', 'w', 'u'}
     for line in stdout.splitlines():
+        line = line.replace("std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >","std::string")
+        line = line.replace("std::__cxx11::list","std::list")
+        line = line.replace("std::__cxx11::basic_stringstream","std::basic_stringstream")
+        line = line.replace("[abi:cxx11]","")
+        line = line.replace("string >","string>")
```
scopatz: PEP8, space after comma

release/smbchk.py
```diff
@@ -76,6 +76,11 @@ def nm(ns):
     names = set()
     ok_types = {'B', 'b', 'D', 'd', 'R', 'r', 'S', 's', 'T', 't', 'W', 'w', 'u'}
     for line in stdout.splitlines():
+        line = line.replace("std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >","std::string")
```
scopatz: PEP8 line over 80 char

scopatz:
Thanks for this @Baaaaam!

Baaaaam:
@scopatz @gonuke could some one take a look here ?

release/smbchk.py
```diff
@@ -70,12 +77,20 @@ def nm(ns):
...
     stdout = subprocess.check_output(['nm', '-g', '-p', '-C', '-fbsd', lib])
     names = set()
-    ok_types = {'B', 'b', 'D', 'd', 'R', 'r', 'S', 's', 'T', 't', 'W', 'w', 'u'}
+    ok_types = {'B', 'b', 'D', 'd', 'R', 'r',
+                'S', 's', 'T', 't', 'W', 'w', 'u'}
     for line in stdout.splitlines():
+        # replace c++11 symbol by standart one
+        for symbol in cpp11_symbols:
+            line  =line.replace(symbol[0],symbol[1])
+        #fix string replament when a '>' follows the string
```
scopatz: PEP8 single space after "#"

release/smbchk.py
```diff
@@ -70,12 +77,20 @@ def nm(ns):
...
     lib = os.path.abspath(os.path.join(ns.prefix, 'lib', 'libcyclus.so'))
     stdout = subprocess.check_output(['nm', '-g', '-p', '-C', '-fbsd', lib])
     names = set()
-    ok_types = {'B', 'b', 'D', 'd', 'R', 'r', 'S', 's', 'T', 't', 'W', 'w', 'u'}
+    ok_types = {'B', 'b', 'D', 'd', 'R', 'r',
+                'S', 's', 'T', 't', 'W', 'w', 'u'}
     for line in stdout.splitlines():
+        # replace c++11 symbol by standart one
+        for symbol in cpp11_symbols:
+            line  =line.replace(symbol[0],symbol[1])
```
scopatz: PEP8 single space on both sides of `=`

scopatz: Also PEP8 single space after commas

release/smbchk.py
```diff
@@ -45,6 +45,13 @@
     'cyclus::Composition::NewDecay',
 }
 
+cpp11_symbols = [
```
scopatz: PEP8 global (module-level) variables should be `UPPERCASE_WITH_UNDERSCORES`

release/smbchk.py
```diff
@@ -45,6 +45,13 @@
...
+    ["std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >", "std::string"],
+    ["std::__cxx11::list", "std::list"],
+    ["std::basic_stringstream", "std::basic_stringstream"],
+    ["[abi:cxx11]", ""]
```
scopatz: The above should be tuples since they must be length-2

release/smbchk.py
scopatz: Also maybe this variable name could have the word "replacements" in it, since that is what it is.

release/smbchk.py
```diff
@@ -70,12 +77,20 @@ def nm(ns):
...
     names = set()
-    ok_types = {'B', 'b', 'D', 'd', 'R', 'r', 'S', 's', 'T', 't', 'W', 'w', 'u'}
+    ok_types = {'B', 'b', 'D', 'd', 'R', 'r',
+                'S', 's', 'T', 't', 'W', 'w', 'u'}
     for line in stdout.splitlines():
+        # replace c++11 symbol by standart one
+        for symbol in cpp11_symbols:
+            line  =line.replace(symbol[0],symbol[1])
+        #fix string replament when a '>' follows the string
+        line = line.replace("string >", "string>")
```
Baaaaam: because those change affect only `string` and `list`:
https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html

Baaaaam:
@scopatz I have addressed all your comments. Thank you for reviewing!

release/smbchk.py
```diff
@@ -70,12 +79,20 @@ def nm(ns):
...
     names = set()
-    ok_types = {'B', 'b', 'D', 'd', 'R', 'r', 'S', 's', 'T', 't', 'W', 'w', 'u'}
+    ok_types = {'B', 'b', 'D', 'd', 'R', 'r',
+                'S', 's', 'T', 't', 'W', 'w', 'u'}
     for line in stdout.splitlines():
+        # replace c++11 symbol by standard one
+        for symbol in CPP11_SYMBOLS_REPLACEMENTS:
+            line = line.replace(symbol[0], symbol[1])
+        # fix string replacement when a '>' follows the string
+        line = line.replace("string >", "string>")
```
scopatz: I still think that there should only be a single replacements list that covers all replacements and that this pair should be included there.

release/smbchk.py
```diff
@@ -45,6 +45,13 @@
     'cyclus::Composition::NewDecay',
 }
 
+CPP11_SYMBOLS_REPLACEMENTS = [
```
scopatz: Maybe this name should be changed to `SYMBOL_REPLACEMENTS` if it is extended beyond C++11

Baaaaam:
I am wondering is this should actually be done:
using C++11 actually implies a potential ABI change: [gcc 5.x dual ABI](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html)

those change will hide it... I am not sure this is what we want...

scopatz:
Assuming you use the same compiler, it will still be ABI compatible.

gonuke:
So does this mean that we are really only guaranteeing ABI compatibility within a given compiler version?  Or within a given ABI version?

gonuke:
And if so - do we want to mask these differences caused by ABI updates?  Or do we just want to recognize that we should not even measure ABI differences across an ABI update?

scopatz:
ABI compatibility does [not typically cross compilers](http://softwareengineering.stackexchange.com/a/235768/196844), at least in part due to the forced linkage to the run time.  C++ standard templates exacerbate this.  

Furthermore, GCC has started treating each minor version as a different compiler. IIRC GCC v5.2 fixes is another breaking version (https://gcc.gnu.org/wiki/Cxx11AbiCompatibility)

I think masking is fine because the presence or lack of these symbols does not impact linking with objected complied differently.  They aren't public symbols and they aren't exposed in any header as part of  the API.

gonuke:
I'm particularly unsure about the SYMOLS_REPLACEMENT addition here.  This particular set of replacements is useful across this ABI change.  I'm trying to wrap my head around the possibility that they could accidentally hide ABI changes in the future?  Otherwise, I'm ready to merge this...

scopatz:
This set really only hides things that are compiler private (which isn't part of the ABI, per say) and does some string cleanup.  I think it is pretty safe going forward.

gonuke:
Is it better to do this symbol replacement, or simply update a new reference ABI that uses the new symbol description?  That is, recognizing that the ABI has changed because of the compiler, should we just have a new base version of the ABI against which we compare that uses the new representation of these symbols so that we don't have to maintain this symbol replacement list going forward?

scopatz:
I think it would be way more work to support the private symbols of many compilers (even in the same family) than it would be support this replacement list.

gonuke:
OK - let's try this and see where it takes us...

gonuke:
Thanks for hunting this down @Baaaaam

PR #1323, opened 2016-12-02
Initial Event Hooks

scopatz:
This adds some initial event hooks into the Cyclus timer. Note that there are no actual event types and the event loop is just a stub. But since it is a pretty annoying change, I figured I'd PR it now.  Also, I have started wrapping some of the core so that we can get a python driver / server running.


scopatz:
@FlanFlanagan

PR #1328, opened 2016-12-08
install cyclus python lib in the default path if path is not specified

### LINKED ISSUE #1324: Python lib default installation path
On MACOS, when not specifying an installation path with installing the full Cyclus suite (Cyclus/Cycamore/Cymetric) the Cyclus new python lib are copied in `~/.local/lib/pythonX.x/package-site` and Cymetric python lib in `~/Library/Python/X.x/lib/python/site-packages`.

I guess it is because the default python lib path is:
- `~/Library/Python/X.x/lib/python/site-packages/` on MACOS
- `~/.local/lib/pythonX.x/package-site` on Linux/Ubuntu...

I think we need consistency so we need either to install everything on the default python lib path (which vary depending of the distribution) or `~/.local/lib/pythonX.x/package-site` and add it to the `LD_LIBRARY_PATH`/`PYTHONPATH`

I don't have any personal preferences, I was wondering what @scopatz think (as your python expertise is way bigger than mine)?

Baaaaam:
this check if a `--prefix` or a `-DCMAKE_PREFIX_PATH`flag  are provided, and install the cyclus python lib:
 - with `--user` flag if no `--prefix`  or a `-DCMAKE_PREFIX_PATH` flag are given 
- with `--prefix=${CMAKE_PREFIX_PTH}` if one of them is given

this is the other solution to fix #1324 issue

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,13 +1,19 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
 #taken from http://geant4.cern.ch/support/source/geant4/CMakeLists.txt
 IF(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
-    message(STATUS "Cyclus requires an out-of-source build.")
-    message(STATUS "Please remove these files from ${CMAKE_BINARY_DIR} first:")
-    message(STATUS "CMakeCache.txt")
-    message(STATUS "CMakeFiles")
-    message(STATUS "Once these files are removed, create a separate directory")
-    message(STATUS "and run CMake from there")
-    message(FATAL_ERROR "in-source build detected")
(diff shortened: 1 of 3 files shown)


Baaaaam:
@FlanFlanagan could you review this please ?

@meghan could you confirm you can load cymetric with this one ? test this with Cymetric/v1.5.0
If should work without having to add anything to your $PYTHONPATH. if you want `Cymetric` executable to work add `~/Library/Python/X.X/bin` to your $PATH


Please do not merge it yet, I like to have @gonuke and/or @scopatz opinion about it.

gonuke:
I endorse the "other" option: continuing to install the python components of Cyclus in a cyclus-specific location, also installing cymetric there, and updating user documentation to reflect this.

As such, I'm closing this PR.

PR #1331, opened 2016-12-17
Updated cython to most recent version

scopatz:
Needed to avoid deprecation issues.


gonuke:
Thanks @scopatz 

Looking at this, it's not clear that we have comprehensive enough docs about updating `cyclus-deps`, but we'll consider that under a different revision.

PR #1330, opened 2016-12-16
startup python in mocksim

scopatz:
This may address #1372 @mbmcgarry @Baaaaam

PR #1333, opened 2016-12-22
WIP: Hdf5 code generation, VLDataset, WriteToBuf, FillBuf

hodger:
I've implemented generation for three more Hdf5 backend functions. Currently, I am seeing all Hdf5 tests passing. I still need to finish support for nested container types before this is merged in. 

Strangely, I seem to be getting a segfault on the bugfix tests- we'll have to see if that shows up on the CI build.

@scopatz


scopatz:
@hodger - let us know when this is good for another look

hodger:
@scopatz Looking at CI, it seems like the unit tests are still segfaulting at Lone Trader, even after I merged in the recent fixes from develop. Any idea why that might be?

scopatz:
Hmm, could you try merging from develop again?  @FlanFlanagan merged in the pywrap branch which I think may have more fixes (or at least works!). Thanks!

hodger:
Looks like Lone Trader still isn't passing. It's strange, since I was seeing all unit tests passing on your branch from #1330.

scopatz:
Software is the worst.  Do the tests pass locally for you? It is possible that this is new segfault in the same test. Note that the Python tests seem to pass OK, maybe.

hodger:
I get a segfault after Lone Trader in the same place that the CI does, unfortunately.

scopatz:
Weird, so do I.  But If I run the Lone test by itself (alone?) it runs just fine.

scopatz:
Ok, the segfault is in importing numpy when we import cyclus.events. The backtrack is 320 calls deep, so I won't include it here.

scopatz:
Ok this problem is happening on develop for me too.

scopatz:
Yeah, this is totally different. I am investigating now.

scopatz:
I think it is this issue: numpy/numpy#7595

scopatz:
You might want to merge in the other PR with the fix

FlanFlanagan:
I've tested this with the numpy fix and cyclus unit tests all pass for me. Is this good to go in @hodger ?

hodger:
I believe that all the generated code is correct at this point. I'll probably need to tweak some of this as we finish generating the other hdf5_back functions and begin adding more complicated types. It's probably safe to put this in though since we're currently trying to generate what was originally there.

scopatz:
Thanks @hodger!

PR #1338, opened 2017-01-05
Fixed numpy segfault

scopatz:
This fixes the segfault that is in the test suite that was caused by numpy. It also cleans up the pyhooks header so downstream packages don't have to have the same compile-time value of `CYCLUS_WITH_PYTHON` that cyclus had.


scopatz:
CC @hodger @FlanFlanagan and anyone else who is interested!

PR #1339, opened 2017-01-13
Python Simulation Input Files

scopatz:
This adds the ability to use Python files as input files to Cyclus.  This PR builds off of the server PR and is probably long overdue.


scopatz:
CC @FlanFlanagan et al.

FlanFlanagan:
Installs properly on linux and passes all unit tests. Is this ready @scopatz ?

scopatz:
I think it is GTG

PR #1340, opened 2017-01-13
WIP: attempt to add `std::tuple` into cyclus

bam241:
- [ ] hooks in cycpp 
  - [ ] header ?
  - [x] _val_tuple
  - [x] read_tuple
- [ ] Add to/from python converters for std::tuple to the top of gentypesystem.py 
- [ ]  CYCLUS_BINDVAL and CYCLUS_LOADVAL C macros in the sqlite backend actually work on your type.
- [ ] Add the required code gen hooks to hdf5_back_gen.py. 
- [ ] Update the documentation generator cyclusagent.py
- [ ] Add tests as necessary.


Baaaaam:
If you ever pass by this PR, feel free to take a look on the checked item, to make sure it is implemented correctly.

thx

cli/cycpp.py
```diff
@@ -746,6 +748,7 @@ def includeloc(self, statement=None):
         'std::list': ('T',),
         'std::pair': ('T1', 'T2'),
         'std::map': ('Key', 'T'),
+        'std::tuple': ('T',),
```
scopatz: So this is incorrect because tuples have an arbitrary number of template args, not 1.  We need to have some way of denoting this. Maybe we need a special list type that returns 'T{n}' for the nth index.

scopatz:
Just ping whenever you want someone to take a look :)

cli/cycpp.py
```diff
@@ -1500,6 +1525,35 @@ def read_pair(self, member, alias, t, uitype=None, ind="  ", idx=None,
...
+    def read_tuple(self, member, alias, t, uitype=None, ind="  ", idx=None,
+                  path=''):
+        uitype = prepare_type(t, uitype)
+        alias = prepare_type(t, alias)
+        
+        # the extra assignment (bub, sub) is because we want the intial sub
+        # rhs to be from outer scope - otherwise the newly defined sub will be
+        # in scope causing segfaults
+        tree_idx = idx or '0'
+        n_val = len(range(alias))
```
scopatz: Can't you just take the len(alias)?

cli/cycpp.py
```diff
@@ -1753,6 +1807,22 @@ def _buildschema(self, cpptype, schematype=None, uitype=None, names=None):
...
             impl += '</interleave>'
             impl += '</element>'
+        elif t == 'std::tuple':
+            name = 'tuple'
+            if names[0] is not None:
+                name = names[0]
+            firstname = 'first' if isinstance(cpptype[1], STRING_TYPES) else ['first']
+            if names[1] is not None:
+                firstname = names[1]
+            secondname = 'second' if isinstance(cpptype[2], STRING_TYPES) else ['second']
```
scopatz: First and second don't apply to tuples.

cli/cycpp.py
```diff
@@ -570,6 +570,7 @@ def _canonize_ann(self, defaults, t, name, ann=None):
         'std::list': (None, 'val'),
         'std::pair': (None, 'first', 'second'),
         'std::map': ((None, 'item'), 'key', 'val'),
+        'std::tuple': ((None, 'item'), 'val'),
```
scopatz: need 'val{n}'

cli/cycpp.py
```diff
@@ -584,6 +585,7 @@ def canonize_alias(self, t, name, alias=None):
         'std::list': (None, ''),
         'std::pair': (None, '', ''),
         'std::map': ((None, ''), '', ''),
+        'std::tuple': ((None, ''), ''),
```
scopatz: needs infinite empty strings

cli/cycpp.py
scopatz: Or you could use some stuff in itertools to fake an infinite list. That is probably better.

bam241:
@scopatz I need to confess this is a bit complicated...

In the `VarDeclarationFilter` class
```
567     _default_aliases = {
568         'std::vector': (None, 'val'),
569         'std::set': (None, 'val'),
570         'std::list': (None, 'val'),
571         'std::pair': (None, 'first', 'second'),
572         'std::map': ((None, 'item'), 'key', 'val'),
573         'std::tuple': ((None, 'item'), 'val'),
574         }
[...]
582     _default_ui = {
583         'std::vector': (None, ''),
584         'std::set': (None, ''),
585         'std::list': (None, ''),
586         'std::pair': (None, '', ''),
587         'std::map': ((None, ''), '', ''),
588         'std::tuple': ((None, ''),''),
589         }
```
If I understand correctly this 2 part are suppose to represent how the `alias` and the `uitype` will be written in the header file of an cyclus archetypes ?
so the first one should be something like:  
'std::tuple': ((None, 'item_1', ..., 'item_n'),
and the second one:
 'std::tuple': (None, '', ... ,''),

But as it is not possible to have the "...", you suggest me to use itertools to make something up, right ?

I am not sure to understand what is this one for (In the `StateAccumulator` class):
```
745     known_templates = {
746         'std::vector': ('T',),
747         'std::set': ('T',),
748         'std::list': ('T',),
749         'std::pair': ('T1', 'T2'),
750         'std::map': ('Key', 'T'),
751         'std::tuple': ('T{n}',),
752         '{0}::toolkit::ResBuf'.format(CYCNS): ('T',),
753         '{0}::toolkit::ResMap'.format(CYCNS): ('K', 'R'),
754         }
```

scopatz:
Ok, So the itertools-based expression I came up with for the default aliases is:

```python
tsa = lambda x, y: str(x) + str(y)
x = chain([None], starmap(tsa, zip(repeat('val'), count(0))))
```

The problem here is that everything that might use this would have to check if it is a (Python) tuple or an iterator. I think this would have to happen no matter what mechanism you use. 

As an alternative to itertools, you could just use a template string, like `val{n}`, and then check if the first element contains a `{` or not. This will also allow you to figure out whether the type allows variadic template arguments or not.  

The important piece is that the places where these are used, need to be updated to check for variadic arguments.  It is not enough to just add a type. 

Does this make sense?

scopatz:
I can try to help out with some of this stuff.

FlanFlanagan:
@bam241 What's the status of this?

PR #1349, opened 2017-02-09
Test generation for Hdf5 back

hodger:
The tests generated by this PR are meant to serve the same purpose as those found in hdf5_back_tests.cc. This PR tests each individual supported datatype. I can add other, more involved tests, but this is currently functional as-is. 

@scopatz the List of Naughty Strings seems to contain bytes that cause errors when passed through the type system, so I haven't implemented that yet.


tests/hdf5_back_gen_tests.py
```diff
@@ -0,0 +1,289 @@
+import os
+import sys
+import json
+import subprocess
+from random import randint
+import uuid
+
+from cyclus.lib import Hdf5Back, Recorder
+import cyclus.typesystem as ts
```
scopatz: PEP8 local and package imports should come after all other imports

tests/hdf5_back_gen_tests.py
```diff
@@ -0,0 +1,289 @@
...
+
+from hdf5_back_gen import resolve_unicode, convert_canonical, init_dicts, ORIGIN_DICT
+
+#Call to hdf5_back_gen function 
+init_dicts()
+
+#Find supported HDF5 types.
+with open(os.path.join(os.path.dirname(__file__), '..', 'share', 
+                       'dbtypes.json')) as f:
+    RAW_TABLE = resolve_unicode(json.load(f))
```
scopatz: The above couple of thing should be put in a `setup()` function.

scopatz: Actually a lot of the stuff in this module can be put in a setup function

scopatz:
Thanks @hodger!

PR #1351, opened 2017-02-12
Add HDF5 support for some Vector, List, Set types

hodger:
This PR adds Hdf5 support for the following types:

- Vector_Blob
- Vector_UUID
- Set_double
- Set_float
- Set_Blob
- Set_UUID
- List_float
- List_double
- List_Blob
- List_UUID

**As well as the VL versions of these types.**

Map and Pair types are forthcoming. 
@scopatz


scopatz:
This looks great, thanks!

PR #1365, opened 2017-04-01
fix the enrichment::UraniumAssay method 

bam241:
this follows the dev mail: https://groups.google.com/forum/#!topic/cyclus-dev/McK0-uOcnbA

it just change the way the UraniumAssay is calculated, previously it return the atomic Uranium Assay, now it returns the mass Uranium Assay.

I also added a method to return the atomic fraction of a set of nuclides.

### DIFF
--- a/src/toolkit/enrichment.h
+++ b/src/toolkit/enrichment.h
@@ -31,7 +31,7 @@ class Assays {
 };
 
 /// @param mat the material inquired about
-/// @return the mass/atom percent of U-235 w.r.t U-235+U-238 in a material
+/// @return the atom percent of U-235 w.r.t U-235+U-238 in a material
 double UraniumAssay(Material::Ptr mat);
 
 /// @param mat the material inquired about
@@ -63,7 +63,8 @@ double FeedQty(double product_qty, const Assays& assays);
 /// @param product_qty the amount of product Uranium
 /// @param assays the assay of product, feed, and tails
 /// @return the quantity of tails resulting from enriching the product
-/// whose units match those of the given product
+/// whose units match those of the given product; product_qty and assay
+/// must share the same units: both in mass or in atom
 double TailsQty(double product_qty, const Assays& assays);
 
 /// @param product_qty the amount of product Uranium
--- a/src/toolkit/enrichment.h
+++ b/src/toolkit/enrichment.h
@@ -31,7 +31,7 @@ class Assays {
 };
 
 /// @param mat the material inquired about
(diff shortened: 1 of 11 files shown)


scopatz:
Looks like you need to merge from develop.

src/toolkit/enrichment.h
```diff
@@ -31,7 +31,7 @@ class Assays {
 };
 
 /// @param mat the material inquired about
-/// @return the atom percent of U-235 w.r.t U-235+U-238 in a material
+/// @return the mass percent of U-235 w.r.t U-235+U-238 in a material
```
scopatz: For stability reasons, I don't think that you can do this.  I think that you need to add a different state var (UraniumAssayByMass?) and then make them both optional.  This change will break existing input files.

src/toolkit/mat_query.cc
```diff
@@ -41,6 +41,19 @@ double MatQuery::atom_frac(Nuc nuc) {
...
 }
 
+double MatQuery::atom_frac(std::set<Nuc> nucs) {
+  CompMap v = m_->comp()->atom();
+  compmath::Normalize(&v);
+  
+  double frac_tot = 0;
+  std::set<Nuc>::iterator it ;
+  for (it = nucs.begin(); it != nucs.end(); ++it) {
+    frac_tot += v[*it];
```
scopatz: You should check if `*it` is in `v` first, so that extra nuclides aren't added to the map.

src/toolkit/enrichment.h
bam241: yes you are right!

bam241:
@scopatz I did as you suggested, before I add test could you confirm this is the correctly to do ?
thx

bam241:
I change the unit test to be correct, I'll re-add a part to check a `UraniumAssayAtom`

scopatz:
Hi @bam241, can you please merge with develop here and see if the CI still works? Thanks!

bam241:
@scopatz just rebased

bam241:
@scopatz seems to be working

scopatz:
Yeah, for no apparent reason.

bam241:
@scopatz @gonuke @FlanFlanagan could you take  look. This should add the missing method in enrichment.
as well as fix the unit test. (the check where ok, but on a wrong physics case)

gonuke:
This looks good to me...

mbmcgarry:
Thanks @bam241 !

PR #1368, opened 2017-04-28
changes for hdf5 table merging

hodger:
LoadTableTypes now calls CreateTable when an unrecognized table is being added. This is required for merging two Hdf5 output files together if the same tables are not present in both.


scopatz:
Thanks @hodger!

PR #1371, opened 2017-05-10
Fix to Schema for rickshaw database merge

hodger:
This fixes the `Schema` bug which was throwing a malloc error when trying to merge two HDF5 databases which use the Separations archetype. 

We were allocating the wrong number of elements in the `shape` vector in `Hdf5Back::Schema`. Instead of allocating for the number of elements, we were only doing the number of dimensions (which was always 1).


hodger:
@scopatz I've tested this a few times on rickshaw files with Separations. Seems to work now.

scopatz:
Great! Thanks!

PR #1379, opened 2017-05-16
now flag clang version when using MACOS clang

### LINKED ISSUE #1376: ExXlateTests.XlateCapacities fails possibly because of reversed rhs and lhs
The ExXlateTests.XlateCapacities test fails with a strange error where it seems that the rhs and lhs are reversed.
This is on the master branch on OSX 10.12.4
> gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 8.1.0 `(clang-802.0.42)`
Target: x86_64-apple-darwin16.5.0
>

Here is the only test that fails
> [ RUN      ] ExXlateTests.XlateCapacities
/Users/yarden/software/git/cyclus/tests/equality_helpers.h:22: Failure
Value of: rhs[i]
  Actual: 0.48999999999999988
Expected: lhs[i]
Which is: 0.69999999999999996
Vectors differ at index 0
/Users/yarden/software/git/cyclus/tests/equality_helpers.h:22: Failure
Value of: rhs[i]
  Actual: 0.69999999999999996
Expected: lhs[i]
Which is: 0.48999999999999988
Vectors differ at index 1
[  FAILED  ] ExXlateTests.XlateCapacities (0 ms)
>

As a side note:  if instead of installing cyclus from sources I install it via miniconda3 than the installation itself finishes ok but
>cyclus_unit_tests 
dyld: Library not loaded: @rpath/libicui18n.56.dylib
  Referenced from: /Users/yarden/software/tools/miniconda3/bin/cyclus_unit_tests
  Reason: image not found
Abort trap: 6
>
which can be overcome with 
>conda install icu=56.1
>
after which all the tests do pass ok as oppose to installing from sources

bam241:
this should do fix #1376

does anyone with a mac computer with macOS `El Capitain` or earlier (10.11 or earlier) could test this ?


scopatz:
Thanks a million @bam241! LGTM, maybe we can wait a day and see if anyone has such a mac

bam241:
@scopatz shall we wait longer ?

scopatz:
No, this is probably good to go in. Thanks @bam241!  Just waiting on  CI

Issue #1375, opened 2017-05-15
Python3 Instructions for Ubuntu should include update-alternatives


katyhuff:
The default behavior on ubuntu is to ignore that you've installed python3 and various python3 packages... ubuntu will just keep pointing at python2.7 until you tell it to look at your new python3 install. To get ubuntu to use, by default, the python3 the user has just installed we now have to use "update-alternatives" (as here: https://benzidwael.wordpress.com/2015/04/17/installing-different-python-versions-in-ubuntu/) 

Without the update-alternatives step, the build will fail to use python3, even if the user is following the python3 track in the current fuelcycle.org instructions. 

This issue can be closed when the installation docs include a hint about changing the python version (using update-alternatives). This hint should appear after the hint about python -V. http://fuelcycle.org/user/DEPENDENCIES.html#installing-dependencies-linux-and-unix

katyhuff:
The python version information will need to be changed in the docs here:
https://github.com/cyclus/cyclus/blob/develop/DEPENDENCIES.rst

This information should be placed right after :+1: 
```
To determine which version of Python is already installed on your computer, run:

python -V
```

skc222:
Can we close this issue?

scopatz:
Yes, feel free to close it.

skc222:
I think I don't have permissions to do so as I cannot see any button to close issue.

PR #1381, opened 2017-05-17
changes in dependencies.rst to resolve python version related issue

skc222:
Updated dependencies.rst file to resolve issue #1375


DEPENDENCIES.rst
```diff
@@ -165,6 +165,14 @@ To determine which version of Python is already installed on your computer, run:
 
    python -V
 
+By default ubuntu keeps pointing at python2.7. To get ubuntu to use, by default, the installed python3, use update-alternatives commands as follows:
```
scopatz: Please keep lines at < 80 characters

scopatz:
Thanks for putting this in @skc222!  Also, this has conflicts, so please resolve those.

DEPENDENCIES.rst
```diff
@@ -165,6 +165,15 @@ To determine which version of Python is already installed on your computer, run:
 
    python -V
 
+
+By default ubuntu keeps pointing at python2.7. To change this default to python3, please use:
```
katyhuff: This looks great! I want to suggest that we include a little more detail


```
Despite having installed python3, Ubuntu installations may still point at python2.7 by default. 
If python -V returns a version of python that is not preferred, the python version can be 
changed system-wide with the update-alternatives command. 

First, you can list alternatives with the following command:

 .. code-block:: bash

   update-alternatives --list python

Ubuntu may not list any alternatives, even though you have a few versions of python installed on your system. You can make Ubuntu aware of python 2.7 and python 3.5, for example, with the following commands:

 .. code-block:: bash

   sudo update-alternatives --install /usr/bin/python python /usr/bin/python2.7 1
   sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.5 2

From now on, you can switch versions with the following command, which will allow you to make a choice between them:


 .. code-block:: bash

   sudo update-alternatives --config python

```

scopatz:
Also @skc222 please ping here when you think it is ready for another look.

skc222:
@scopatz  @katyhuff  It is ready for another look.

scopatz:
LGTM, thanks!

PR #1386, opened 2017-05-26
Changed stub/src .cc files to #include instead of #inlcude pyhooks.h

emmaredfoot:


scopatz:
Thanks @emmaredfoot! This looks good to me after CI runs.

bam241:
nice catch! this should fix #1386

PR #1390, opened 2017-06-06
GIS toolkit

gyutaepark:
Allows institutions, facilities, agents and regions to have geographic coordinates.
Allows simple as-the-crow-flies distance calculations between two objects with GIS coordinates.


src/toolkit/gis.cc
```diff
@@ -0,0 +1,158 @@
+#include "gis.h"
+
+using namespace std;
```
scopatz: This is not allowed under the GCSG

FlanFlanagan:
Can you add a set_position(lat, long) function?

src/toolkit/gis.h
```diff
@@ -0,0 +1,158 @@
...
+///
+/// Brief Description of ISO 6709:
+/// 1. Latitude comes before Longitude
+/// 2. North latitude is positive
+/// 3. East longitude is positive
+/// 4. Fraction of degrees (decimal values) is prefered in digital data
+/// exchange.
+///
+/// The class is adapted from 'https://github.com/jaime-olivares/coordinate'
+/// under the MIT License.
```
scopatz: The MIT license requires the copyright notice to be included, specifically here `Copyright (c) 2015 Jaime Olivares`

src/toolkit/gis.h
```diff
@@ -0,0 +1,158 @@
...
+/// to the jaime-olivares, allows the coordinate elements such as degrees,
+/// minutes, and seconds to "remain on the integral portion of values, with the
+/// exception of decimal of seconds, avoiding loss of precision." This is
+/// calculated by multiplying decimal degrees by 3600.
+/// example: 05.2169 -> 18780.84
+
+class GIS {
+ public:
+  /// The default constructor for GIS. This will creat an object with
+  /// lat =0, and long=0.
```
scopatz: odd space before `=`

src/toolkit/gis.h
```diff
@@ -0,0 +1,158 @@
...
+#include <string>
+
+#include "cyclus.h"
+
+// forward declarations
+namespace cyclus {
+namespace toolkt {
+class GIS;
+}
+}
```
scopatz: This forward declaration is not needed since the class is declared right below.

src/toolkit/gis.h
```diff
@@ -0,0 +1,158 @@
...
+ public:
+  /// The default constructor for GIS. This will creat an object with
+  /// lat =0, and long=0.
+  GIS();
+
+  /// GIS constructor with latitude and longditude as degrees as a decimal
+  /// number.
+  /// @param deciaml_lat latitude expressed in degrees as a decimal number.
+  /// @param decimal_lon longditude expressed in degrees as a decimal number.
+  GIS(float decimal_lat, float decimal_lon);
```
scopatz: Why is this based on floats, rather than doubles? Is this truly enough precision? Or should this be a templated class based on the floating point precision?

src/toolkit/gis.h
```diff
@@ -0,0 +1,158 @@
...
+
+class GIS {
+ public:
+  /// The default constructor for GIS. This will creat an object with
+  /// lat =0, and long=0.
+  GIS();
+
+  /// GIS constructor with latitude and longditude as degrees as a decimal
+  /// number.
+  /// @param deciaml_lat latitude expressed in degrees as a decimal number.
```
scopatz: This is perhaps overly nitpicky, so forgive me in advance, but the number here is not decimal.  Rather the computer uses  binary, which is what is being used when you use `float`.  If you to meant to say real numbers here instead, that is not strictly correct either as floating point is an approximation to the reals. 

I would simply say floating point instead of decimal.

src/toolkit/gis.h
```diff
@@ -0,0 +1,158 @@
...
+  /// @param decimal_lon longditude expressed in degrees as a decimal number.
+  GIS(float decimal_lat, float decimal_lon);
+
+  /// The default destructor for GIS
+  ~GIS();
+
+  /// Returns the current latitude.
+  /// @return decimal representation of latitude.
+  float get_latitude() const;
```
scopatz: Small stylistic thing, which is non-blocking, but it is more concise to just call this getter function `latitude()`.  Same applies to the other accessors below.

src/toolkit/gis.h
```diff
@@ -0,0 +1,158 @@
...
+
+  /// Returns the current longitude.
+  /// @return decimal representation of longitude.
+  float get_longitude() const;
+
+  /// Sets a new latitude
+  /// @param lat latitude expressed in decimal degrees.
+  void set_latitude(float lat);
```
scopatz: Like the accessors above, you can also just call this function `latitude(float lat)` and rely on function overloading. Other parts of Cyclus do this.

src/toolkit/gis.h
```diff
@@ -0,0 +1,158 @@
...
+  /// @parblock
+  ///       The format of output. Options are '1', '2', or '3':
+  ///
+  ///     1:   return in degrees format
+  ///     2:   return in degrees and minutes format
+  ///     3:   return in egrees minutes seconds format
+  /// @endparblock
+  /// @return String representation of the GIS object that complies with ISO6709
+  /// Annex H
+  std::string ToString(int return_format) const;
```
scopatz: I recommend having an enum that represents these options, so the user code is more readable. Also, I recommend having a default value (probably of 1) so that users don't have to worry about always supplying this value if they just want to get a string.

src/toolkit/gis.h
```diff
@@ -0,0 +1,158 @@
...
+///
+/// Longitude and Latitude is stored as seconds of degrees. This was also
+/// adapted from 'https://github.com/jaime-olivares/coordinate'. This, according
+/// to the jaime-olivares, allows the coordinate elements such as degrees,
+/// minutes, and seconds to "remain on the integral portion of values, with the
+/// exception of decimal of seconds, avoiding loss of precision." This is
+/// calculated by multiplying decimal degrees by 3600.
+/// example: 05.2169 -> 18780.84
+
+class GIS {
```
scopatz: GIS - a global information system -  seems to be too broad of a name for what is here right now.  The state that this class represents is really just a Position right now.  I tend to think of a GIS as a suite of tools that manage or arrange many positions. The only tool here is distance, which I agree does make sense to live on a Position object. 

If the concern is that `Position` is too general of a name and that you may want to add other tools in the future, I think adding a `gis` namespace is a good idea. This would thus live in `cyclus::toolkit::gis`.

src/toolkit/gis.cc
```diff
@@ -0,0 +1,158 @@
...
+
+namespace cyclus {
+namespace toolkit {
+
+GIS::GIS() : latitude_(0), longitude_(0) {}
+
+GIS::GIS(float decimal_lat, float decimal_lon) {
+  latitude_ = SetPrecision(decimal_lat * 3600, 1);
```
scopatz: 3600 Should probably be a compile-time constant so it doesn't look like a magic number everywhere.

src/toolkit/gis.cc
```diff
@@ -0,0 +1,158 @@
...
+
+void GIS::set_longitude(float lon) { longitude_ = SetPrecision(lon * 3600, 1); }
+
+double GIS::Distance(GIS target) const {
+  double curr_longitude = this->get_longitude() * M_PI / 180;
+  double curr_latitude = this->get_latitude() * M_PI / 180;
+  double target_longitude = target.get_longitude() * M_PI / 180;
+  double target_latitude = target.get_latitude() * M_PI / 180;
+  double dlong = target_longitude - curr_longitude;
+  double dlat = target_latitude - curr_latitude;
```
scopatz: It is especially odd to me that the distance upcasts to double if it is storing as a float.

src/toolkit/gis.h
```diff
@@ -0,0 +1,158 @@
+#ifndef CYCLUS_SRC_GIS_H_
+#define CYCLUS_SRC_GIS_H_
+
+#include <math.h>
+#include <stdio.h>
+#include <sstream>
```
scopatz: The above includes should be moved to the *.cc file so that compile times are reduced. The utilities that they include are not actually used in the header itself.

src/toolkit/gis.h
```diff
@@ -0,0 +1,158 @@
+#ifndef CYCLUS_SRC_GIS_H_
+#define CYCLUS_SRC_GIS_H_
```
scopatz: By the GCSG, this should be `CYCLUS_SRC_TOOLKIT_GIS_H_`

src/toolkit/gis.cc
```diff
@@ -0,0 +1,158 @@
...
+
+  double temp =
```
scopatz: temp and temp2 are not great variable names, please consider something more descriptive

scopatz:
Thanks a million for putting this in @gyutaepark! I think that this is a very valuable addition. Most of my comments related to style guide guide violations (see [the GCSG](https://google.github.io/styleguide/cppguide.html)) or top-level interface issues.

My major concern is over the use of floats rather than doubles. Is there anything that you can point me to that specifies that floats are sufficient? It seems the moment that they are used they get upcast to doubles, so why not use doubles?

Thanks for humoring my pedantic requests :)

src/toolkit/gis.h
gyutaepark: I think float should have enough precision for this case but I may be wrong. Originally, I decided to store latitude and longitude as float as for the same reason that jaime-olivares, has:

> The coordinate components will be stored as "seconds of degree" in float variables. This way, all coordinate elements (degrees, minutes, and seconds) will remain on the integral portion of values, with the exception of decimal of seconds, avoiding loss of precision. 
(https://github.com/jaime-olivares/coordinate)

The webpage also has a rationale against using float rather than double.

So, I think float should provide enough precision for this task. However, if I am wrong, please let me know and I will fix that right away.

src/toolkit/gis.h
gyutaepark: Thank you for pointing that out. I've used the term 'decimal degrees' because it was the term used to refer to the floating point representation of the coordinates by ISO 6709 but I can see that the term 'decimal degrees' is inaccurate in that sense.

Should I keep the term as it is or amend it to floating point degrees?

src/toolkit/gis.cc
gyutaepark: I've done this because I've read that calculations involving floating points can become inaccurate, especially, when I am multiplying by PI, or calculating the sin or cos so, I thought it would be better to use doubles here.

I'm new to programming so if these practices are poor or not recommended, please let me know. Thank you for the comments.

gyutaepark:
Thank you for you're comments. I've learnt a lot from it!
I've just pushed the changes.
If there are any other improvements or changes that need to be made, please let me know!

src/toolkit/gis.h
scopatz: Having read through his justifications about using float over double, some of them seem off to me.

First, the point you reference about loss of precision is only relevant for floating point arithmetic. Floating points need to undergo an operation in order for them to lose precision due to underflow or overflow of values. However, this comment is references storage, where the precision is constant.  That is, if you are storing (in memory or on disk) a floating point value, the precision is fixed by the number of mantissa bits that you chose to store.  

On the website, he claims that

> [Floats have] half the space and half the time to load in memory [of doubles].

This is likely not true for the standard systems out there.  Load time is dependent on the default word size of the processor and the OS. [Disclaimer: you always need to profile, of course :)] For a 64 bit system, the word size is 64 bits.  So a float is going to load in the same time as a double. Depending on how memory is aligned, a float could take up 64 bits too, with the last 32 bits being unused.  Furthermore, for most 64 bit architectures, floating point arithmetic is faster for doubles than it is for floats.  This is because these systems have been optimized for doubles and `float` operations require a cast to double first!

The place where floats truly are half the size and half the time to load into memory of doubles is in large arrays numbers.  However, that is not the situation we have here, where we have a couple of data members attached to a class.  

The last point, 

> The precision is enough compared to GPS precision (5-10 meters). A longitude with a maximum value of 180 degrees at the Equator that has 60 nautical miles per minute. When stored in a float variable, will have a minimum precision of about 2.4 meters.

This is pretty valid.  I am not sure we are GPS focused, but the question becomes, "Is 2.4 meters enough?" I don't really know how to answer this because it seems very problem-dependent. However, it does seem that if we just went with doubles, we wouldn't have to worry about whether 2.4 meters is enough.  By analogy, the precision of doubles is 4.4 nm, much smaller than anything we can conceive of worrying about for the time being.

src/toolkit/gis.cc
scopatz: Yeah, I think it is correct to perform the calculation as double precision.

src/toolkit/position.h
```diff
@@ -0,0 +1,187 @@
...
+/// example: 05.2169 -> 18780.84
+
+class Position {
+  static const int DecimalSecondsMultiplier = 3600;
+
+ public:
+  enum StringFormat {
+    D = 1,
+    DM,
+    DMS,
```
scopatz: I don't think these enum value names are particularly readable?  Maybe it is because I am not a GIS expert but can you give them longer, more descriptive names please? Thanks!

src/toolkit/position.h
```diff
@@ -0,0 +1,187 @@
...
+  ///                                         'Position::StringFormat::DMS':
+  ///
+  ///     Position::StringFormat::D  :   return in degrees format
+  ///     Position::StringFormat::DM :   return in degrees and minutes format
+  ///     Position::StringFormat::DMS:   return in egrees minutes seconds format
+  /// @endparblock
+  /// @return String representation of the Position object that complies with
+  /// ISO6709
+  /// Annex H
+  std::string ToString(Position::StringFormat format = StringFormat::D) const;
```
scopatz: GCSG, I don't think that there are spaces around the equals sign here.

src/toolkit/position.h
```diff
@@ -0,0 +1,187 @@
...
+/// Longitude and Latitude is stored as seconds of degrees. This was also
+/// adapted from 'https://github.com/jaime-olivares/coordinate'. This, according
+/// to the jaime-olivares, allows the coordinate elements such as degrees,
+/// minutes, and seconds to "remain on the integral portion of values, with the
+/// exception of decimal of seconds, avoiding loss of precision." This is
+/// calculated by multiplying decimal degrees by 3600.
+/// example: 05.2169 -> 18780.84
+
+class Position {
+  static const int DecimalSecondsMultiplier = 3600;
```
scopatz: Sorry, I think I wasn't clear enough. This could / should be a preprocessor macro constant.

```C
#define DECIMAL_SECOND_MULTIPLIER = 3600
```

scopatz:
Thanks @gyutaepark! I am pretty convinced that for the long-term API stability, it is probably best to have the Position class be double-based rather than float-based. See https://github.com/cyclus/cyclus/pull/1390#discussion_r121173735 Or alternatively, this could be a templated class that is templated on float/double/etc. Either way is fine with me.  Sorry for the run around.

src/toolkit/gis.h
gyutaepark: @scopatz 
just in case you missed this comment.

gyutaepark:
@scopatz 
Thank you. I just pushed the changes based on your review.

src/toolkit/position.h
```diff
@@ -0,0 +1,192 @@
+#ifndef CYCLUS_SRC_TOOLKIT_POSITION_H_
+#define CYCLUS_SRC_TOOLKIT_POSITION_H_
+#define DECIMAL_SECOND_MULTIPLIER 3600
```
scopatz: This macro should either be prefixed by `CYCLUS_` or moved to the *.cc file.

src/toolkit/position.h
```diff
@@ -0,0 +1,192 @@
...
+/// exception of decimal of seconds, avoiding loss of precision." This is
+/// calculated by multiplying decimal degrees by 3600.
+/// example: 05.2169 -> 18780.84
+
+class Position {
+ public:
+  enum StringFormat {
+    Degrees = 1,
+    DegreesMinutes,
+    DegreesMinutesSeconds,
```
scopatz: Please see the GCSG for enum naming rule https://google.github.io/styleguide/cppguide.html#Enumerator_Names  I prefer `UPPER_WITH_UNDERSCORES`

src/toolkit/position.cc
```diff
@@ -0,0 +1,175 @@
...
+  std::stringstream lat_string;
+  double lat_int;
+  lat_string << ToStringHelper(lat);
+
+  lat = modf(lat, &lat_int);
+  if (((int)lat_int) / 10 == 0) {
+    lat_string << "0";
+  }
+  switch (mode) {
+    case 1:
```
scopatz: Please use the enum names here to make this more readable

src/toolkit/position.cc
```diff
@@ -0,0 +1,175 @@
...
+  return earth_radius * angular_distance;
+}
+
+std::string Position::ToString(Position::StringFormat format) const {
+  std::stringstream lat_string;
+  std::stringstream lon_string;
+  double lat = this->latitude();
+  double lon = this->longitude();
+  switch (format) {
+    case 1:
```
scopatz: Please use the enum names here to make this more readable

src/toolkit/position.cc
```diff
@@ -0,0 +1,175 @@
...
+  lon_string << ToStringHelper(lon);
+  lon = modf(lon, &lon_int);
+  if (((int)lon_int) / 100 == 0) {
+    if (((int)lon_int) / 10 == 0) {
+      lon_string << "0";
+    }
+    lon_string << "0";
+  }
+  switch (mode) {
+    case 1:
```
scopatz: Please use the enum names here to make this more readable

scopatz:
Thanks @gyutaepark! Just a couple of minor issues, and then I think this is good to go!

scopatz:
This looks great, thanks @gyutaepark!

Issue #1389, opened 2017-06-01
Compile error on macOS from CMake flags


sskutnik:
I can no longer compile Cyclus on macOS Sierra, due to what appears to be unidentified compiler flags that XCode treats as errors. Specifically I get the following errors:

> [  0%] Building C object src/CMakeFiles/cyclus.dir/transmute.c.o
> clang: error: unknown argument: '-fcx-fortran-rules'
> clang: error: unknown argument: '-fcx-limited-range'
> clang: error: unknown argument: '-ftree-sra'
> clang: warning: optimization flag '-ftree-ter' is not supported [-Wignored-optimization-argument]
> clang: warning: optimization flag '-fexpensive-optimizations' is not supported [-Wignored-optimization-argument]

This appears to be introduced by the [introduction of Transmutagen](https://github.com/cyclus/cyclus/commit/6a7077def3dd772284ccd8009a618f52a47a5861) #1383, i.e., the offending flags are in src/CMakeLists.txt, lines 166-168, commited:

```
set_source_files_properties(transmute.c PROPERTIES COMPILE_FLAGS
    "-O0 -fcx-fortran-rules -fcx-limited-range -ftree-sra -ftree-ter -fexpensive-optimizations"
    )
```

Note that Ubuntu doesn't seem to have a problem with this, but clang doesn't know how to process these flags, which will default to an error state. Similarly, if I comment out the above-offending lines, it will compile on macOS (although I assume these flags are included for a specific performance reason).

scopatz:
Hi @sskutnik , thanks for pointing this out!  Out of curiosity, how long does it take to build with no flags installed? Does it seem to take a lot longer?

@asmeurer can you please figure out which flags should be used for clang and transmutagen?

scopatz:
A potential fix is in #1392

scopatz:
Note that the fix in #1392 is just a temporary solution to reallow transmutagen to compile. We still need the compiler flags for this @asmeurer :)

asmeurer:
I looked around, but I couldn't find anything that helps other than `-ffast-math`. Sadly, even with this, it is about 4x slower than with GCC. I tried `-O1`, but then the compilation doesn't finish. I'll keep looking to see if I can find anything else.

asmeurer:
`-O1` compilation eventually crashed with `fatal error: error in backend: IO failure on output stream.`

asmeurer:
[...]

PR #1392, opened 2017-06-08
Transmutagen working fix to compile with clang

scopatz:
This should address #1389, temporarily.


bam241:
works for me , thx @scopatz

Issue #1388, opened 2017-05-30
NameError: name 'cyclib' is not defined


bam241:
After reinstalling `cyclus` I got this message when running the unit test:

```
Exception ignored in: 'eventhooks.eventloophook'
Traceback (most recent call last):
  File "/Users/mouginot/.local/lib/python3.5/site-packages/cyclus/__init__.py", line 5, in <module>
    from cyclus.lib import py_init_hooks
ImportError: cannot import name 'py_init_hooks'
Exception ignored in: 'eventhooks.eventloophook'
Traceback (most recent call last):
  File "/Users/mouginot/.local/lib/python3.5/site-packages/cyclus/__init__.py", line 5, in <module>
    from cyclus.lib import py_init_hooks
ImportError: cannot import name 'py_init_hooks'
Exception ignored in: 'pymodule.py_del_agent'
NameError: name 'cyclib' is not defined
```

I don't know if this is related to #1366 or if this is a separated issue, or if this is related to my configuration...

bam241:
note that the unit_test are not failing.

katyhuff:
we think (thanks to @scopatz ) this is a deep python import error related to not having pandas.

scopatz:
And to @bam241 ;)

katyhuff:
This is closed via #1391 and #1394 .

PR #1394, opened 2017-06-09
Check for pandas if cython is found

### LINKED ISSUE #1391: Finding Pandas lib when Cython is present
`CMake` don't try to find Pandas lib when `Cython` is present

Pandas is listed as a required lib, but not checked during compilation

adityapb:
Fixes #1391 and #1388


scopatz:
Thanks a million @adityapb!

PR #1396, opened 2017-06-24
removed binary symbols checks

scopatz:
This removes the ABI stability checks.  This was never as robust as it would have needed to be and has caused more problems than it solved.  I think we should remove it, which is what this PR does. 

@FlanFlanagan @bam241 and others!


FlanFlanagan:
This PR works fine for me. Anyone else have thoughts?

katyhuff:
This makes sense and I am in support.

gonuke:
Dev call discussion (17/08/17) also tended to support removing this check from CI and the release procedure, but thought it might be useful to leave the tools in place for possible future use.

scopatz:
@gonuke so should I reinstate the tools but not the tests?

gonuke:
Yes - that's what we discussed.  Could we keep the test somewhere that is not invoked?  Or is there a simple way, with git, to discover the names of all the files that it ever tracked.  I worry about never being able to find this file if it only lives in the git history, and someone becomes curious about it.

scopatz:
Ok! Ihave updated the PR to deprecate, rather than remove the tests and utility.

scopatz:
Alright, it looks like this passes CI

scopatz:
@gonuke - Yes.  I don't think it holds any value after we deprecate. If and when someone reinstates this feature, they'll have to set a new baseline symbol collection.  This will diverge from the current and past symbol collections.  Maybe the prior symbol collections are interesting as a historical artifact, but in that case the file can be recovered from git history, to my mind.

bam241:
thx @scopatz.
merging!

scopatz:
Thanks all!

PR #1401, opened 2017-06-29
Fix for #1400

### LINKED ISSUE #1400: ModuleNotFoundError: No module named 'cyclus.lazyasd'
@adityapb 
Since #1393, I am getting a module not found error.
Seems like a simple fix by removing the import line from
```
actions.py", line 13, in <module>
    from cyclus.lazyasd import lazyobject
```

Currently running Ubuntu 16.04, with Python 3.6

gyutaepark:
This should fix #1400.


scopatz:
Thanks for putting this in @gyutaepark!

PR #1404, opened 2017-07-05
Update simstate.py

orpheus92:


scopatz:
Thanks for putting this in @orpheus92!

PR #1405, opened 2017-07-06
WIP: Improving  inventory table?

bam241:
This is an attempt to split-up the ExplicitInventory table from the database.
Have the inventory masses in one tables and the compositions in an other one, as the Transaction are written...

this is not ideal yet, I have PR this to get some feedback...

@scopatz @katyhuff @gonuke @FlanFlanagan


src/timer.h
```diff
@@ -97,6 +97,8 @@ class Timer {
   /// started.
   int time_;
 
+  int CompositionId;
```
scopatz: member variables should be named with `lower_case_and_underscores_` and have a trailing underscore for private vars

scopatz:
Hi @bam241 - thanks for putting this in and tackling this problem!  I think that this still fails to address what I see as the fundamental problem with material storage: each nuclide is stored in its own row.  

What I would like to see is a single inventory table that has columns for:

* resource id
* composition id
* the total mass
* a map<int, double> for the composition

bam241:
let's close this one...

bam241:
not this one, an other one :)

Issue #1408, opened 2017-08-09
Unable to find HDF5 Library


gyutaepark:
@scopatz 
I've tried to build Cyclus using the latest develop on Ubuntu 16.04 with anaconda installed. Unfortunately, there seems to be issues with finding the correct libraries, especially HDF5. I've tried the following versions of HDF5: 1.8.16, and 1.8.17.

This issue with building Cyclus did not exist when I tried building from `v1.4.2-release`.

Could you look into this please?

I've also attached a text file of the build message.
[return message.txt](https://github.com/cyclus/cyclus/files/1211431/return.message.txt)

scopatz:
Thanks for reporting @gyutaepark. What conda channels are you building based off of? I am using conda-forge, which I think has more updated versions of cmake and HDF5.

scopatz:
Also, did you do a clean build? Also what line did you use at the command line? Also do you have the `cyclus-build-deps` package installed? This is part of conda-forge and needed for building cyclus on top of a conda installation.

gyutaepark:
Oops sorry for the confusion. I'm building _CYCLUS_ from source. It's just that my python library is set up with Anaconda 3.

I've also tried a clean build but that did not help.
I am using the following line to build cyclus from source.
`python install.py --clean-build`

scopatz:
What does `conda list cyclus-build-deps` say?

gyutaepark:
Hmmm it returns nothing..
```
# packages in environment at /home/user/anaconda3:
#
```

scopatz:
right, so you should/need to install that package to be able to build Cyclus. I also recommend that you add the conda-forge channel to your condarc file

gyutaepark:
Alright will do! Thanks for the help. I was just confused because this has not happened before.

gyutaepark:
Installing `cyclus-build-deps` did work really well for me!
However, I'm still not sure why this error hasn't come up before or if this issue is just happening to me.
This was a clean Ubuntu (where I just installed Ubuntu for a new PC) and I've successfully installed Cyclus without having to install `cyclus-build-deps` through conda-forge.
Can someone that hasn't installed `cyclus_build~deps` try building from the latest develop to see if this issue is reproducible?
Or would it be better to just add this to the documentation?

gyutaepark:
Actually, it seems like installing `cyclus-build-deps` on conda would be a good practice in general.

gyutaepark:
[...]

PR #1409, opened 2017-08-10
Added instructions when installing Cyclus with conda libraries

gyutaepark:
This is related to #1408.
If this is unnecessary or out of place, please let me know.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -178,6 +178,18 @@ All variables can be set using ``-DMY_VARIABLE=MY_VARIABLES_VALUE``.
 
 .. website_custom_end
 
+Note on Conda
+-------------
+
+If your python libraries are installed using Conda, install cyclus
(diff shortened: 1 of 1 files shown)


INSTALL.rst
```diff
@@ -178,6 +178,18 @@ All variables can be set using ``-DMY_VARIABLE=MY_VARIABLES_VALUE``.
 
 .. website_custom_end
 
+Note on Conda
```
scopatz: Maybe change this to "Note on Building Cyclus with Conda".  I think that specificity helps here.

scopatz:
Thanks for putting this in and sorry for taking so long to get to it!

INSTALL.rst
gyutaepark: No worries. I'll update it.

Thank you.

PR #1413, opened 2017-09-02
remove unnecessary buggy line...

bam241:
@scopatz sorry I added a bad line in the deploy section....

this should be good... hopefully....


scopatz:
thanks!

PR #1415, opened 2017-09-07
Develop to master

scopatz:
Some transitions needed for migrating from develop to master. CC @bam241


scopatz:
OK it looks like this works!

bam241:
LGTM thx @scopatz

PR #1421, opened 2017-09-26
Fix for downstream segfaults in Unit tests

scopatz:
A segfault was discovered in ShortTermStorage that was being caused by the simulation information not being set up properly.


bam241:
LGTM! thx @scopatz

scopatz:
Thanks @bam241!

PR #1422, opened 2017-09-26
Filter out NaNs from decay heat calculation

scopatz:
Also, fixes 4-space indents to 2-space. CC @FlanFlanagan


FlanFlanagan:
Looks fine, by my concern is that we're band aiding an issue. This means that any nuclide that returns a NaN is not included, is it possible to miss a large chunk of decay heat because of the skipped nuclides?

PR #1428, opened 2017-10-10
helping ubuntu to find xml2

bam241:
this should fix the issue on ubuntu to find the xml2 libs...

thx @scopatz for the help on this one !


scopatz:
Thanks @bam241!

PR #1434, opened 2017-11-10
Timeseries

FlanFlanagan:
Adds the capability for an agent to request a notification when a RecordTimeSeries is called.  This is a way to allow agents to know, for example, the total power of a system without touching the output database.


scopatz:
Thanks @FlanFlanagan!

.gitignore
```diff
@@ -39,6 +39,8 @@ tests/libcyclus-orig.h5
 tests/libcyclus-orig.sqlite
 tests/libcyclus-test.h5
 tests/libcyclus-test.sqlite
+test/dummy.json
```
katyhuff: I think this should be plural: "tests/dummy.json" and "tests/dummy.h5" . (There is no "test" directory it's "tests".)

PR #1435, opened 2017-11-14
Boost variant

FlanFlanagan:
This allows the new time series recorder system to take more  than just the cyclus timeseriestypes.


scopatz:
Thanks @FlanFlanagan! Looks good to me!

PR #1437, opened 2017-11-20
prevents expected bytes got string error

FlanFlanagan:
fixes an issue with the string timeseries recorders


FlanFlanagan:
@scopatz

PR #1439, opened 2017-12-04
Fast compile for cyclus, from assembly

scopatz:
This adjusts Cyclus to use pyne's cram and bateman solvers directly. This enables fast compilation of these solvers from pre-generated assembly.


scopatz:
CC @cyclus/pushers

bam241:
it seems not to be working on my mac:
got:
```
/Users/mouginot/work/app/cyclus/cyclus/src/decay-clang.s:3:11: error: mach-o section specifier uses an unknown section type
 .section .text.startup,"ax",@progbits
          ^
/Users/mouginot/work/app/cyclus/cyclus/src/decay-clang.s:5:2: error: unknown directive
 .type __cxx_global_var_init,@function
```
```
/Users/mouginot/work/app/cyclus/cyclus/src/decay-clang.s:57:11: error: mach-o section specifier requires a segment whose length is between 1 and 16 characters
 .section .text._ZN6h5wrap24_get_PYTABLES_COMPLEX128Ev,"axG",@progbits,_ZN6h5wrap24_get_PYTABLES_COMPLEX128Ev,comdat
```
```
/Users/mouginot/work/app/cyclus/cyclus/src/decay-clang.s:976:2: error: unknown directive
 .size _ZN4pyne8decayers8decay_heEdRSt23_Rb_tree_const_iteratorISt4pairIKidEERSt3mapIidSt4lessIiESaIS4_EERA4141_d, .Lfunc_end10-_ZN4pyne8decayers8decay_heEdRSt23_Rb_tree_const_iteratorISt4pairIKidEERSt3mapIidSt4lessIiESaIS4_EERA4141_d
```
and the same with the `cram-clang.s` file


I build using:
```
git clean -fx
python install.py --clean-build -j4
```

coming soon with a gist with the full log

bam241:
the compilation did not yet failed but [here](https://gist.github.com/bam241/831bbbe034dbc5bf845e6fa09e337292) is a partial log

scopatz:
It seems like some of these are warnings, rather than failures:

```
[ 25%] Building ASM object src/CMakeFiles/cyclus.dir/decay-clang.s.o
9 warnings generated.
[ 26%] Building ASM object src/CMakeFiles/cyclus.dir/cram-clang.s.o
9 warnings generated.
```

Note that it looks like some of these seem like they are because I generated them on linux rather than a mac, which I don't have access to.  There should hopefully be a way to cross compile in order to remove those warnings.

bam241:
@scopatz shall I retry without the `-j 4`, to avoid overlap on the different msg ?

scopatz:
@bam241 that might be nice.

bam241:
https://gist.github.com/bam241/9d7b01e2d6fa4726621a65323200d904

bam241:
@scopatz [here](https://gist.github.com/bam241/9d7b01e2d6fa4726621a65323200d904) are the install log on 1 core

FlanFlanagan:
`[ 35%] Linking CXX shared library ../lib/libcyclus.so
CMakeFiles/cyclus.dir/decay-gnu.s.o: In function `pyne::decayers::decay(std::map<int, double, std::less<int>, std::allocator<std::pair<int const, double> > >, double)':
decay.cpp:(.text+0x145604c): multiple definition of `pyne::decayers::decay(std::map<int, double, std::less<int>, std::allocator<std::pair<int const, double> > >, double)'
CMakeFiles/cyclus.dir/pyne_decay.cc.o:pyne_decay.cc:(.text+0x1351a75): first defined here
CMakeFiles/cyclus.dir/decay-gnu.s.o:(.rodata+0x1e0): multiple definition of `pyne::decayers::all_nucs'
CMakeFiles/cyclus.dir/pyne_decay.cc.o:(.rodata+0x1e0): first defined here
/usr/bin/ld: Warning: size of symbol `_ZN4pyne8decayers8all_nucsE' changed from 15916 in CMakeFiles/cyclus.dir/pyne_decay.cc.o to 16564 in CMakeFiles/cyclus.dir/decay-gnu.s.o
collect2: error: ld returned 1 exit status
src/CMakeFiles/cyclus.dir/build.make:1944: recipe for target 'lib/libcyclus.so' failed
make[2]: *** [lib/libcyclus.so] Error 1
CMakeFiles/Makefile2:1070: recipe for target 'src/CMakeFiles/cyclus.dir/all' failed
make[1]: *** [src/CMakeFiles/cyclus.dir/all] Error 2
Makefile:162: recipe for target 'all' failed
make: *** [all] Error 2
`

This is the error that I get out of an install command. Looks like an issue in libcyclus

scopatz:
@FlanFlanagan - did you do a clean build? including the `git clean -fdx`

FlanFlanagan:
Fixed it. Thank you.

src/cram.hpp
```diff
@@ -0,0 +1,41 @@
...
+  double* decay_matrix;
+} pyne_cram_transmute_info_t;
+
+extern pyne_cram_transmute_info_t pyne_cram_transmute_info;
+
+int pyne_cram_transmute_ij(int i, int j);
+
+int pyne_cram_transmute_nucid_to_i(int nucid);
+
+
```
FlanFlanagan: Is the double new line here a style choice?

scopatz: Whoops thanks for noticing this!

scopatz:
Hi All, So I think that this is because I assembled the clang file on Linux, and so Clang on Mac is failing to compile them.  Also, as @asmeurer pointed out, we don't really want to use the Clang assemebly anyway (at least not for cram) because GCC does a much better job of optimizing the source code (Clang is around ~4x slower at run time).  

So there are a number of options to fix this ultimately:

1. Use GCC on mac - at least for the ASM, if it is availble.
2. Use YASM or similar cross-platform assembler, if GCC is not available 
3. Don't use assembly on Mac, and force a long compile.

For now, I have gone with #3, in part because I have no way to test 1 & 2 myself since I don't have a mac. Though 1 & 2 should be implemented soonish. 

@bam241 - would you please try this out again? Thanks!

scopatz:
Thanks for pointing this out @bam241. A potential fix is in.

scopatz:
...if you would please try it again :)

bam241:
I have a archetypes compilation running. I try this right after.

bam241:
LGTM ! 
@scopatz thx for putting this in.

scopatz:
Please feel free to merge!

PR #1443, opened 2017-12-05
Link against libraries, don't recompile everything

scopatz:
This fixes the build system so that ALL of the sources aren't recompiled into the unit tests. This should speed up clean build times considerably.

CC @FlanFlanagan @cyclus/pushers

### DIFF
--- a/agents/CMakeLists.txt
+++ b/agents/CMakeLists.txt
@@ -9,6 +9,8 @@ use_cyclus("agents" "prey")
 use_cyclus("agents" "predator")
 install_cyclus_module("agents" "" "NONE")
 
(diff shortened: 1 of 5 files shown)


scopatz:
Also, it will make the binaries smaller, which is better for packaging.

scopatz:
On my laptop, a clean compile now takes about 5 min with 4 threads in debug mode.

scopatz:
cycamore fix for this is available at https://github.com/cyclus/cycamore/pull/479

scopatz:
Test seem to pass now!

FlanFlanagan:
Merging this in.

scopatz:
Thank you so much @FlanFlanagan!

PR #1444, opened 2017-12-07
Master

FlanFlanagan:
This fixes an issue that the rng-print command was showing up sometimes when not requested. I'm not sure why this fixes the issue, and am happy to dig deeper if someone thinks that's appropriate.


scopatz:
OK this is insane that this fixes the problem for you, but I see no harm in it.

PR #1448, opened 2017-12-12
Python Agent Attribute Access

scopatz:
This fixes Python agents to enable standard attribute access, such as prototype and parent.

CC @FlanFlanagan @cyclus/pushers


FlanFlanagan:
This looks like it was a frustrating process. Thanks for putting it in!

PR #1455, opened 2018-01-25
Added dbopen in lib.pyx

robi111432:


cyclus/lib.pyx
```diff
@@ -312,6 +312,16 @@ cdef class _SqliteBack(_FullBackend):
         self.flush()  # just in case
         (<cpp_cyclus.SqliteBack*> self.ptx).Close()
 
+    def dbopen(fname):
```
scopatz: This should not be inside of a class, but rather a function at the top level

cyclus/lib.pyx
```diff
@@ -312,6 +312,16 @@ cdef class _SqliteBack(_FullBackend):
...
+        """Opens a Cyclus database."""
+        _, ext = os.path.splitext(fname)
+        if ext not in EXT_BACKENDS:
```
scopatz: EXT_BACKENDS also needs to be present

scopatz:
Thanks for putting this in @robi111432 . This also needs a test and news: http://fuelcycle.org/kernel/contributing_to_cyclus.html#changelog

PR #1458, opened 2018-03-03
fix for none values inventory capacities

scopatz:
CC @FlanFlanagan


bam241:
thx @scopatz 
does it work for the c++ agent as well ?

scopatz:
This problem doesn't arise in C++

PR #1461, opened 2018-03-05
Mocksim udpate

jbae11:
Adds an additional `MockSim::DummyProto` so that the user can define the capacity and commodity name of dummy prototype


src/mock_sim.cc
```diff
@@ -176,6 +176,20 @@ void MockSim::DummyProto(std::string name) {
   ctx_.AddPrototype(name, a);
 };
 
+void MockSim::DummyProto(std::string name, std::string commod, double capacity) {
```
scopatz: Doesn't this need to be declared in the header? Also where is it being called?

scopatz:
Thanks for putting this in @jbae11! Please add a news entry, as per http://fuelcycle.org/kernel/contributing_to_cyclus.html#changelog

jbae11:
I forgot to push the header file, my bad. Also, I don't see the `news/` directory. Is that in the cyclus directory?

scopatz:
The news directory is in the root level of the repo.  Maybe you need to update your branch from upstream master?

jbae11:
ah, it's updated! thanks

src/mock_sim.cc
```diff
@@ -176,6 +176,20 @@ void MockSim::DummyProto(std::string name) {
...
+  Agent* a = DynamicModule::Make(&ctx_, AgentSpec(":agents:Source"));
+
+  std::stringstream xml;
+  xml << "<facility><name>" << name << "</name>"
+      << "<config><foo>"
+      << "<commod>" << "commod" << "</commod>"
+      << "<capacity>" << capcity << "</capacity>"
```
scopatz: Looks like there is a typo here.  Middle one should be `capacity`

jbae11:
so sorry about that, I will make sure it builds next time!

scopatz:
Thanks @jbae11!

Issue #1468, opened 2018-03-19
Archetype tutorial issues 


gwenchee:
In the process of attempting the Cyclus archetype tutorial, two problems with the stubs generated from the Cycstub CLI have been found. In this issue, each problem is listed. @gyutaepark and I will work on fixing them. 

1) There is an error that occurs when trying to ``` python install.py ``` . This can be fixed by changing the Cmake file so that it will find Cyclus libraries. 

2) An error occurs when trying to run the example xml file. This is fixed by changing the archetype spec path to the path that leads to the ``` libarchetypetutorial.so ``` file where the term archetypetutorial is whatever you named the mylib when generating cycstub from the CLI.

scopatz:
Thanks for reporting!

gyutaepark:
Sorry for the delay and Thanks Gwen for making the issue.
I'll make a PR for this issue today.

gyutaepark:
#1470 has been merged. I believe the issue is now fixed. 
This issue will be closed.

PR #1470, opened 2018-03-21
Stub tutorial fix

gyutaepark:
This PR fixes issue #1468. 
- Changes to `CMakeLists.txt` may need a more thorough review to be safe @scopatz .

- There was an issue with stub_unit_tests where it fails to find the `libstub.so` file within `.local/lib/cyclus/libpath`. I tried to fix it so that the unit test could find the shared object file but could not do so. This was instead worked around by placing the *.so file along with the other cyclus so files in `.local/lib/cyclus` (this was done by editing the INSTALL_CYCLUS_MODULE macro.

- The README located in stub was also updated to reflect the changes./


scopatz:
Looks good to me!

scopatz:
Thanks @gyutaepark!

gonuke:
Do the changes here imply that the `<path>` component of an archetype specification is not working?

scopatz:
I think they just imply that the stubs were not installing into a sub-directory of the CYCLUS_PATH

gyutaepark:
@gonuke @scopatz 
Sorry for the lack of clarification.
When I run `python install.py`, executables are installed in `.local/bin` and library shared objects are installed in `.local/lib/cyclus/libpath` where lib_path is specified upon calling:

`cycstub --type facility libpath:libname:tutorialfacility`

I am not sure how I could test if the path component is working but for the moment being, cycstub_unit_tests is unable to locate its library shared object. Moving the .so file from libpath to `.local/lib/cyclus/` did fix this issue.

PR #1477, opened 2018-06-19
Time series

FlanFlanagan:
This adds the name of the time series listener to the listening function. Backwards compatible as it doesn't touch any other part of the code, and only d3ploy really utilizes the listeners right now.


tests/bear_deploy.py
```diff
@@ -60,7 +62,7 @@ def tick(self):
             self.context.schedule_build(self, proto)
 
 
-    def extract_supply(self, agent, time, value):
+    def extract_supply(self, agent, time, value, commod):
```
gwenchee: should this function return ` commodity_supply`

FlanFlanagan: I'm not sure what you mean here. This function doesn't return anything.

src/toolkit/timeseries.h
```diff
@@ -29,11 +29,11 @@ enum TimeSeriesType : int{
...
-    std::function<void(cyclus::Agent*, int, bool)>,
-    std::function<void(cyclus::Agent*, int, int)>,
-    std::function<void(cyclus::Agent*, int, float)>,
-    std::function<void(cyclus::Agent*, int, double)>,
-    std::function<void(cyclus::Agent*, int, std::string)>
+    std::function<void(cyclus::Agent*, int, bool, std::string)>,
+    std::function<void(cyclus::Agent*, int, int, std::string)>,
+    std::function<void(cyclus::Agent*, int, float, std::string)>,
+    std::function<void(cyclus::Agent*, int, double, std::string)>,
+    std::function<void(cyclus::Agent*, int, std::string, std::string)>
```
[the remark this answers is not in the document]
FlanFlanagan: just following the previous std, but I'm not sure it's harmful though.

FlanFlanagan:
@gwenchee @jbae11 

This needs a review.

tests/test_bear_deploy.py
```diff
@@ -31,7 +31,7 @@
     'config': {
      'NOInst': {
       'growth_commod': 'bears',
-      'growth_rate': 0.05,
```
jbae11: I think these input parameters have to be updated to match the updated input parameters for NOInst

tests/test_bear_deploy.py
```diff
@@ -19,8 +19,8 @@
    'config': {
     'DemandFac': {
      'commodity': 'bears',
-     'production_rate_max': 12,
-     'production_rate_min': 8,
+     'production_rate_max': 10,
+     'production_rate_min': 10,
```
katyhuff: What's this about? Why is the max=min ?

Issue #1479, opened 2018-07-20
MatlBuyPolicy Accepts materials to break buffer capacity limit:


jbae11:
The Cycamore::Storage archetype uses the MatlBuy /MatlSell Policy to aid users develop archetypes without having to define multiple functions for trading. However, I found out that if you set multiple `in_commods` for a single `buypolicy`, the archetype will accept all commodities with the requested amount equal to the buffer capacity.

A small error-outputting example is below. Here, `storage` should only accept 1e299 kg of `commodity2`, since that is the capacity of its inventory buffer. However, it accepts 1e299 kgs of both `commodity1` and `commodity2`, thus causing the simulation to fail, saying:
`resource pushing breaks capacity limit: space=0, rsrc->quantity()=1e+299`
```
<!-- 1 Source Source, 1 Sink Sink -->

<simulation>
  <control>
    <duration>10</duration>
    <startmonth>1</startmonth>
    <startyear>2000</startyear>
  </control>

  <archetypes>
    <spec>
      <lib>cycamore</lib>
      <name>Source</name>
    </spec>
    <spec>
      <lib>cycamore</lib>
      <name>Storage</name>
    </spec>
    <spec>
      <lib>agents</lib>
      <name>NullRegion</name>
    </spec>
    <spec>
      <lib>agents</lib>
      <name>NullInst</name>
    </spec>
  </archetypes>

  <facility>
    <name>Source1</name>
    <config>
      <Source>
        <outcommod>commodity1</outcommod>
        <outrecipe>commod_recipe</outrecipe>
      </Source>
    </config>
  </facility>
  <facility>
    <name>Source2</name>
    <config>
      <Source>
        <outcommod>commodity2</outcommod>
        <outrecipe>commod_recipe</outrecipe>
      </Source>
    </config>
  </facility>


    <facility>
        <name>storage</name>
        <config>
            <Storage>
                <in_commods>
                    <val>commodity1</val>
                    <val>commodity2</val>
                </in_commods>
                <in_commod_prefs>
                    <val>1.0</val>
                    <val>2.0</val>
                </in_commod_prefs>
                <out_commods>
                    <val>cooled_sfr</val>
                </out_commods>
            </Storage>
        </config>
    </facility>


  <region>
    <name>SingleRegion</name>
    <config><NullRegion/></config>
    <institution>
      <name>SingleInstitution</name>
      <initialfacilitylist>
        <entry>
          <prototype>Source1</prototype>
          <number>1</number>
[...]

PR #1480, opened 2018-07-20
Buypolicy fix

jbae11:
I was able to fix issue #1479 by changing the `GetMatlRequests` function.

However I'm not sure if there was a specific reason for using a 
`std::map<int, std::vector<Request<Material*>>`.
There doesn't seem to be a change in `Cycamore::Storage` archetype behavior.


FlanFlanagan:
What behavior does this result in? Using your example in #1479, does this result in only one facility sending material to the storage?

jbae11:
Without this fix, using the example in #1479 causes the storage to accept material from both facilties, causing its buffer to overflow, ending the simulation.
With the fix, the storage facility will accept material from only one facility.

FlanFlanagan:
@jbae11  can you make this up to date with master again? Thanks

PR #1484, opened 2018-09-07
fix mat_buy_pol when multiple commodities can be accepted

### LINKED ISSUE #1483: toolkit buy mat policy
I believe I find a bug in the `matl_buy_policy`.
if a facility using the `matl_buy_policy` can accept multiple commodities in some case the facility can get more than what its own capacity allows.
ie:
- `Storage` has a capacity of 10, and can accept commodity `c_src1` and `c_src2`
- `Src1` can produce 10 fo `c_src1`
- `Src2` can produce 10 fo `c_src2`

`Storage` will receive 10 of `c_src1` and 10 of `c_src2`
one example of this can be found [here](https://github.com/bam241/cyclus_play/blob/buy_mat_pol/sample.xml)

bam241:
this should fix #1483.

I am not quite sure why...

@FlanFlanagan

### DIFF
--- a/src/toolkit/matl_buy_policy.cc
+++ b/src/toolkit/matl_buy_policy.cc
@@ -160,9 +160,8 @@ std::set<RequestPortfolio<Material>::Ptr> MatlBuyPolicy::GetMatlRequests() {
     port->AddMutualReqs(grps[i]);
     ports.insert(port);
   }
-}
 
-return ports;
+  return ports;
 }
 
 void MatlBuyPolicy::AcceptMatlTrades(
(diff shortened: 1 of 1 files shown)


bam241:
maybe some unit test should be added to make sure this did not occur in the future....

src/toolkit/matl_buy_policy.cc
```diff
@@ -158,17 +157,10 @@ std::set<RequestPortfolio<Material>::Ptr> MatlBuyPolicy::GetMatlRequests() {
...
     }
-
-    // if there's more than one commodity, then make them mutual
-    if (grps.size() > 1) {
-      std::map<int, std::vector<Request<Material>*> >::iterator grpit;
-      for (grpit = grps.begin(); grpit != grps.end(); ++grpit) {
-        port->AddMutualReqs(grpit->second);
-      }
-    }
+    port->AddMutualReqs(grps[i]);
```
FlanFlanagan: The previous behavior seems to have caused this. Not sure why it was implemented in the first place. I imagine there is a good reason but I can't think of one at the moment.

FlanFlanagan:
@bam241 Could you at a unit test for this? I think that is a good idea.

FlanFlanagan:
It actually looks like this and #1480 might address the same issue?

bam241:
already dressed by #1480 
closing

Issue #1488, opened 2018-09-28
python `record_time_series` clashes with cycamore::Reactor


jbae11:
Calling `lib.record_time_series('POWER', self, power_cap)` in a python archetype clashes with the cycamore::Reactor ` cyclus::toolkit::RecordTimeSeries<cyclus::toolkit::POWER>(this, power_cap);`, and gives an error that says:
```
^^SQL error: CREATE TABLE TimeSeriesPOWER (SimId BLOB, AgentId INTEGER, Time INTEGER, Value REAL); table TimeSeriesPOWER already exists
```

This stems from the [cyclus/lib.pyx](https://github.com/cyclus/cyclus/blob/master/cyclus/lib.pyx) not recognizing `POWER` as a string, but  a TimeSeriesType.

FlanFlanagan:
Can you explain the issue more closely. 

Is it trying to create TimeSeriesPOWER for cycamore twice?

jbae11:
Yeah, I think it's more appropriate to have this issue in cyclus/cyclus, since it is more of a cyclus framework problem than a problem in Cycamore.  You can close this issue if you want!

FlanFlanagan:
issue moved to cyclus/cyclus

PR #1489, opened 2018-09-28
Python record timeseries edit

jbae11:
I edited the python interface so that the user can use the `record_times_series` function by passing a string as `tstype` for specific flags such as POWER, ENRICH_SWU. 

This may be from my ignorance, but the previous method was less intuitive for the python archetype developer to call.  For example, now once can record power by writing `lib.record_time_series('POWER', self, power_cap)`. This has not worked previously and clashed with the Cycamore::Reactor. (issue #1488).

Also, the if statement order is changed so that the default flags (`POWER', 'ENRICH_SWU`, `ENRICH_FEED`) is considered before the `value` instance checking.


FlanFlanagan:
Looks like you'll need to update the record timeseries tests. 

I'll need to look at the actual issue, but I'm not sure I understand the issue that is happening.

jbae11:
I can't say if this is necessarily a bug, but I think for developers it is more intuitive to pass a string for their `tstype` variable.

FlanFlanagan:
This change would break backwards compatibility with d3ploy right now and my PR for Cycamore. I also think it breaks Cyclus convention, but that's more of a personal thing. 

Also, this is still failing nosetests.  You need to update that before this can proceed. 

Was the order changed to accommodate these being strings now?

jbae11:
I think this is what I'm missing - why do we need special functions for POWER, ENRICH_SWU, and ENRICH_FEED? Why can't they be treated just like any other 'commodity', sort of how we do it in d3ploy?

FlanFlanagan:
It was mostly just to give developers a quick way to access those specific outputs. We had to updated a fair chunk of code to update cyclus to work the way it does with d3ploy. 

I'm not sure how to proceed. I understand your idea that it doesn't make as much sense any more. There might be some cymetric reasons to keep it like this. @bam241 ?

Other than that, I'm okay with the change as long as you also update the cycamore and d3ploy modules that are affected by this.

jbae11:
If you don't see a large problem with this, I'm fine as well. I guess there isn't a major need for this edit. Maybe this can be closed for now and maybe reopened when some developmental issue comes up. Thanks.

bam241:
I have no clue if this will change anything for cymetric. if it does, it can probably be fixed if I can get my hand on a working version of cycamore for this...
I can try it to be sure, but @jbae11  do I need a special version of cycamore ?

katyhuff:
I don't think this is necessary.

At the very least, making this change should include changes everywhere these things are:

https://github.com/cyclus/cyclus/blob/master/cyclus/cpp_cyclus.pxd#L920

PR #1490, opened 2018-11-19
adding a new typesystem: mapstringmapstringdouble 

gwenchee:
In this PR, i add a mapstringmapstringdouble type system to cyclus for future use in cyder and d3ploy


FlanFlanagan:
Looks good Gwen. Is it usable from the Python bindings ?

gwenchee:
I’ve only tried it with C++ so far. Will try to see if it works with python

src/query_backend.h
```diff
@@ -347,6 +347,24 @@ enum DbTypes {
...
+  VL_MAP_STRING_VL_MAP_STRING_DOUBLE,  // ["std::map<std::string, std::map<std::string, int>>", 3, ["HDF5", "SQLite"], ["VL_MAP", "STRING", ["VL_MAP", "STRING", "INT"]], true]
+  VL_MAP_STRING_MAP_VL_STRING_DOUBLE,  // ["std::map<std::string, std::map<std::string, int>>", 3, ["HDF5", "SQLite"], ["VL_MAP", "STRING", ["MAP", "VL_STRING", "INT"]], true]
+  VL_MAP_STRING_VL_MAP_VL_STRING_DOUBLE,  // ["std::map<std::string, std::map<std::string, int>>", 3, ["HDF5", "SQLite"], ["VL_MAP", "STRING", ["VL_MAP", "VL_STRING", "INT"]], true]
+  VL_MAP_VL_STRING_MAP_VL_STRING_DOUBLE,  // ["std::map<std::string, std::map<std::string, int>>", 3, ["HDF5", "SQLite"], ["VL_MAP", "VL_STRING", ["MAP", "VL_STRING", "INT"]], true]
+  VL_MAP_VL_STRING_VL_MAP_STRING_DOUBLE,  // ["std::map<std::string, std::map<std::string, int>>", 3, ["HDF5", "SQLite"], ["VL_MAP", "VL_STRING", ["VL_MAP", "STRING", "INT"]], true]
+  VL_MAP_VL_STRING_VL_MAP_VL_STRING_DOUBLE,  // ["std::map<std::string, std::map<std::string, int>>", 3, ["HDF5", "SQLite"], ["VL_MAP", "VL_STRING", ["VL_MAP", "VL_STRING", "INT"]], true]
```
bam241: I don't believe comments are correct here.

bam241:
@gwenchee  I try to draft a doc in the issue of the repo see #1508 
I think you need to fix the comment in the query_backend file and add an Update function corresponding to your type in the `Sha1` class (at end of the `query_backend.h`)

src/query_backend.h
bam241: `int` need to be `double` and I believe the 3 needs to be a 4 (4 variable length types in `map<string,map<string,double>>`)

katyhuff: @gwenchee : I think I agree with Baptiste. Can you contemplate and respond?

src/query_backend.h
```diff
@@ -352,6 +352,24 @@ enum DbTypes {
...
+
```
katyhuff: I think this blank line is extraneous.
```suggestion
```

src/query_backend.h
gwenchee: Sorry! I must have missed your ping @bam241, I will make appropriate changes to the comments now.

gwenchee:
I have updated the PR to include your suggested changes @bam241 :)

src/query_backend.h
```diff
@@ -1058,6 +1075,14 @@ class Sha1 {
     }
   }
 
+  inline void Update(const std::map<std::string, std::map<std::string, double>>& x) {
```
katyhuff: This function would be improved by a docstring explaining what it's doing (especially since it has such a vague name.)

katyhuff: Nevermind, I see now that this is the pattern. 

However, notably, your function is more complex than the others. I wonder if perhaps a little documentation regarding why this update function works the way it does ... would be helpful.

gwenchee: To be honest, I'm also not sure what it does but it is done for each typesystem and so i followed the pattern @katyhuff

gwenchee: The function that I added is copied from the type system `mapstringmapstringint` in line 1053 onwards

src/query_backend.h
```diff
@@ -352,6 +352,23 @@ enum DbTypes {
...
+  VL_MAP_STRING_VL_MAP_STRING_DOUBLE,  // ["std::map<std::string, std::map<std::string, double>>", 3, ["HDF5", "SQLite"], ["VL_MAP", "STRING", ["VL_MAP", "STRING", "DOUBLE"]], true]
+  VL_MAP_STRING_MAP_VL_STRING_DOUBLE,  // ["std::map<std::string, std::map<std::string, double>>", 3, ["HDF5", "SQLite"], ["VL_MAP", "STRING", ["MAP", "VL_STRING", "DOUBLE"]], true]
+  VL_MAP_STRING_VL_MAP_VL_STRING_DOUBLE,  // ["std::map<std::string, std::map<std::string, double>>", 3, ["HDF5", "SQLite"], ["VL_MAP", "STRING", ["VL_MAP", "VL_STRING", "DOUBLE"]], true]
+  VL_MAP_VL_STRING_MAP_VL_STRING_DOUBLE,  // ["std::map<std::string, std::map<std::string, double>>", 3, ["HDF5", "SQLite"], ["VL_MAP", "VL_STRING", ["MAP", "VL_STRING", "DOUBLE"]], true]
+  VL_MAP_VL_STRING_VL_MAP_STRING_DOUBLE,  // ["std::map<std::string, std::map<std::string, double>>", 3, ["HDF5", "SQLite"], ["VL_MAP", "VL_STRING", ["VL_MAP", "STRING", "DOUBLE"]], true]
+  VL_MAP_VL_STRING_VL_MAP_VL_STRING_DOUBLE,  // ["std::map<std::string, std::map<std::string, double>>", 3, ["HDF5", "SQLite"], ["VL_MAP", "VL_STRING", ["VL_MAP", "VL_STRING", "DOUBLE"]], true]
```
bam241: I think it should be a 4 in the comment, as you have 4 VL variables

src/query_backend.h
bam241: those Update method don't have a docstring... there is a small one line 715...

src/query_backend.h
gonuke: I'm not sure we all understand this parameter correctly - see my comments in #1509

gwenchee: I'm not sure what you mean @gonuke. How does adding metadata relate to this PR?

gonuke: I should have been more precise - the comments to which I was referring were specifically about the correct values of this quantity.  We discussed it on yesterday's call and I think your change makes it correct

gonuke:
@bam241 how does this pass the news test?

bam241:
Don't know :( 
Not supposed to...

gwenchee:
I've added the news file @bam241 @gonuke

newtypesystem
```diff
@@ -0,0 +1,12 @@
...
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:** None
+
+**Security:** None
```
bam241: @gwenchee I did not notice this before...
could you actually place this file in news folder ?
and name it something like `typesystem.rst` ?

gwenchee: @bam241 done! Sorry that was what I intended to do.

gonuke:
It looks like all of @katyhuff comments are resolved, so I'll merge.

Thanks for this addition @gwenchee

PR #1496, opened 2019-02-20
bunch of fixes

bam241:
Allows to build a conda recipe for cyclus
Also address deprecation from boost 1.68 also.

change in `src/OsiCbcSolverInterface.cpp` were required to build on `condaforge/linux-anvil-comp7`, I don't know if it will break the linux build... 

the change in `cycpp.py` and in `UseCyclus` should be transparent unless you have a `$CPP` variable setup on your system, if so `cmake` will use this value in place of `cpp` executable...

### DIFF
--- a/cmake/UseCyclus.cmake
+++ b/cmake/UseCyclus.cmake
@@ -89,7 +89,11 @@ MACRO(USE_CYCLUS lib_root src_root)
 
 
     # set cpp path
-    
+    # Now use by default the $CPP enviromnement variable first, if not fallback
+    # on the previous behavior, so in order:
+    #   1- if defined use ${CPP}
+    #   2- if clang++ present use it
+    #   3- otherwise use cpp
     IF(DEFINED ENV{CPP})
         SET(SYS_CPP "$ENV{CPP}")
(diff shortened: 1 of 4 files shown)


bam241:
@katyhuff @gonuke @FlanFlanagan could you look into this, I probably need this to fix the conda builds :)

cmake/UseCyclus.cmake
```diff
@@ -89,11 +89,15 @@ MACRO(USE_CYCLUS lib_root src_root)
 
 
     # set cpp path
-    IF("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-        SET(PREPROCESSOR "--cpp-path=clang++")
+    
+    IF(DEFINED ENV{CPP})
```
gonuke: add a comment to explain what's happening here

bam241: @gonuke does it look better ?

FlanFlanagan:
Looks good to me.

bam241:
ping @gonuke

cmake/UseCyclus.cmake
katyhuff: lgtm

Issue #1500, opened 2019-03-11
Map structure in python does not work 


gwenchee:
The map structure for python archetypes does not work. Details on the errors is provided here: https://github.com/arfc/d3ploy/issues/33. PR

gwenchee:
https://github.com/cyclus/cyclus/pull/1501 introduces a workaround to make a single level map structure work. However, multi-level map structures still don't work. And map structures that have vectors within them also do not work.

PR #1501, opened 2019-03-11
FIX: Single level map structure python

gwenchee:
This PR adds a work around to make a single level map structure work in the python interface. This partially addresses https://github.com/cyclus/cyclus/issues/1500, however, does not fix the entire issue.


katyhuff:
This is great - it'll enable the missing piece to the python map situation. 
Very confusing why cymetric test is failing -- good luck figuring it out! 
Maybe something pythonic is happening in the dataframe evaluations in cymetric that relies on the gentypesystem ... or maybe the @dbtests decorator is doing something fancier than it appears...

PR #1502, opened 2019-03-15
Test

gwenchee:


bam241:
@gwenchee  I think the problem comes from cymetric, for whatever reason, the last merged PR, broken cymetric nosetests as well

gwenchee:
Okay, thanks baptise. I was wondering why this was happening. What should we do? Should I wait for the cycmetric nose tests to pass before merge #1501

FlanFlanagan:
Is this from the last Cymetric PR or Cyclus PR?

bam241:
in theory the `cymetric` test, are there only for information, meaning that we have something to fix with `cymetric`, so I would say no need to wait... but may be make a issue in the `cymetric` repo about it :)

bam241:
@FlanFlanagan I don't know, it seems that it just appear....

PR #1509, opened 2019-05-16
[WIP] adding Metadata, and some other change

bam241:
This is an attempt to allow agent to have metadata.
I inspired myself from the position.cc/.h, and I had some though on how to improve it....

1. `Position` change:
The way the `Position` is working now, each agent have to define there `coordinates` (all the `cycamore` have them) and have there own method to `Record` them (all have the same), what I did, is add the `coordinates` and the RecordPosition at the Agent level (each cycamore agent still have to get latitude and longitude on their own), each agent calling the `RecordPosition` at the `Cyclus::Agent::EnterNotify()` level, avoiding copy/pasting the RecordPosition method over and over...

I also added a `RecordPosition(Agent* agent)` method in Position, so the `Position` knows how to register itself in the database. (this add agent and context include in the Position, if we don't like it, it will be as easy to add the `RecordPosition` in the agent itself...)
I think this last change make sense, as it allows to keep the `Agent` class fairly simple and will allow each attribute (Position) to know how to Record himself...

2. `Metadatas`:
  - `Metadata`:
Allows user to add metadata as it is done in the `pyne::Material`, mapping a keyword (`string`) with almost any value and record them in a Metadata table for the output file that looks like:
`SimId | AgentId | Keyword | Value | Type`, the `Value` will be written as a `string`, the `Type` column allows to identify it,
  - `UsageMetadatas`:
Yet some kind of metadata (but the more I think about it, the less I believe it should inherit from `Metadata`).
The UsageMetadata are supposed to record the amount of some usage of any kind per deployment, per decomission, per timestep and per throughput. Each type of usage will have its own keyword (user defined) and to each keyword 4 different usages could be associated.
example of usage: manpower, CO2, water, lollipop, electricity, truck, pretzels,...
the `UsageMetadata` Table will look like:
`SimId | AgentId | Keyword | Value | Usage` (I might add a unit here...)

For the RecordMetadata and RecordUsageMetadata, I plan to implement it the same way I did for `Position`, (already did it for the UsageMetadata...)

this will be followed by a PR on the `cycamore` side.

To be added to this PR:
- [ ] add Metadata to Cyclus::Agent
- [ ] unit test for all

The big questions are:
-----------------------
 - Do you like the implementation change I am introducing ???
- Is it CEP worthy ?


gonuke:
"data" is plural so there is no "metadatas"

src/agent.cc
```diff
@@ -151,6 +151,10 @@ void Agent::Build(Agent* parent) {
 
 void Agent::EnterNotify() {
   ctx_->RegisterAgent(this);
+
+  usagesdata.RecordMetadatas(this);
+  metadata.RecordMetadatas(this);
+
```
katyhuff:
```suggestion
```

src/agent.cc
```diff
@@ -151,6 +151,10 @@ void Agent::Build(Agent* parent) {
 
 void Agent::EnterNotify() {
   ctx_->RegisterAgent(this);
+
+  usagesdata.RecordMetadatas(this);
```
katyhuff: "usages" is probably supposed to be "usage"
```suggestion
  usagedata.RecordMetadata(this);
```

src/agent.h
```diff
@@ -483,7 +485,16 @@ class Agent : public StateWrangler, virtual public Ider {
   /// an instance-unique ID for the agent
   int id_;
 
+
+
```
katyhuff: How many blank lines are needed here? Seems like too many.
```suggestion
```

src/agent.h
```diff
@@ -483,7 +485,16 @@ class Agent : public StateWrangler, virtual public Ider {
...
   Context* ctx_;
+  
+// Additionnal Data that are not required for a Cyclus Run
```
katyhuff:
```suggestion
// Additional data that are not required for a Cyclus run
```

src/agent.h
```diff
@@ -483,7 +485,16 @@ class Agent : public StateWrangler, virtual public Ider {
...
+ protected :
+  /// agent Metadata
+  cyclus::toolkit::Metadatas metadata;
```
katyhuff:
```suggestion
  cyclus::toolkit::Metadata metadata;
```

src/agent.h
```diff
@@ -483,7 +485,16 @@ class Agent : public StateWrangler, virtual public Ider {
...
+  /// agent Usage Metadata
+  cyclus::toolkit::UsageMetadatas usagesdata;
```
katyhuff:
```suggestion
  cyclus::toolkit::UsageMetadata usagedata;
```

src/agent.h
```diff
@@ -16,6 +16,8 @@
 #include "query_backend.h"
 #include "resource.h"
 #include "state_wrangler.h"
+#include "toolkit/metadatas.h"
```
katyhuff:
```suggestion
#include "toolkit/metadata.h"
```

src/agent.h
katyhuff: Slightly unclear what this comment is supposed to convey.

src/cyclus.h
```diff
@@ -56,6 +56,7 @@ extern "C" {
 #include "toolkit/commodity_recipe_context.h"
 #include "toolkit/enrichment.h"
 #include "toolkit/infile_converters.h"
+#include "toolkit/metadatas.h"
```
katyhuff:
```suggestion
#include "toolkit/metadata.h"
```

src/query_backend.h
```diff
@@ -192,14 +192,14 @@ enum DbTypes {
...
-  VL_MAP_VL_STRING_PAIR_DOUBLE_MAP_INT_DOUBLE,  // ["std::map<std::string, std::pair<double, std::map<int, double>>>", 4, ["HDF5", "SQLite"], ["VL_MAP", "VL_STRING", ["PAIR", "DOUBLE", ["MAP", "INT", "DOUBLE"]]], true]
-  VL_MAP_STRING_PAIR_DOUBLE_VL_MAP_INT_DOUBLE,  // ["std::map<std::string, std::pair<double, std::map<int, double>>>", 4, ["HDF5", "SQLite"], ["VL_MAP", "STRING", ["PAIR", "DOUBLE", ["VL_MAP", "INT", "DOUBLE"]]], true]
-  MAP_VL_STRING_PAIR_DOUBLE_VL_MAP_INT_DOUBLE,  // ["std::map<std::string, std::pair<double, std::map<int, double>>>", 4, ["HDF5", "SQLite"], ["MAP", "VL_STRING", ["PAIR", "DOUBLE", ["VL_MAP", "INT", "DOUBLE"]]], false]
-  VL_MAP_VL_STRING_PAIR_DOUBLE_VL_MAP_INT_DOUBLE,  // ["std::map<std::string, std::pair<double, std::map<int, double>>>", 4, ["HDF5", "SQLite"], ["VL_MAP", "VL_STRING", ["PAIR", "DOUBLE", ["VL_MAP", "INT", "DOUBLE"]]], true]
+  MAP_STRING_PAIR_DOUBLE_MAP_INT_DOUBLE,  // ["std::map<std::string, std::pair<double, std::map<int, double>>>", 3, ["HDF5", "SQLite"], ["MAP", "STRING", ["PAIR", "DOUBLE", ["MAP", "INT", "DOUBLE"]]], false]
```
katyhuff: Wait... is this really supposed to be 3... or is it supposed to be 4.... I'm not looking closely at these comments, and I don't quite know what the number is supposed to indicate.

src/toolkit/metadatas.cc
```diff
@@ -0,0 +1,223 @@
+#include <math.h>
```
katyhuff: Replace metatdatas with metadata throughout this file, including the filename. Same with Metadatas (should be Metadata).

src/toolkit/metadatas.cc
```diff
@@ -0,0 +1,223 @@
...
+void Metadatas::RecordMetadatas(Agent* agent) {
+  std::string tblname = "Metadatas";
+  Json::Value::Members keys = metadatas.getMemberNames();
+  Json::Value::Members::const_iterator ikey = keys.begin();
+  Json::Value::Members::const_iterator ikey_end = keys.end();
+  for (; ikey != ikey_end; ++ikey) {
+    std::string value = "";
+    std::string type = "";
+    switch (metadatas[*ikey].type()) {
+      case Json::nullValue:
```
katyhuff: is the json namespace really capitalized like this?

src/toolkit/metadatas.cc
```diff
@@ -0,0 +1,223 @@
...
+        break;
+      case Json::stringValue:
+        type = "string";
```
katyhuff: I feel like this loop could be more concise.. it seems very verbose at the moment. Maybe some structure exists in boost that could be helpful?

src/toolkit/metadatas.cc
```diff
@@ -0,0 +1,223 @@
...
+        type = "double";
+        value = std::to_string(metadatas[*ikey].asDouble());
+        break;
+      case Json::arrayValue:
+      case Json::objectValue:
+        ValueError("Type is not convertible to string");
+      default:
+        ValueError("Type is not known.");
+    }
+    if (value != "") {
```
katyhuff: I think that `value != ""` will evaluate as true even if value is Null. Maybe worth a test?

src/toolkit/metadatas.cc
```diff
@@ -0,0 +1,223 @@
...
+          ->AddVal("AgentId", agent->id())
+          ->AddVal("keyword", *ikey)
+          ->AddVal("Value", value)
+          ->AddVal("Type", type)
+          ->Record();
+    }
+  }
+}
+
+void Metadatas::LoadData(std::map<std::string, std::string> datas) {
```
katyhuff:
```suggestion
void Metadatas::LoadData(std::map<std::string, std::string> data) {
```

katyhuff: (throughout)

src/toolkit/metadatas.cc
```diff
@@ -0,0 +1,223 @@
...
+          metadatas[keyword] = uint(std::stoul(value));
+          break;
+
+        case d:
+          metadatas[keyword] = std::stod(value);
+          break;
+
+        default:
+          std::stringstream msg;
+          msg << "Allowed Usage Keyw are:"
```
katyhuff:
```suggestion
          msg << "Allowed usage keywords are:"
```

src/toolkit/metadatas.cc
```diff
@@ -0,0 +1,223 @@
...
+              << " deploymemt, decomision, timestep and throughput";
```
katyhuff:
```suggestion
              << " deployment, decommission, timestep, and throughput";
```

src/toolkit/metadatas.cc
```diff
@@ -0,0 +1,223 @@
...
+          throw ValueError(msg.str());
+    }
+    } else {
+      ValueError("type encoding not reconised");
```
katyhuff:
```suggestion
      ValueError("type encoding not recognized");
```

src/toolkit/metadatas.h
```diff
@@ -0,0 +1,62 @@
...
+  /// container for arbitrary metadata, following the JSON rules.
+  Json::Value metadatas;
+
+
+};
+
+
+class UsageMetadatas: public Metadatas {
+ public:
+  /// The default constructor for Metadatas.
```
katyhuff:
```suggestion
  /// The default constructor for UsageMetadatas. 
```

src/toolkit/metadatas.h
```diff
@@ -0,0 +1,62 @@
...
+  ~UsageMetadatas() {};
+
+  void LoadData(std::map<std::string, std::map<std::string, double >> data);
+
+  // write the metadata in the output table
+  void RecordMetadatas(Agent* agent); 
+  // check consistency in the keyword names
+  void CheckConstistency();
+
+
```
katyhuff:
```suggestion
```

src/toolkit/metadatas.h
```diff
@@ -0,0 +1,62 @@
...
+  /// The default destructor for Metadatas
+  ~UsageMetadatas() {};
+
+  void LoadData(std::map<std::string, std::map<std::string, double >> data);
+
+  // write the metadata in the output table
+  void RecordMetadatas(Agent* agent); 
+  // check consistency in the keyword names
+  void CheckConstistency();
+
```
katyhuff:
```suggestion
```

src/toolkit/metadatas.cc
katyhuff: (and datas->data and METADATAS->METADATA)

src/toolkit/metadata.cycpp.h
```diff
@@ -0,0 +1,33 @@
+
+// This includes the required header to add geographic coordinates to a
```
gonuke: Is this a cut & paste error - this comment is about metadata and not geo coordinates

src/toolkit/metadata.cc
```diff
@@ -0,0 +1,174 @@
...
+    std::make_pair("d", d)   // double
+};
+
+const std::list<std::string> usages = std::list<std::string>(
+    {"decommission", "deployment", "timestep", "throughput"});
+
+Metadata::Metadata() {}
+Metadata::~Metadata() {}
+
+void Metadata::RecordMetadata(Agent* agent) {
```
gonuke: I'm not sure this is the best function name.  Maybe `RegisterMetadata`?  No recording of metadata happens here, this function simply adds it to the output tables to be recorded in the future.

gonuke: I suppose you are basing it on the naming of `Datum::Record()` which is perhaps also poorly named...  Although in that case one can interpret it as "start recording"

src/toolkit/metadata.cc
```diff
@@ -0,0 +1,174 @@
...
+          throw ValueError(msg.str());
+      }
+    } else {
+      ValueError("type encoding not recognized");
+    }
+  }
+}
+
+void Metadata::LoadData(
+    std::map<std::string, std::map<std::string, double>> datas) {
```
gonuke: `datas` should be singular

src/toolkit/metadata.cc
```diff
@@ -0,0 +1,174 @@
...
+  for (auto keyword_datas : datas) {
```
gonuke: `keyword_datas` should be singular

src/query_backend.h
katyhuff: @gwenchee Can you make some thoughts or comments that clarify the meaning of this number (I know you have worked with this before).

katyhuff: @gwenchee nvm, I think we have arrived at an understanding of this with @bam241 .

Issue #1508, opened 2019-05-15
?Missing? doc to add a new type


bam241:
to add a new supported type in Cyclus, needs to be done:
- add you type in the `DbTypes` enum list in src/query_backend.h
   - each variation length type (used in you custom type) need to be both as `TYPE` and `VL_TYPE`
     - each line need a comment formatted as follow: 
        ` // {"c++ type (with namespacing etc...)", NB_OF_VL_TYPE, ["HDF5","SQLite"],["CYCLUS TYPE DEFINITION"],true/false]`
     - the "CYCLUS TYPE DEFINITION" should be:
        - for `std::map<X,Y>` : `["MAP", "X", "Y"]` or `["VL_MAP","X","Y"]`
        - for a `std::vector<Z>`: `["VECTOR","Z"]` or `["VL_VECTOR","Z"]`
        - ...
     - I am not sure what the `true` or `false` rule is...
     - HDF5 and/or Sqlite should appear depending on needs (at least one is required)

- add the `inline void Update(` method at the end of the `src/query_backend.h` file in the `Sha1` class

- for SQLite type support (i.e. if you added "SQLite" in the comment) edit `src/sqlite_back.cc` in 3 locations:
    - add you method in `#define CYCLUS_BINDVAL(D, T)`:
       - add a method mapping your new type with Cyclus naming:
           - `,` because `CYCLUS_COMMA` when describing the cpp type
           - cyclus naming is all capital underscore separation:
                  - std::map<std::string, std::vector<double>>` -> `MAP_STRING_VECTOR_DOUBLE`
    - the same way add you method in `#define CYCLUS_LOADVAL(D, T)`
    
    - add you type in `DbTypes SqliteBack::Type(` :
      - `type_map[&typeid(FULL_CPP_TYPE_WITH_NAMESPACE)] = CYCLUS_NAME_TYPE`
      - in this case the CYCLUS_NAME_TYPE is the one described just before
                  - std::map<std::string, std::vector<double>>` -> `MAP_STRING_VECTOR_DOUBLE`

- for HDF5 it should be done automaticity (hopefully), a `git clean -fx` might be required (clear all file not followed by git)

with that you should be good to go !

bam241:
(Note this might not be complete, feel free to add anything I might have missed

gwenchee:
An example of the addition of a new type can be seen in https://github.com/cyclus/cyclus/pull/1490

PR #1510, opened 2019-05-20
Position implementation change

bam241:
`Position` change:
The way the `Position` is working now, each agent have to define their `coordinates` (all the `cycamore` have them) and have there own method to `Record` them (all have the same), what I did, is add the `coordinates` and the RecordPosition at the Agent level (each cycamore agent still have to get latitude and longitude on their own), each agent calling the `RecordPosition` at the `Cyclus::Agent::EnterNotify()` level, avoiding copy/pasting the RecordPosition method over and over...

I also added a `RecordPosition(Agent* agent)` method in Position, so the `Position` knows how to register itself in the database. (this add agent and context include in the Position, if we don't like it, it will be as easy to add the `RecordPosition` in the agent itself...)
I think this last change make sense, as it allows to keep the `Agent` class fairly simple and will allow each attribute (Position) to know how to Record himself...

This is redondant with #1508 , but @gonuke advised me to do a separate PR for this change.


gonuke:
A better way to streamline this without forcing every agent to have a `Position` (since that's not in the spirit of the toolkit):
#. add `RecordPosition()` to `position.h` (as already one here)
#. all agents with a position do the following things
    * inherit from `Toolkit::Position`
    * `#include` a special snippet that adds variables with cyclus pragmas in header (e.g. `position.cycpp`)
    * add `Position(latitude,longitude)` to the initializer list of their constructor
    * call `RecordPosition(this)` from `EnterNotify()`

src/toolkit/position.cycpp
```diff
@@ -0,0 +1,19 @@
+#pragma cyclus var { \
```
katyhuff: Is this file really intended to be part of the PR?

src/toolkit/position.h
```diff
@@ -3,11 +3,13 @@
 #define CYCLUS_DECIMAL_SECOND_MULTIPLIER 3600
 
 #include <string>
-
 #include "cyclus.h"
 
+
+
```
katyhuff: Three seems like a lot of blank lines at once. (Though, I don't remember the style guide rule on this). 
```suggestion
```

src/toolkit/position.h
```diff
@@ -128,6 +130,9 @@ class Position {
   std::string ToString(
       Position::StringFormat format = StringFormat::DEGREES) const;
 
+
```
katyhuff: Should there be 2 blank lines here, or only one?
```suggestion
```

tests/toolkit/position_tests.cc
```diff
@@ -1,7 +1,7 @@
 #include <gtest/gtest.h>
-
 #include <iostream>
 #include <string>
+
```
katyhuff: Why add this blank line here in this place?
```suggestion
```

src/toolkit/CMakeLists.txt
```diff
@@ -24,8 +24,10 @@ endif()
 SET(CYCLUS_CORE_SRC ${CYCLUS_CORE_SRC} ${cc_files} PARENT_SCOPE)
 
 FILE(GLOB h_files "${CMAKE_CURRENT_SOURCE_DIR}/*.h")
+FILE(GLOB cycpp_files "${CMAKE_CURRENT_SOURCE_DIR}/*.cycpp")
```
katyhuff: I'm not sure I understand this part of this PR.

bam241:
@katyhuff thx for your review !
To reply to your comment, the idea here is to include the position.cycpp in the different archetype we develop (`#include toolkit/position.cycpp`) to avoid having to copy paste over and over:
```
#pragma cyclus var { \
"default": 0.0, \
"uilabel": "Geographical latitude in degrees as a double", \
"doc": "Latitude of the agent's geographical position. The value should " \
       "be expressed in degrees as a double." \
}
double latitude;

#pragma cyclus var { \
"default": 0.0, \
"uilabel": "Geographical longitude in degrees as a double", \
"doc": "Longitude of the agent's geographical position. The value should " \
       "be expressed in degrees as a double." \
}
double longitude;
```

so adding position on an archetype, one will only have to:
- inherit from Position
- `#include position.cycpp`
- call the `position` constructor for the archetype constructor
- call the `RecorPosition(this)` from the EnterNotify() of the archetype

bam241:
you can see the related PR draft on cyclus/cycamore#502

bam241:
@gonuke @katyhuff I believe this is ready for a second review :)

FlanFlanagan:
@bam241 Are we concerned about the cymetric failures? It seems like the tests are returning larger tables than expected.

bam241:
@FlanFlanagan I did not investigate this yet, but this is just showing that Cyclus has evolve to something that `cymetric` don't understand.
So I am not concern about it, but we need to PR a fix for symmetric.
the `cymetric_master` and the `cycamore_master` are just there as highlight that the impact of a cyclus change on cycamore and cymetric

tests/toolkit/position_tests.cc
gonuke: I think we tend to separate standard includes from project includes (style guide?). But there is an opposite change above...

src/toolkit/position.cycpp.h
```diff
@@ -0,0 +1,32 @@
+
+// This includes the required header to add geographic coordinates to a
+// archetypes.
+// One only need to:
+// - '#include "toolkit/position.cycpp"' in the core of the archetype class (as
```
gonuke:
```suggestion
// - '#include "toolkit/position.cycpp.h"' in the core of the archetype class (as
```

src/toolkit/position.cycpp.h
```diff
@@ -0,0 +1,32 @@
+
+// This includes the required header to add geographic coordinates to a
+// archetypes.
```
gonuke: I thinks we need a longer message here that explains how it all works, since it is a brand new concept,

Also, can we use one of the archetypes that’s in `cyclus/cyclus` to demonstrate?

src/toolkit/position.cycpp.h
```diff
@@ -0,0 +1,32 @@
...
+
+#pragma cyclus var { \
+"default": 0.0, \
+"uilabel": "Geographical latitude in degrees as a double", \
+"doc": "Latitude of the agent's geographical position. The value should " \
+       "be expressed in degrees as a double." \
+}
+double latitude;
+// required for compilation but not added by the cycpp preprocessor...
+std::vector<int> cycpp_shape_latitude;
```
gonuke: What is this for?

bam241: this is because `cycpp` is not run on the file generate by the `cpp`

see https://github.com/cyclus/cyclus/blob/279a758fbfb33f0bf142e0126bf269673c3364e7/cmake/UseCyclus.cmake#L122
and 
https://github.com/cyclus/cyclus/blob/279a758fbfb33f0bf142e0126bf269673c3364e7/cli/cycpp.py#L2328-L2335

so when using this snippet method we miss generated code for the additional included variables...
I try to change the `--pass3-use-orig` to `--pass3-use-pp` and `cyclus` is not compiling :(

I don't know what the best solution is...

bam241: cc @gonuke this is a follow-up on the conversation we had this morning...

gonuke:
This needs a news file

gonuke:
will merge with news file!

bam241:
shall we not wait for the CEP27 to be ready ?

gonuke:
This was paused waiting for CEP27 to be accepted.  That [did happen in March 2021](https://github.com/cyclus/cyclus.github.com/pull/297), so this should be reconsidered.

A similar approach, based on CEP27, would be very valuable for @nuclearkatie's #1634

bam241:
@gonuke just rebased this. This does not break any new things :)

gonuke:
For some reason, this update will cause cymetric to fail, but only on Ubuntu 22.04 and only those that build with apt and not conda.  This usually means that there is some dependency version conflict in the python stack that we don't have enough control over....

src/toolkit/position.cycpp.h
gonuke: @dean-krueger I think we should include this code in a cyclus/cyclus archetype like you did for the Econ stuff.  It only makes sense for a facility.

gonuke:
I think it's probably OK to go forward with this.  I don't know why it would cause any failures and maybe they were transient?  I made a suggestion to demonstrate the implementation in a Cyclus/cyclus test agent. Perhaps adding that will trigger tests and we can see if the situation improves?

katyhuff (approved):
Looks good to me!

Issue #1517, opened 2019-06-28
Add test to check new file associated to new PRs


bam241:
Adding CI-test to verify that a new news file has been created

bam241:
this has been done.
Clonsing

PR #1522, opened 2019-07-02
Add news test

bam241:
this should add a test to ensure a new news file have been created


bam241:
this should address #1517

PR #1523, opened 2019-07-02
Improve the CI speed

bam241:
this should improve CI speed by removing installation of `git` and `openssh` in the docker image

### DIFF
--- a/news/faster_ci.rst
+++ b/news/faster_ci.rst
@@ -1,0 +1,12 @@
+**Added:** None
+
+**Changed:** None
+-  git and open-ssh added to the dockerfile (removed from CI)
+
+**Deprecated:** None
+
+**Removed:** None
+-  git and open-ssh form CI (added to the dockerfile)
+**Fixed:** None
+
+**Security:** None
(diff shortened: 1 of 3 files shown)


bam241:
(this might not work until the cyclus-deps docker gets updated)

gonuke:
Does this need a rebase?  It looks good to me (since it's passing), but I'm not sure how we ensure that the docker image has git now?

bam241:
> Does this need a rebase? It looks good to me (since it's passing), but I'm not sure how we ensure that the docker image has git now?

it was already there (otherwise CI would have failed). but I added it to the apt-get in the docker file just to be sure it will work in the future....

news/faster_ci.rst
```diff
@@ -0,0 +1,12 @@
+**Added:** None
+
+**Changed:** None
+-  git and open-ssh added to the dockerfile (removed from CI)
+
+**Deprecated:** None
+
+**Removed:** None
+-  git and open-ssh form CI (added to the dockerfile)
```
gonuke:
```suggestion
-  git and open-ssh from CI (added to the dockerfile)
```

gonuke:
Thanks @bam241

PR #1527, opened 2019-08-02
House cleaning

bam241:
Just remove the mention to the different `develop` branches, as we don't use it anymore....


gonuke:
Looks like this needs a `news` file

PR #1530, opened 2019-08-09
Life time

FlanFlanagan:
This adds a new lifetime_force function that is made to force a change in lifetime for agents already in existence. 

The reason this behavior is required is that automated decommissioning can not always be done by calling the Decommission() function on an agent (if the Decommission_Condition returns false). This allows the user to change the lifetime of the agent, and hopefully force it to begin it's decommission process on it's own.

### DIFF
--- a/cyclus/agents.pyx
+++ b/cyclus/agents.pyx
@@ -586,7 +586,7 @@ cdef cppclass CyclusFacilityShim "CyclusFacilityShim" (cpp_cyclus.Facility):
         (<object> this.self).decision()
 
     cpp_bool CheckDecommissionCondition() except +:
-        rtn = (<object> this.self).check_decomission_condition()
+        rtn = (<object> this.self).check_decommission_condition()
         return bool_to_cpp(rtn)
 
     std_set[shared_ptr[cpp_cyclus.RequestPortfolio[cpp_cyclus.Material]]] GetMatlRequests() except +:
@@ -1101,7 +1101,7 @@ cdef class _Region(_Agent):
         rtn = bool_to_py((<CyclusRegionShim*> (<_Agent> self).shim).DecendentOf(cpp_other))
         return rtn
 
-    def decomission(self):
+    def Decommission(self):
         """Decommissions the agent, removing it from the simulation. Results in
         destruction of the agent object. If agents write their own decommission()
         function, they must call their superclass' decommission function at the
@@ -1199,7 +1199,7 @@ cdef class _Region(_Agent):
         """The default time step at which this agent will exit the
         simulation (-1 if the agent has an infinite lifetime).
 
-        Decomissioning happens at the end of a time step. With a lifetime of 1, we
+        decommissioning happens at the end of a time step. With a lifetime of 1, we
         expect an agent to go through only 1 entire time step. In this case, the
         agent should be decommissioned on the same time step it was
         created. Therefore, for agents with non-infinite lifetimes, the exit_time
@@ -1301,7 +1301,7 @@ cdef class _Institution(_Agent):
         rtn = bool_to_py((<CyclusInstitutionShim*> (<_Agent> self).shim).DecendentOf(cpp_other))
         return rtn
 
-    def decomission(self):
+    def Decommission(self):
         """Decommissions the agent, removing it from the simulation. Results in
         destruction of the agent object. If agents write their own decommission()
         function, they must call their superclass' decommission function at the
@@ -1399,7 +1399,7 @@ cdef class _Institution(_Agent):
         """The default time step at which this agent will exit the
         simulation (-1 if the agent has an infinite lifetime).
 
-        Decomissioning happens at the end of a time step. With a lifetime of 1, we
+        decommissioning happens at the end of a time step. With a lifetime of 1, we
         expect an agent to go through only 1 entire time step. In this case, the
         agent should be decommissioned on the same time step it was
         created. Therefore, for agents with non-infinite lifetimes, the exit_time
@@ -1502,13 +1502,15 @@ cdef class _Facility(_Agent):
         rtn = bool_to_py((<CyclusFacilityShim*> (<_Agent> self).shim).DecendentOf(cpp_other))
         return rtn
 
-    def decomission(self):
+    def Decommission(self):
         """Decommissions the agent, removing it from the simulation. Results in
         destruction of the agent object. If agents write their own decommission()
         function, they must call their superclass' decommission function at the
         END of their decommission() function.
         """
+        print('decomm2py')
         (<CyclusFacilityShim*> (<_Agent> self).shim).Decommission()
+        print('enddecomm2py')
 
     @property
     def annotations(self):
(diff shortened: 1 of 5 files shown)


FlanFlanagan:
WIP. 

Need to push some errors from c++ to python still

FlanFlanagan:
As an update. This works to 'retire' reactor facilities, but it does not cause them to decommission. It does cause other facilities to decommission though.

src/agent.h
```diff
@@ -382,12 +382,17 @@ class Agent : public StateWrangler, virtual public Ider {
...
 
-  /// Sets the number of time steps this agent operates between building and
+  ///Sets the number of time steps this agent operates between building and
   /// decommissioning (-1 if the agent has an infinite lifetime).  This should
   /// generally only be called BEFORE an agent is added to a context as a
   /// prototype.  Throws ValueError if the agent has already been deployed.
   void lifetime(int n_timesteps);
 
+  /// Sets the number of time steps this agent operates between building and
+  /// decommissioning (-1 if the agent has an infinite lifetime).  The boolean
```
gonuke: what "boolean" are you referring to in the comment?

src/agent.h
```diff
@@ -382,12 +382,17 @@ class Agent : public StateWrangler, virtual public Ider {
...
+  /// allows for an agent to override the ValueError.
+  void lifetime_force(int n_timesteps);
```
gonuke: Since this is likely to happen after the agent has been deployed and operated for some number of timesteps, this argument is the sum of already (completed time steps) + (remaining time steps).  What will happen if `n_timesteps` < (completed time steps)?  Will that trigger decommissioning?  Or will we have missed the chance for that?

src/agent.h
FlanFlanagan: Ahh, thanks. I was originally using a boolean but couldn't get it to work with the python bindings. I'll update that.

src/agent.h
FlanFlanagan: Yeah, it is possible that it will miss the chance to decommission. I guess I can build in a catch to ensure that it can't be set for historic time steps.

cyclus/agents.pyx
```diff
@@ -1101,7 +1101,7 @@ cdef class _Region(_Agent):
         rtn = bool_to_py((<CyclusRegionShim*> (<_Agent> self).shim).DecendentOf(cpp_other))
         return rtn
 
-    def decomission(self):
+    def Decommission(self):
```
gonuke: This is an odd case change.... can’t really tell if it’s correct or an accident?  Did this not work before, at all?

src/agent.cc
```diff
@@ -112,6 +112,20 @@ void Agent::lifetime(int n_timesteps) {
   lifetime_ = n_timesteps;
 }
 
+void Agent::lifetime_force(int n_timesteps) {
+  try{
+    lifetime(n_timesteps);
+  }
+  catch (ValueError e){
+    if(enter_time_+n_timesteps <= context()->time()){
+      throw ValueError("cannot set lifetime to be less than current context time");
```
gonuke: What if, instead, we set the lifetime to decommission ASAP?  That may have some unintended consequences....,  but might better match expectations? 🤷‍♂️

cyclus/agents.pyx
FlanFlanagan: It was on purpose. It was to fix the spelling.

src/agent.cc
FlanFlanagan: Yeah, I can do that. Thanks!

cyclus/agents.pyx
gonuke: I know the spelling was fixed, but I’m not clear on what the style guide says about the capitalization

FlanFlanagan: Ahh. Yes! That was a mistake. I was lazy and did a simple find replace. I'll go through and fix these up. Thanks!

cyclus/cpp_cyclus.pxd
```diff
@@ -655,7 +655,7 @@ cdef extern from "agent.h" namespace "cyclus":
         void EnterNotify()
         void BuildNotify(Agent*)
         void DecomNotify(Agent*)
-        void Decommission()
+        void Decommission() except +
```
gonuke: my cython isn't good enough to know what the correct capitalization is after all the back-and-forth?

cyclus/lib.pyx
```diff
@@ -1381,7 +1381,7 @@ cdef class _Agent:
         rtn = bool_to_py((<cpp_cyclus.Agent*> self.ptx).DecendentOf(cpp_other))
         return rtn
 
-    def decomission(self):
+    def Decommission(self):
```
gonuke: capitalization?

src/agent.cc
```diff
@@ -112,6 +112,20 @@ void Agent::lifetime(int n_timesteps) {
   lifetime_ = n_timesteps;
 }
 
+void Agent::lifetime_force(int n_timesteps) {
+  try{
+    lifetime(n_timesteps);
+  }
+  catch (ValueError e){
+    if(enter_time_+n_timesteps <= context()->time()){
+      lifetime(context()->time()+1);
```
gonuke:
```suggestion
      lifetime(context()->time() - enter_time + 1);
```

cyclus/cpp_cyclus.pxd
FlanFlanagan: My understanding is that the pxd file is what makes the c++ functions available to python. So in the pyx file it looks like this
decommission()
         (<somestuff>).Decommission()

The decommission() function is called in python and Decommission() is called through that via cython.

gonuke:
Thanks @FlanFlanagan - this LGTM - I wonder if it's worth a community discussion (probably not a CEP - maybe an amendment to a CEP) to make sure everyone's aware of the potential change in behavior?

FlanFlanagan:
I think it does. I was planning on bringing it up at the last dev meeting. I will bring it up next meeting.

src/agent.cc
```diff
@@ -112,6 +112,20 @@ void Agent::lifetime(int n_timesteps) {
...
+void Agent::lifetime_force(int n_timesteps) {
+  try{
+    lifetime(n_timesteps);
+  }
+  catch (ValueError e){
+    if(enter_time_+n_timesteps <= context()->time()){
+      lifetime(context()->time() - enter_time_ + 1);
+    }
+    else{
+      lifetime_ = n_timesteps;
```
bam241: As the only check in the `lifetime()` method is if the number of timestep (n_timesteps) is negative...
if `lifetime()` throws an error your implementation will actually overwrite the check and still set the lifetime to a negative's value.
If I understand correctly, is this intended ?

bam241:
beside my one question/comment, it looks good to me...

src/agent.cc
gonuke: No - `lifetime()` checks the value of `enter_time` not `n_timesteps`....

bam241: True !

I think I understand now. facility that have not being deployed have a `enter_time_ = -1` but once deployed it is set to the actual `enter_time`.
so if `lifetime()` throw an error that means the facility has been deployed
if the facility has been deployed, 2cases:
- the new lifetime ends before the actual time (so we set it to the end at the next timestep)
- the new lifetime ends after the actual time, we just update the lifetime...

bam241:
thank you @FlanFlanagan !
Merging!

bam241:
oups I forgot about the cep :(

bam241:
@FlanFlanagan I definitely was to quick to merge, could you PR a news file for this ?

PR #1533, opened 2019-09-09
fix news test trigger...

bam241:
first fix did not work, still running the news test after a merge...
is this one better ?


bam241:
it worked never mind....

I saw the "Merge branch 'master' into position" that was in opened PR....

PR #1537, opened 2019-12-20
Fix Conda install on MAC OS

bam241:
This prevent building `libcyclus` and `cyclus` binary against PYTHONLIB on MACOS.

I believe MACOS does not do exactly the same thing as Linux based system when building static/charged libs.

Without this, when importing cyclus python packages the system ends up importing twice the PYTHONLIB (ones because it is in python environment one because it is included in libcyclus, on which cyclus python package relies on...) — strangely it only occurs when using Conda python...

I tried this on macOS (no Conda) and Ubuntu (no Conda) 
The CI will check Ubuntu (Conda) and @yarden-livnat has tested in on MACOS(Conda)


CMakeLists.txt
```diff
@@ -337,7 +337,10 @@ IF(NOT CYCLUS_DOC_ONLY)
 
         # make sure we know about having python
         add_definitions(-DCYCLUS_WITH_PYTHON)
-        set(LIBS ${LIBS} ${PYTHON_LIBRARIES})
+        IF (NOT APPLE)
+            MESSAGE("ININININHBJGVHGCVGFCGFCGFCGFCGFCGFCGFCGFC:")
```
gonuke: Is this necessary?? or something leftover from debugging?

bam241: pushy sorry, forgot to clean this up....

bam241:
> I sure wish I understood this better, but will trust your testing experience.
> 
> Needs a news file!

I think it is related to how MacOS forms the Static lib if (I understand correctly part of the answer can be found there: https://stackoverflow.com/questions/2339679/what-are-the-differences-between-so-and-dylib-on-osx)
basically it build static lib as bundles and if you link against an other static lib it cannot be loaded as a shared lib but only as a "bundle".
This might results in loading twice the PYTHONLIB.
For a reason, I ignore (they might be built differently), when using the `macport` PYTHONLIB you don't get that issue.

Also on MACOS (regardless on the` PYTONLIB` provenance), importing `PYTHONLIB` is only required when building the binary.
Technically the `libcyclus` (C++) would need it too, but as it is only loaded thought the Python package or the binary, MACOS gets the `Python` symbols either from the Python Environment or the binary that is linked against it.
This does not work on Ubuntu, libcyclus needs to be linked against `PYTHONLIB`.

bam241:
@gonuke just added the news file

CMakeLists.txt
```diff
@@ -337,7 +337,9 @@ IF(NOT CYCLUS_DOC_ONLY)
 
         # make sure we know about having python
         add_definitions(-DCYCLUS_WITH_PYTHON)
-        set(LIBS ${LIBS} ${PYTHON_LIBRARIES})
+        IF (NOT APPLE)
+            set(LIBS ${LIBS} ${PYTHON_LIBRARIES})
```
gonuke: I guess another question - do we need to add `PYTHON_LIBRARIES` for linux.  I guess it's clear we don't need it for MacOS, but could we do without it for Linux?

gonuke: That is, linux may tolerate it, but it still may not be necessary??? (pure speculation...)

bam241: I'll try this

CMakeLists.txt
```diff
@@ -337,7 +337,9 @@ IF(NOT CYCLUS_DOC_ONLY)
...
+        ENDIF (!APPLE)
```
nuclearkatie: Is there a reason you open the if statement with ```NOT APPLE``` but end it with ```!APPLE``` ? Does it matter if they exactly match?

CMakeLists.txt
bam241: this is required for `linux`

CMakeLists.txt
bam241: fair point fixing it

cli/CMakeLists.txt
```diff
@@ -8,10 +8,14 @@ INCLUDE_DIRECTORIES(BEFORE ${CYCLUS_CORE_INCLUDE_DIRS})
...
     cyclus.cc
     )
+IF (APPLE)
+    SET(LIBS ${PYTHON_LIBRARIES} ${LIBS})
+ENDIF(APPLE)
 
 ADD_EXECUTABLE(cyclus_cli ${CYCLUS_SOURCE})
 set_target_properties(cyclus_cli PROPERTIES OUTPUT_NAME cyclus)
 
+
```
stompsjo:
```suggestion
```

I don't think this extra line is needed

bam241:
@gonuke we need you to approve the PR to be able to merge.

gonuke:
Thanks @bam241

PR #1544, opened 2020-04-27
allowing 0 capacity constrain

bam241:
Allows 0 capacity constrain to be set, this allows developer to set a capacity constrain based on a variable without having to check that the constrain they are setting is `>0` , also this might be hard to debug as it will only show up at run time. 

It is very hard for a developer to own it need to design test for it... So I think that instead of trowing an error at runtime we should allow it, as 0 capacity bid should not impact the DRE solver.

what do you think @gonuke ?

### DIFF
--- a/src/capacity_constraint.h
+++ b/src/capacity_constraint.h
@@ -68,7 +68,7 @@ class CapacityConstraint {
       : capacity_(capacity),
         converter_(converter),
         id_(next_id_++) {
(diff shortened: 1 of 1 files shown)


gonuke:
This seems like something that could have deep but subtle implications to the functioning of the DRE!  I wonder if it is worth considering what this means in the context of the formal definition of the DRE and its constraints.

PR #1551, opened 2021-02-05
Fix redeclaration error

maxschalz:
I ran into the following issue when compiling Cyclus with GCC and G++ v10.2.1:
```
/home/maxschalz/NVD/cyclus/src/hdf5_back.cc: In member function ‘void cyclus::Hdf5Back::CreateTable(cyclus::Datum*)’:
/home/maxschalz/NVD/cyclus/src/hdf5_back.cc:7672:14: error: redeclaration of ‘template<class _Tp, class _Alloc> class std::__cxx11::list’
 7672 |   using std::list;
      |              ^~~~
```
When compiling with clang, I did not encounter this problem but the compilation failed for other reasons, so deleting the declaration in `src/hdf5back.cc.in` solved the issue and allowed me to stick with g++.

### DIFF
--- a/news/redeclaration_error.rst
+++ b/news/redeclaration_error.rst
@@ -1,0 +1,12 @@
+**Added:** None
+
+**Changed:** None
+
+**Deprecated:** None
+
+**Removed:** None
+
+**Fixed:**
+- Redeclaration error in ``src/hdf5_back.cc`` during compilation with G++
+
+**Security:** None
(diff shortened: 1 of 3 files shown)


gonuke:
Thanks for this update @maxschalz.  

Please also add a news update:
- make a copy of news/TEMPLATE.rst in the news directory
- modify that file with a brief description of the change (1 line) in the appropriate section
- add that file to your PR

maxschalz:
Sorry @gonuke , I was not aware of that, have added an update now.

gonuke:
Thanks @maxschalz - I don't think we have documented it well...

PR #1552, opened 2021-02-25
few change in sell_policy

bam241:
this should allow quantize to be equal to 0, which corresponds to no quantize.
(the call to `Excl()` before each call to quantize makes sure that quantize is used only if `> 0`)

this also fix a problem that were happening when the buyer capacity was in between 0 and the quantize. it would still place bid without the capacity to fulfill them.

cc @nuclearkatie


gonuke:
Is this being tested anywhere?  I'm guessing not, because it didn't fail the test... So maybe we should add a test?

bam241:
we planned to do that in peer programming with @nuclearkatie

bam241:
@gonuke , @nuclearkatie and I just added a test to check:
- quantity < quanta -> no offer are sent
-  quanta < quantity -> an integer number of quanta is offered

tests/toolkit/matl_sell_policy_tests.cc
```diff
@@ -100,9 +100,23 @@ TEST_F(MatlSellPolicyTests, Bids) {
...
+                  qty / 2);
+  ASSERT_EQ((*(*obs.begin())->bids().begin())->offer()->comp(), comp1);
+
+  // If available quantity is bigger than the quanta only an integer number of quanta is offered
+  p.Init(NULL, &buff, "", qty, true, 2).Set(commod);
+  obs = p.GetMatlBids(reqs);
+  ASSERT_EQ(obs.size(), 1);
+  ASSERT_EQ((*obs.begin())->bids().size(), 1);
+  ASSERT_FLOAT_EQ((*(*obs.begin())->bids().begin())->offer()->quantity(),
+                  2);
```
gonuke, line 112: I'm not sure I understand why this test behaves differently from the previous one???  The initializer for `p` is the same except for the value of the final entry - why only a single bid in this case?

tests/toolkit/matl_sell_policy_tests.cc
```diff
@@ -100,9 +100,23 @@ TEST_F(MatlSellPolicyTests, Bids) {
   ASSERT_EQ(obs.size(), 1);
   ASSERT_EQ((*obs.begin())->bids().size(), 2);
   ASSERT_FLOAT_EQ((*(*obs.begin())->bids().begin())->offer()->quantity(),
-                  mat->quantity() / 2);
+                  qty / 2);
+  ASSERT_EQ((*(*obs.begin())->bids().begin())->offer()->comp(), comp1);
+
+  // If available quantity is bigger than the quanta only an integer number of quanta is offered
+  p.Init(NULL, &buff, "", qty, true, 2).Set(commod);
```
gonuke, line 107: Should we use a variable for the final argument, and then reuse that variable in line 112?  What is important about this variable vs the previous test?  That it's an integer? That it's smaller than `qty` does it matter how much smaller?

gonuke (changes requested):
Some questions about the tests, only because I don't fully understand how this additional test is different from the previous one.

tests/toolkit/matl_sell_policy_tests.cc
bam241: maybe one need to clarify the initial condition for this (all?) problem(s)

tests/toolkit/matl_sell_policy_tests.cc
bam241: there is a single bid because the available quantity is 3, the throughput is 3 but the quanta is 2, so only 1 single shipment can fulfill the quanta restriction

bam241:
@gonuek I added comments to clarify the test cases.

bam241:
@gonuke I added comment to clarify the test... (this time pinging without typo...)

gonuke (approved):
Thanks for the detailed work here @nuclearkatie & @bam241

PR #1557, opened 2021-03-15
Cbc 2.10

bam241:
This should enable compatibility with `cbc 2.10` 

I tested that on a docker ubuntu 20.04 it compiles and unit_tests are running fine.

maybe @nuclearkatie could confirm if that's is working for her...


nuclearkatie:
I was able to successfully install and run unit tests on a Ubuntu 20.04 VM using this. Thanks @bam241

src/OsiCbcSolverInterface.cpp
```diff
@@ -7,6 +7,8 @@
 #include "OsiCbcSolverInterface_2_8.cpp"
 #elif CBC_VERSION_MAJOR == 2 && CBC_VERSION_MINOR == 9
 #include "OsiCbcSolverInterface_2_9.cpp"
+#elif CBC_VERSION_MAJOR == 2 && CBC_VERSION_MINOR == 10
```
gonuke, line 10: I know this pattern was inherited from before, but why not just include the `hpp` file here, instead?

gonuke (changes requested):
Thanks for this quick fix @bam241  - just one comment

gonuke:
Do we want to update CI to test with 2.10? Or both?

bam241:
this is linked to #1262

src/OsiCbcSolverInterface.cpp
```diff
@@ -7,6 +7,8 @@
 #include "OsiCbcSolverInterface_2_8.cpp"
```
gonuke, line 7: I think we decided to include the `hpp` file here instead of duplicating this block

src/OsiCbcSolverInterface.hpp
```diff
@@ -13,6 +13,8 @@
 #include "OsiCbcSolverInterface_2_8.hpp"
 #elif CBC_VERSION_MAJOR == 2 && CBC_VERSION_MINOR == 9
 #include "OsiCbcSolverInterface_2_9.hpp"
+#elif CBC_VERSION_MAJOR == 2 && CBC_VERSION_MINOR == 10
+#include "OsiCbcSolverInterface_2_10.hpp"
```
gonuke, line 17: I think we decided:
```suggestion
#elif CBC_VERSION_MAJOR == 2 && CBC_VERSION_MINOR <= 10
#include "OsiCbcSolverInterface_2_10.hpp"
```

gonuke: We can probably also delete the v2.9 hpp file.

gonuke:
I think there is still a question about how we test this.  If we had access to the right versions, we could test at merge, but even that may be a lot of effort.  I don't think we want/need to test in CI for every PR.

src/OsiCbcSolverInterface.cpp
```diff
@@ -7,6 +7,8 @@
...
 #elif CBC_VERSION_MAJOR == 2 && CBC_VERSION_MINOR == 9
 #include "OsiCbcSolverInterface_2_9.cpp"
+#elif CBC_VERSION_MAJOR == 2 && CBC_VERSION_MINOR == 10
+#include "OsiCbcSolverInterface_2_10.hpp"
```
bam241, line 11: this should be `.cpp`

bam241:
2.9 and 2.10 should be merged together as they are identical.

bam241:
@gonuke There is something I don't understand.

the cmake file adds automatically all the *.cc files and manually `OsiCbcSolverInterface.cpp`
with this news implementation we only include `OsiCbcSolverInterface.hpp` from `OsiCbcSolverInterface.cpp`.
 `OsiCbcSolverInterface.hpp` conditionally include the proper  `OsiCbcSolverInterface_x_Y.hpp` header.
but how are the symbol added to the library as we don't add the `OsiCbcSolverInterface_x_Y.cpp` to the built library.

(I believe the include that existed before served that purpose...)

gonuke:
I think you are right.  I probably misread the includes and assumed they were HPP files, but they are in fact CPP files.

gonuke (approved):

src/OsiCbcSolverInterface.cpp
```diff
@@ -1,15 +1,16 @@
...
+// // CBC_VERSION_MAJOR defined for Cbc > 2.5
+// #ifndef CBC_VERSION_MAJOR
+// #include "OsiCbcSolverInterface_2_5.hpp"
+// #elif CBC_VERSION_MAJOR == 2 && CBC_VERSION_MINOR <= 8
+// #include "OsiCbcSolverInterface_2_8.hpp"
+// #elif CBC_VERSION_MAJOR == 2 && ( CBC_VERSION_MINOR == 9 || CBC_VERSION_MINOR == 10)
+// #include "OsiCbcSolverInterface_2_9_10.hpp"
+// #else
+// #error "Cyclus cannot yet handle your version of CoinCBC. Please open an issue with your CoinCBC version."
+// #endif
```
gonuke, line 13: I think we need to revert this change, right? and make sure to include the `*.cpp` files here.

gonuke (changes requested):
Be sure to include cpp files where they are needed

gonuke (approved):

gonuke:
this looks good - I'll wait for tests to finish

gonuke:
Why are all our material tests failing?

bam241:
root cause might come from composition.

I need to investigate....

gonuke:
We definitely changed PyNE in this space - but have we updated PyNE in CI or in our build since those changes?
Should we be submoduling PyNE?

bam241:
this is very strange, I can't reproduce that in docker using the same image we run on CI...

bam241:
can this be a ressource issue ?

bam241:
this needs #1567 to pass

gonuke:
rebase this for tests to work!?!?

gonuke:
Thanks @bam241

PR #1558, opened 2021-03-15
use update-alternatives even with single python version

nuclearkatie:
I was running into build errors on a fresh Ubuntu 20.04 VM. The fix, helpfully provided by @bam241, was using `update-alternatives` to point `python` at the right version. `DEPENDENCIES.rst` currently makes it seem like the user only needed to use `update-alternatives` if they had both python 2.7 and python3 (3.5) installed on their machine, this PR adds a few lines noting that it may be necessary even with a single version of python installed.

### DIFF
--- a/DEPENDENCIES.rst
+++ b/DEPENDENCIES.rst
@@ -187,6 +187,20 @@ From now on, to switch between different versions, use:
 
    sudo update-alternatives --config python
 
+Even if you only have a single python installation, you may have to use update-alternatives to point python at the correct version to avoid build errors.
+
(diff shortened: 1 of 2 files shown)


bam241:
thx for the doc update.

I am unsure exactly why this was happening, but it seems that at least one of our scripts require a `python` executable to exist.

the `update-alternative` allows this to exist

gonuke:
Thanks @nuclearkatie

gonuke (approved):
LGTM

Issue #1562, opened 2021-06-23
cyclus-build-deps in need of update


dermesser:
Also relates to #1497.

I have some hard-earned knowledge on building cyclus within an anaconda environment. While initially being led astray by `cyclus-build-deps`, and the outdated INSTALL and DEPENDENCIES files, I have managed to assemble an anaconda environment that enables building cyclus in a reliable manner.

As Anaconda YAML file, the following works (for me), for cyclus 1.5.5; please let me know if you'd be interested in adding such information to the existing install documentation. Especially fickle details, like which boost package to use, linking errors among glib/xml++/..., are time-consuming to figure out.

```yaml
channels:
  - conda-forge
  - defaults
dependencies:
  - _libgcc_mutex=0.1=main
  - _openmp_mutex=4.5=1_gnu
  - _sysroot_linux-64_curr_repodata_hack=3=haa98f57_10
  - binutils_impl_linux-64=2.35.1=h27ae35d_9
  - binutils_linux-64=2.35.1=h454624a_30
  - blas=1.1=openblas
  - boost-cpp=1.70.0=h7b93d67_3
  - bzip2=1.0.8=h7f98852_4
  - ca-certificates=2021.5.25=h06a4308_1
  - certifi=2021.5.30=py36h06a4308_0
  - cmake=3.19.6=h973ab73_0
  - coincbc=2.9.9=blas_openblash0f40ac4_1002
  - cython=0.29.23=py36h2531618_0
  - expat=2.4.1=h2531618_2
  - gcc_impl_linux-64=9.3.0=h6df7d76_17
  - gcc_linux-64=9.3.0=h1ee779e_30
  - gettext=0.19.8.1=h0b5b191_1005
  - glib=2.68.3=h9c3ff4c_0
  - glib-tools=2.68.3=h9c3ff4c_0
  - glibmm=2.58.1=h778ead0_1
  - gxx_impl_linux-64=9.3.0=hbdd7822_17
  - gxx_linux-64=9.3.0=h7e70986_30
  - hdf5=1.10.6=hb1b8bf9_0
  - icu=67.1=he1b5a44_0
  - jinja2=3.0.0=pyhd3eb1b0_0
  - kernel-headers_linux-64=3.10.0=h57e8cba_10
  - krb5=1.18.2=h173b8e3_0
  - ld_impl_linux-64=2.35.1=h7274673_9
  - libblas=3.8.0=17_openblas
  - libcblas=3.8.0=17_openblas
[...]

PR #1563, opened 2021-06-29
Add notices about installation pitfalls

dermesser:
For #1562. tl;dr: Don't send users down a rabbit hole by implying that you just need the right dependencies, and instead point to reliable ways like binary/virtualbox/docker install.

I added a promise that `cyclus-build-deps` will at some point be updated again. Let me know if this is inaccurate! In general, I'm happy about suggestions here, as I'm not fully familiar with your preferred way of writing docs.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -12,7 +12,10 @@ provided with the source.
 Dependencies
 ************
 
-A full list of the Cyclus package dependencies is shown below:
+A full list of the Cyclus package dependencies is shown below; note however
+that many of these versions are not available anymore from common sources
+(anaconda, Linux distributions, etc.), and Cyclus doesn't always work with any
+newer version.
 
 ====================   ==================
 Package                Minimum Version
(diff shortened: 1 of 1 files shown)


gonuke:
Thanks for these great notes - our first goal is to try to fix some of these problems and then see where the documentation should stand for identifying what's left.

abachma2:
We've made quite a few updates to the build and dependency versions. Do we still need some of these documentation updates? @bennibbelink?

gonuke:
Yes -I think we do need documentation updates, but these particular ones may no longer be correct.

I think once we have a PR with relevant updates, we can ask for this PR to be withdrawn, with reference to that new PR.

gonuke:
Probably need to update all the places that dependencies are defined and identify a minimum version so that users know the minimum version that we choose to support.

abachma2:
We've made quite a few updates to the build and the documentation for the build. I don't think the conda packages are updated yet (see #1686 for progress on that). Do the new build and documentation address the issues addressed by this PR @dermesser? Do we need further changes/updates to this PR to address those updates?

abachma2:
Hi @dermesser Have you had any time to review the latest updates to Cyclus and the build? Do they address the concerns you address through this PR?

gonuke:
We've made widespread changes to our dependencies and their documentation.  We hope it addresses your problems, but if you experience continued challenges, please let us know @dermesser.

PR #1568, opened 2021-07-30
CircleCI 2 GithubAction

bam241:
attempt to convert our CircleCI CI to GithubAction


bam241:
reopening from cyclus org (first workflow)

PR #1569, opened 2021-07-30
Circle 2 GitHubAction

bam241:
attempt to convert our Circle CI to GithubAction

PR #1570, opened 2021-08-04
adding workflow to build docker img

bam241:


gonuke:
Why does the nosetest fail here when testing the image but not in the normal build & test that is basically ready to merge (#1569)?

bam241:
not sure quite yet, might be a permission issue or a path issue....

it seems that githubAction `runs` inside /github/home/... which is not the default location, I am not sure about the `build` part, some path might change, or some permission have to be fixed, or some ENV variable are missing....
probably one (or multiple) of the three...

gonuke:
This is being superseded by #1588

gonuke:
There is a useful update to the pattern of handling input files in the tests to help them be runnable from different locations.  Perhaps that can be lifted and abstracted into a few places in the testing.

bam241:
this is super outdate.
closing.

PR #1571, opened 2021-08-04
adding housekeeping scripts

bam241:
Updating news check to use GithubAction

switching to unified  CHANGELOG entries


CHANGELOG.rst
```diff
@@ -2,7 +2,39 @@
 cyclus Change Log
 =================
 
-.. current developments
+Since last release
+====================
+
+**Added:**
+
+* adding compatibility with cbc 2.10
```
gonuke, line 10: Should we put PR numbers on each entry?

gonuke (changes requested):
This is supposed to fail the news test on CircleCI, right?  Should we remove that test as part of this PR?

CHANGELOG.rst
bam241: we did not in the past.
I can do it for the coming release, and set it as a expectation...

bam241: I actually thought it is better to have the PR numbers, so I added them.

bam241:
I also removed housekeeping from CircleCI

gonuke (approved):
Thanks @bam241

PR #1587, opened 2023-09-08
2023 maintenance update

gonuke:
Update build for success in modern environments:
* Ubuntu 22.04
* both conda (libxml++4.0) and apt (libxml++2.6)
* implement github actions for building CI containers as well as biuld/test

Issue #1585, opened 2023-07-19
Cyclus install fail on Ubunutu VM in windows


DanJack1987:
CMake Error at cmake/UseCyclus.cmake:136 (message):
  cycpp failed on '/home/daniel/cyclus/tests/test_agents/test_agent.cc' with
  exit code '1'
Call Stack (most recent call first):
  cmake/UseCyclus.cmake:245 (USE_CYCLUS)
  tests/test_agents/CMakeLists.txt:3 (install_cyclus_standalone)

DanJack1987:
[CMakeError.log](https://github.com/cyclus/cyclus/files/12089996/CMakeError.log)
[CMakeOutput.log](https://github.com/cyclus/cyclus/files/12089997/CMakeOutput.log)

gonuke:
Thanks for the feedback @DanJack1987 .  We are working to resolve and modernize the build issues in #1588 and will check back with you when it's ready.

gonuke:
Per discussion in #1700, we will no longer povide VM images since we provide a number of other avenues for installation

Issue #1580, opened 2022-05-17
Update CMake version compatibility


gonuke:
Update to the newest version available on Ubuntu 20.04: CMake 3.16, and resolve other CMake warnings.

gonuke:
This was resolved by #1588

Issue #1572, opened 2021-08-13
Switch to GithubAction for CI


bam241:
Here is all the capabilities we have in CircleCI that would (?) need to be transferred to GithubAction:
- [x] changelog/news checker (#1571)
- [x] build / nosetest / unit test (#1569)
- [x] build cyclus/cyclus:latest docker image (on merge) (#1570 WIP)
- [ ] build cyclus/cyclis:stable docker image (on release) (#1570 WIP) 
- [ ] generate `.deb` package (on release) 
- [ ] trigger cycles-suite test (to ensure `cycamore` and `cymetric` are not broken by the latest changes)

I'll edit this list with the corresponding PR number

gonuke:
A github action for release was added in #1597. It doesn't do all of these things yet, but we can discuss how they fit in our new strategy.

gonuke:
Let's close this in favor of the [project](https://github.com/orgs/cyclus/projects/4/views/1)

PR #1588, opened 2023-09-13
2023 maintenance update

gonuke:
Update build for success in modern environments:

- Ubuntu 22.04
- both conda (libxml++4.0) and apt (libxml++2.6)
- implement github actions for building CI containers as well as build/test

This is being launched from a branch on the upstream repo so that the docker image push behavior will work.  This should be the standard practice for PR's that require docker image pushes.  This replaces #1587 

Hopefully this helps address the following: #1547, #1549, #1556, #1562, #1572, #1580, #1585


tests/test_cycluslib.py
```diff
@@ -3,20 +3,17 @@
...
-import nose
-from nose.tools import assert_equal, assert_less
-
 from cyclus import lib
 
 from tools import libcyclus_setup, dbtest
 
 
-setup = libcyclus_setup
+#setup = libcyclus_setup
```
abachma2, line 11: Can we remove this line?

tests/test_null_sink.py
```diff
@@ -68,11 +67,11 @@ def check_null_sink(fname, given_spec):
...
     sink_id = find_ids(given_spec, spec, agent_ids)
     # Test if one SimpleSink is deployed
-    yield assert_equal, len(sink_id), 1
+    assert len(sink_id) == 1
 
     # No resource exchange is expected
-    yield assert_false, tables_exist(outfile, illegal_paths)
-
+    assert not tables_exist(outfile, illegal_paths)
+    assert 1 == 2 # FIX ME
```
abachma2, line 74: I think this still needs to be fixed.

tests/toolkit/infile_converters_tests.cc
```diff
@@ -86,11 +86,12 @@ TEST(InfileConverters, JsonPyRoundTrip) {
   cyclus::PyStart();
   string inp = cyclus::toolkit::XmlToJson(MakeInput());
   string p1 = cyclus::toolkit::JsonToPy(inp);
-  string j1 = cyclus::toolkit::PyToJson(p1);
+  string j1 = cyclus::toolkit::PyToJson(p1) + "\n";
```
abachma2, line 89: Is this all it takes to get this test working? Why?

tests/tools.py
```diff
@@ -10,18 +10,16 @@
...
 
 from cyclus import lib as libcyclus
 
 if sys.version_info[0] >= 3:
     basestring = str
 
-unit = attr('unit')
-integration = attr('integration')
+#unit = attr('unit')
+#integration = attr('integration')
```
abachma2, line 22: If these are being commented out, can we remove them?

xCmake/FindGlib.cmake
abachma2, line 1: Any reason for moving these cmake files?

.github/workflows/build_test.yml
```diff
@@ -1,50 +1,48 @@
...
+          ]
+        pkg_mgr : [
+          apt,
+          conda,
+        ]
 
-    container:
-      image: cyclus/cyclus-deps
-    
+    name: Installing Dependencies, Building cyclus and running tests
```
abachma2, line 26: Maybe I missed it, but it looks like this workflow just builds cyclus, but does it run any tests? The file name implies that t does.

docker/Dockerfile
```diff
@@ -0,0 +1,128 @@
...
+    wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda.sh && \
+    /bin/bash ~/miniconda.sh -b -p /opt/conda && \
+    rm ~/miniconda.sh
+
+ENV PATH /opt/conda/bin:$PATH
+
+RUN conda config --add channels conda-forge
+RUN conda update -n base -c defaults conda
+RUN conda install -y mamba
+RUN conda update -y --all && \
```
abachma2, line 64: So everything is being installed via apt and conda? why?

abachma2 (commented):
Hi @gonuke. Thanks for getting this PR in to update various parts of the code base. I have a  lot of mostly small comments, or comments to just understand what is going on in some of the parts. I would definitely recommend having someone else look over some of the parts, as I am not as well versed in some of the more nuanced things in CMake.

xCmake/FindGlib.cmake
gonuke: I think I just wanted them to be hidden from the CMake process and see if it worked without them - I'll check

CMakeLists.txt
```diff
@@ -162,24 +160,29 @@ IF(NOT CYCLUS_DOC_ONLY)
...
-    SET(LIBS ${LIBS} ${LibXML++_LIBRARIES})
-    message("-- LibXML++ Include Dir: ${LibXML++_INCLUDE_DIR}")
-
+    message("-- LibXML2 Include Dir: ${LIBXML2_INCLUDE_DIR}")
+
+    # Then use pkg-config for locate specific package
+    pkg_check_modules(LIBXMLXX IMPORTED_TARGET libxml++-4.0)
+    IF ( NOT LIBXMLXX_LIBRARIES )
+       pkg_check_modules(LIBXMLXX REQUIRED IMPORTED_TARGET libxml++-2.6)
+    ENDIF ( NOT LIBXMLXX_LIBRARIES )
```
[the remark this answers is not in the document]
gonuke: This is because Ubuntu packages only support up to v2.6 right now.  If we want that to be a path available to users, we need to support one or the other.

docker/Dockerfile
gonuke: Using the (standard) `buildx` environment, it only builds the stages that are necessary, based on a dependency check. The `cyclus` stage is based on a stage named `${pkg_mgr}-deps`.  In the case when `pkg_mgr == apt` it will only build the `apt-deps` stage and when `pkg_mgr == conda`` it will only build the `conda-deps` stag.

tests/toolkit/infile_converters_tests.cc
gonuke: It took a little more, but not much. There was an inconsistency between two ways of wrapping lines for JSON files.  One way always put a single entry on each line and other way would put a comma separated lists of entries if they fit on a line.  I was able to change the latter by saying lines had to wrap if they were longer than length 1.  There was still a discrepancy of the new line at the end and this was the easiest way to address it for now.

.github/workflows/build_test.yml
gonuke: The Dockerfile has a `cyclus-test` stage that is being build in this workflow

CMakeLists.txt
```diff
@@ -10,20 +10,16 @@ IF(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
...
-cmake_policy(SET CMP0042 OLD)
-
 # This project name is cyclus.
 PROJECT(CYCLUS)
 
 # check for and enable c++11 support (required for cyclus)
 INCLUDE(CheckCXXCompilerFlag)
-CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
+CHECK_CXX_COMPILER_FLAG("-std=c++14" COMPILER_SUPPORTS_CXX14)
 IF(COMPILER_SUPPORTS_CXX11)
```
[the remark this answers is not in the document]
gonuke: I've leapfrogged all the way to c++-17 and updated it all consistently

CMakeLists.txt
```diff
@@ -108,8 +104,10 @@ IF(NOT CYCLUS_DOC_ONLY)
     endif()
 
     # Tell CMake where the modules are
-    LIST(APPEND CMAKE_MODULE_PATH
-        "${CMAKE_DIR}/share/cmake-2.8/Modules" "${PROJECT_SOURCE_DIR}/cmake")
+    MESSAGE("-- CMAKE_MODULE_PATH: ${CMAKE_MODULE_PATH}")
+    LIST(APPEND CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}
+ #       "/opt/conda/share/cmake-3.25/Modules"
```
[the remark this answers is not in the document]
gonuke: testing now without this

test_foo.py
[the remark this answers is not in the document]
gonuke: removed

tests/cycpp_tests.py
```diff
@@ -28,7 +28,7 @@
 
 import cycpp
 
-assert_equal.__self__.maxDiff = None
+# assert_equal.__self__.maxDiff = None  !! this may not be necessary for pytest???
```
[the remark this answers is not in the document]
gonuke: I've removed it.  I don't think there is an equivalent in pytest and/or it may be a commandline options

tests/test_cycluslib.py
gonuke: I've put it back in and am testing.  I can't recall why we dropped it?

tests/tests_hdf5_back_gen.py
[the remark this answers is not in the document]
gonuke: I'll look in the pytest output

tests/tools.py
gonuke: I'm not sure these were every used in Cyclus... I think (!?!) these came over 10 years ago when this file was lifted from another project.

gonuke:
I just realized that this has not been running any of the python tests!! ☹️

gonuke:
I've converted this to draft as I add in the python tests that haven't been run during all my development

CMakeLists.txt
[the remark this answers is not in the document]
abachma2: That makes sense, but then in line 493 we set the `CPACK_DEBIAN_PACKAGE_DEPENDS` to `libxml++4.0-dev (>=4.0.0)`. Would that interfere with using either version of libxml++?

CMakeLists.txt
```diff
@@ -10,20 +10,16 @@ IF(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
...
 
-# Set some policies
-cmake_policy(SET CMP0040 OLD)
-cmake_policy(SET CMP0042 OLD)
-
 # This project name is cyclus.
 PROJECT(CYCLUS)
 
-# check for and enable c++11 support (required for cyclus)
+# check for and enable c++14 support (required for cyclus)
```
abachma2, line 16:
```suggestion
# check for and enable c++17 support (required for cyclus)
```
For consistency.

docker/Dockerfile
abachma2: That makes sense. That just wasn't clear to me.

abachma2 (commented):
A few more thoughts on the changes. 

For the Docker files, it looks like you've pulled everything into one Docker image that builds the dependencies and builds cyclus, instead of have a separate docker image for each step. Is that correct?

docker/Dockerfile
bennibbelink: Out of curiosity, why are the apt and conda stages installing different versions of cython?

gonuke:
This still has a few failing tests in the python suite, and I worry that it will take me a while to dig into them.  Maybe it's a good place for @bennibbelink to take over.  I'll add a couple of issues...

gonuke:
I think the big question is whether we are happy with the current state of this PR for now, or do we want to tackle the two remaining pytest failures first?

.github/workflows/build_test_publish.yml
```diff
@@ -0,0 +1,56 @@
+name: Build, Test & Publish docker images for future CI and other users
+
+on:
+  # allows us to run workflows manually
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - master
```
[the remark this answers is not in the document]
gonuke: Good catch

CMakeLists.txt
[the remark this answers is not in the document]
abachma2: Did we figure out if these two things conflict if a user has libxml++ 2.6?

cli/cyclus.cc
```diff
@@ -240,14 +240,14 @@ int ParseCliArgs(ArgInfo* ai, int argc, char* argv[]) {
       ("restart", po::value<std::string>(),
        "restart from the specified simulation snapshot [db-file]:[sim-id]:[timestep]")
       ("schema",
-       "dump the cyclus master schema including all installed module schemas")
+       "dump the cyclus main schema including all installed module schemas")
```
abachma2, line 243: I'm guessing you just did an auto find and replace for changing the default branch name from master to main. Did we also want to update the vocabulary for the schema?

cyclus/main.py
```diff
@@ -314,7 +314,7 @@ def make_parser():
                    help='restart from the specified simulation snapshot, '
                         'not supported.')
     p.add_argument('--schema', action=Schema,
-                   help='dump the cyclus master schema including all '
+                   help='dump the cyclus main schema including all '
```
abachma2, line 317: Similar comment as the other file about changing the terms for the schema

cyclus/simstate.py
```diff
@@ -58,9 +58,9 @@ class SimState(object):
         The initial registry to start the in-memory backend with. Defaults
         is True, which stores all of the tables.
     schema_path : str or None, optional:
-        The path to the cyclus master schema.
+        The path to the cyclus main schema.
```
abachma2, line 61: Same comment about the schema vocabulary.

doc/release_notes/v0.3.rst
```diff
@@ -85,7 +85,7 @@ New features
 - updated namespace name to reflect directory name
 - made mock fac/inst/region/market classes to be used with testing. added the stub facility in a new stub namespace and related tests.
 - added initial stub directory and adjusted cmake files to include their tests.
-- moved all dynamic loading into xml_file_loader. Added a method for listing installed/discoverable dynamic modules to env class. added rng schema methods to test agents and removed rng files from them. removed rng installation from cmake module macro. added master and module schema dumping to cyclus binary. added schema agent test (that schema parses). moved heavy stuff out of xml-file-loader constructor. renamed LoadAll to LoadSim.
+- moved all dynamic loading into xml_file_loader. Added a method for listing installed/discoverable dynamic modules to env class. added rng schema methods to test agents and removed rng files from them. removed rng installation from cmake module macro. added main and module schema dumping to cyclus binary. added schema agent test (that schema parses). moved heavy stuff out of xml-file-loader constructor. renamed LoadAll to LoadSim.
```
abachma2, line 88: Same comment about the schema vocabulary

docker/Dockerfile
abachma2: That's a good question. Does this build if you have cython 3 installed? Previously I haven't been able to get it to build with that version.

tests/test_source_to_sink.py
```diff
@@ -97,15 +106,9 @@ def check_source_to_sink(fname, source_spec, sink_spec):
...
         clean_outs()
 
 
-def test_source_to_sink():
-    cases = [("source_to_sink.xml", ":agents:Source", ":agents:Sink"),
-             ("source_to_sink.py", ":cyclus.pyagents:Source", ":cyclus.pyagents:Sink"),
-             ]
-    for case in cases:
-        for x in check_source_to_sink(*case):
-            yield x
```
abachma2, line 111: Why is this test deleted?

tests/test_stubs.py
```diff
@@ -16,7 +15,8 @@ def tmpdir():
 
 @contextmanager
 def tmplog(fname):
-    yield io.open(fname, mode='w')
+    file_ptr = open(fname, mode='w')
+    yield file_ptr
```
abachma2, line 19: Does this test still run even though you have the `yield`? I didn't think that was supported by pytest.

tests/tools.py
```diff
@@ -170,19 +167,17 @@ def libcyclus_setup():
...
-            yield f, db, fname, backend
+            f(db, fname, backend)
     return wrapper
 
 
 
-
 #
 # Here there be Hackons!
 #
```
abachma2, line 183: What? :laughing:

abachma2 (commented):
Some more comments of things that I've noticed. About half are about the vocabulary for the schema. I'm fine with the term change from `master` to `main` for that, I just wanted to make sure that was intentional. Especially since there is somewhere that you use the term `primary`

abachma2:
It looks like the GitHub Actions are all passing, but why is there still the CircleCI pipeline action? One of the purposes of the PR is to replace Circle CI wit GitHub Actions, correct? Will merging this PR remove the call to Circle CI? Is that something that we will have to change manually?

cli/cyclus.cc
gonuke: I was slightly selective... I thought this was an OK change because it's not a reserved word but a general adjective, but I didn't change the method names in the code because I didn't want to risk breaking something

cyclus/main.py
gonuke: as above - this is a generic adjective word that I thought was OK

cyclus/simstate.py
gonuke: ditto

doc/release_notes/v0.3.rst
gonuke: ditto

cli/cyclus.cc
abachma2: Do we need an adjective for this? I get that there is the core Cyclus schema, but in that sentence we say that it includes the module schemas. So I'm not sure if an adjective is needed here. Nothing that would prevent merging this PR, just an interesting note.

CMakeLists.txt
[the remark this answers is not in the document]
gonuke: If making a debian package it means we are using apt get so it should be 2.6.  I updated that.

cli/cyclus.cc
gonuke: I think it makes sense to distinguish the 'main' schema from the 'module' schemas

docker/Dockerfile
gonuke: Alas, they are the same version.  The apt build uses the (poorly named) `cython3` package which is cython version v0.29 for python 3.  The conda version is the same.

tests/test_source_to_sink.py
gonuke: It's replaced with a pytest fixture above on about line 20

tests/test_stubs.py
gonuke: yield works fine with pytest - it's used in a few places

gonuke:
> It looks like the GitHub Actions are all passing, but why is there still the CircleCI pipeline action? One of the purposes of the PR is to replace Circle CI wit GitHub Actions, correct? Will merging this PR remove the call to Circle CI? Is that something that we will have to change manually?

Yes, we'll need to drop the webhook manually... not sure the best practice on when to do that

abachma2 (approved):
Approving and merging this PR.

PR #1584, opened 2023-05-30
Add funcationality for DRE to python API

abachma2:
This PR adds some functionality for the DRE to the python API that was missing. This includes:
- `AddMutualReqs` added to the `RequestPortfolio` class -- was previously defined but not used anywhere
- `AddRecipe` added to 'lib' class (`cyclus/lib/pyx`)
- Add the ability to specify a request as exclusive.

To implement the `AddMutualReqs` and the exclusivity of the request, the structure of the dictionary to define a request, was modified to properly attach a material preference to each commodity that can meet a request and define which commodity requests are mutual.  This dictionary is read into the `request_request_portfolio_to_cpp` and `material_request_portfolio_to_cpp` functions in the `cyclus/typesystem.py` file.  

The previous format for this dictionary was: `[{"commodities":{'commod_name':Material object,...},"constraints":number},...]`. The new format is: `[{"commodities":[{'commod_name':Material object,"preference":number, "exclusive":Bool},...], "constraints":number}]`. 

In the previous format, each outer dictionary (list element) was treated as a separate request and each commodity and Material pair in the nested dictionary was treated with AND logic, such that each material was part of the entire request to be made. In the new format, each outer dictionary is still treated as separate requests. The nested inner list of dictionaries assigns the preference of each commodity, and define which commodity requests should be mutual requests (apply an OR logic to them). This new format helps to keep preferences and exclusive settings paired with each request.

The `AddRecipe` functionality is less related to the DRE, but add the ability to record recipes to the output database during a 
simulation. This was already defined in the c++ API.

PR #1594, opened 2023-10-05
Add functionality to python API

abachma2:
**This PR is a replication of #1584, but moved to an upstream branch** 

This PR adds some functionality for the DRE to the python API that was missing. This includes:

* ``AddMutualReqs`` added to the RequestPortfolio class -- was previously defined but not used anywhere
* ``AddRecipe`` added to 'lib' class (cyclus/lib/pyx)
* Add the ability to specify a request as exclusive.

To implement the ``AddMutualReqs`` and the exclusivity of the request, the structure of the dictionary to define a request, was modified to properly attach a material preference to each commodity that can meet a request and define which commodity requests are mutual. This dictionary is read into the ``request_request_portfolio_to_cpp`` and ``material_request_portfolio_to_cpp`` functions in the ``cyclus/typesystem.py`` file.

The previous format for this dictionary was: ``[{"commodities":{'commod_name':Material object,...},"constraints":number},...]``. The new format is: ``[{"commodities":[{'commod_name':Material object,"preference":number, "exclusive":Bool},...], "constraints":number}]``.

In the previous format, each outer dictionary (list element) was treated as a separate request and each commodity and Material pair in the nested dictionary was treated with AND logic, such that each material was part of the entire request to be made. In the new format, each outer dictionary is still treated as separate requests. The nested inner list of dictionaries assigns the preference of each commodity, and define which commodity requests should be mutual requests (apply an OR logic to them). This new format helps to keep preferences and exclusive settings paired with each request.

The AddRecipe functionality is less related to the DRE, but add the ability to record recipes to the output database during a
simulation. This was already defined in the c++ API.


abachma2:
Given that Docker pushes are required for running the CI (why the tests in #1584 always failed), it seems that all PRs must come from upstream branches. Is this something we want to make more flexible to allow PRs from a fork? Can we make this more flexible?

The conda CI test is failing with this PR. The SymbolicFunctionTests.expfunc unit test fails, I don't think it ran the pytests because of the failing unit test.

abachma2:
In #1584 @gonuke asked: "Can you point to where this Python data structure is now matching something that exists already in C++?"

In C++, the structures are defined in `src/request_portfolio.h`. The `RequestPortfolio` is a class with attributes `target`, `commodity`, `preference`, `exclusive`, and others that aren't relevant for these changes. This class has the `AddRequest` method that defines these attributes for an instance of the class. This class also has the `AddMutualReqs` method for creating multicommodity requests, which accepts vectors of requests. A full request portfolio is a `RequestPortfolio` pointer (which is approximated as the puter-most list in python) and each `RequestPortfolio` is approximated with a dictionary (each element of the outer-most list). The vector of requests for `AddMutualReqs` is approximated with the list of dictionaries nested into the value of `commodities`. I'm not super clear exactly how the python to c++ conversion happens, but this is my best approximations of how things translate.

gonuke:
> Given that Docker pushes are required for running the CI (why the tests in #1584 always failed), it seems that all PRs must come from upstream branches. Is this something we want to make more flexible to allow PRs from a fork? Can we make this more flexible?

Yes, we should change this. I should have known better from my experience with pyne, so it's my fault.  I was just too excited about streamlining it so much that I forgot.  I'll work on a revision that build docker images with one action and tests them with another.


> The conda CI test is failing with this PR. The SymbolicFunctionTests.expfunc unit test fails, I don't think it ran the pytests because of the failing unit test. 

I've relaunched the tests.  I thought it was transient, but I've experienced this test failing once and then passing on subsequent instances. I hoped it was gone, but it may require more investigation.

abachma2:
Interesting. Well, I'm glad the tests are all passing (although I'm not sure why they all pass if it's still running the `tests/test_main.py` file). I'll make an issue about adding flexibility for the Docker pushing. Feel free to review this PR (again I guess) and we'll move forward from there.

gonuke:
Alas... this is not passing `test_main.py` as it's not running any of the python tests.  I think once you do a rebase, we'll see how it does.  That said, we've temporarily commented out `test_main.py` in the latest PR

gonuke:
Also - is there a way to test your new python capability?  Can we run a test on the python side that is similar to an existing test on the C++ side?

abachma2:
I'll dig into the python tests some to figure something out. I can't make any promises on when I'll get around to it as I'm getting close to my defense.

gonuke:
If you can identify whether or not/where this is tested in C++, I can perhaps propose some python tests.  It may not be tested there, in which case, I'm not sure I'd hold this up for python tests.

abachma2:
A quick look leads me to the `tests/bid_portfolio_tests.cc` would be the most applicable c++ tests. I do think covering this in the python tests (not sure which one) should be done before meging this. I made some small checks while making these changes to give me confidence but we do need tests for this.

FlanFlanagan:
The code looks good to me. I'm going to test on a virtual machine this weekend!

gonuke:
I cant figure out why this is not passing tests?

abachma2:
I'm not sure either. All that should have changed since the last time they ran was pulling from the main branch. I've installed locally and the unit tests pass, but I haven't gotten around to the python tests yet.

abachma2:
The same test does not pass locally, so I'll dig into that at some point.

gonuke:
I've double checked that this test passes on the upstream

abachma2:
I have not forgotten about this PR. I have merged in some of the latest updates from the ``main`` branch. Right now there is one python test that doesn't pass. I'm spending some time to dig into this.

abachma2:
I was able to get all of the tests passing for this PR by updating the pyagents Sink archetype. This is ready for another review if you have some time @gonuke @FlanFlanagan.

cyclus/lib.pyx
```diff
@@ -1808,44 +1826,63 @@ cpdef dict normalize_request_portfolio(object inp):
     if not isinstance(inp, Mapping):
         inp = dict(inp)
     if 'commodities' in inp:
-        commods = inp['commodities']
+        commods = []
+        prefs = []
```
gonuke, line 1830: I don't think you need this variable

gonuke (changes requested):
I'm not extremely familiar with all of the detail of some of these sections of code, but only found one thing that probably needs to be updated.

cyclus/lib.pyx
abachma2: I removed the `prefs` list, because that is unused. The `commods` list is used, so that remains.

gonuke:
I think this needs a quick rebase to resolve a conflict in the Changelog

abachma2:
Let's give that a shot. I keep forgetting this PR is from an upstream branch and not from my fork.

gonuke (approved):

Issue #1595, opened 2023-10-05
Remove Docker image dependency from CI tests


abachma2:
Currently, Docker image pushing is only available from upstream branches (i.e., from cyclus/cyclus.git). Docker image pushing is required for the CI tests, so allowing Docker image pushing from any forked branch is desired so that CI can properly test for changes on PRs from a a fork. 

This issue can be closed when the Docker image can be pushed from a forked branch.

gonuke:
This is, unfortunately, not quite the solution- GitHub fundamentally prevents this as a security measure.  The real fix is for CI tests to not rely on pushing docker images.

abachma2:
Renamed the issue. So this issue can now be closed when CI tests for building and testing can be run in PRs from forked branches (which was always the ultimate goal, just adjusting how that is accomplished).

gonuke:
I think the strategy employed in svalinn/DAGMC is the best strategy here and @bennibbelink discussed today.  Hopefully it can be implemented fairly quickly.

PR #1597, opened 2023-10-13
Address Issue #1595

bennibbelink:
This PR separates the Cycles workflows into: 

- One that is run from the upstream repo that pushes new dependency images (and is only triggered on changes to the workflow/Dockerfile)
- One that can be run from contributor's forks that pulls the dependency image from cyclus/cyclus package repository, and builds/tests cyclus with those dependencies

I also added a third workflow that is triggered on a release.  It is mostly copied from the build-test-publish.yml, the main differences are the triggers and the image tag

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -1,4 +1,4 @@
-name: Linux Build/Test for PR and collaborator push
+name: Build/Test for PR and collaborator push
 
 on:
   # allows us to run workflows manually
@@ -35,7 +35,7 @@ jobs:
         ]
 
     container:
-      image: ghcr.io/${{ github.repository_owner }}/cyclus_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}
+      image: ghcr.io/cyclus/cyclus_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/${{ matrix.pkg_mgr }}-deps
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -8,6 +8,7 @@ on:
       - main
     paths-ignore:
       - '.github/workflows/build_test_publish.yml'
+      - '.github/workflows/build_test.yml'
       - '.github/workflows/changelog_test.yml'
       - 'docker/**'
       - 'doc/**'
@@ -21,7 +22,7 @@ on:
       - 'doc/**'
 
 jobs:
-  BuildTest:
+  build-and-test:
     runs-on: ubuntu-latest
 
     strategy:
@@ -45,7 +46,7 @@ jobs:
         run: python install.py -j 2 --build-type=Release --core-version 999999.999999 
 
       - name: Cyclus Unit Tests
-        run: cyclus_unit
+        run: cyclus_unit_test
 
       - name: Cyclus Python Tests
         run: pytest tests
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -12,8 +12,6 @@ on:
       - 'docker/**'
       - 'doc/**'
   push:
-    branches:
-      - main
     paths-ignore:
       - '.github/workflows/build_test_publish.yml'
       - '.github/workflows/changelog_test.yml'
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -34,7 +34,7 @@ jobs:
             apt,
             conda
         ]
-
+    
     container:
       image: ghcr.io/cyclus/cyclus_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/${{ matrix.pkg_mgr }}-deps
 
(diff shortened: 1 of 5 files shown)


bennibbelink:
I expect the Publish Dependency Image workflow to fail on this PR, since it is coming from outside the upstream.  I also expect the Build/Test workflow to fail, just because the python tests still segfault (I'm working on that).  


How should we handle contributors making changes to the Dockerfile or Publishing workflow (such as this PR)?  They run successfully on my fork, so I now have some packages in my ghcr :) but how do we test the changes on the upstream repo?

gonuke:
we may want to remote `test_main.py` from the test scheme while it's failing so that we can at least see the other tests pass

bennibbelink:
😂 oops that last commit was supposed to say "use $HOME instead of hardcoded path"

Anyways... the workflow is in a working state now

gonuke:
the conda build of the dependencies in your fork's actions failed in a strange way - failed to install `mamba`.  Can you try relaunching that to see if it was transient?

abachma2 (commented):
Hi @bennibbelink. Thanks for working on this. 

To clarify: 
- `build_test.yml` can be called from a PR from any fork, and just runs the unit and integration tests using an already created Docker image (where is this Docker image stored? in cache? I see the path for it but it's not clear to me)
- `build_test_publish.yml` can only be run (or at least pass) from an upstream branch, and it creates a new build of Cyclus in a Docker image, with the unit and integration tests run from the Docker image.
- `publish_release.yml` is meant to be run when we have a new release of Cyclus, and it runs the unit and integration tests after building a new Docker image -- so then this needs to be run from an upstream branch

Some of that was stream of conscious, but I'm trying to make sure I understand when each job is run, and when we need to have PRs from upstream branches.

bennibbelink:
> we may want to remote `test_main.py` from the test scheme while it's failing so that we can at least see the other tests pass

Re-ran and it passed, not sure what happened there.  Sometimes when I'm working locally conda will hang when trying to solve the environment, I'll keep an eye out to see if thats happening in the Actions

bennibbelink:
> the conda build of the dependencies in your fork's actions failed in a strange way - failed to install `mamba`. Can you try relaunching that to see if it was transient?

@gonuke It appears that it was hanging on the conda solve (which happens to me locally as well).  There is a new [solver](https://www.anaconda.com/blog/a-faster-conda-for-a-growing-community) that Conda released last year.  This works much faster for me, unfortunately it has compatibility issues with mamba (being addressed, see [here](https://github.com/conda-forge/miniforge/issues/497)).  Thus the docker file now installs mamba with libmamba, then uninstalls libmamba.  Another option is to perform the whole install with conda and libmamba instead of mamba, but running into linking issues right now. I'll add another comment when I get this working smoothly

gonuke:
> * `build_test_publish.yml` will only be triggered by changes to the dependencies, and should pass in a fork however the images will be published to the owners package registry (for example, this PR is coming from my fork, so you can see the packages in my package [registry](https://github.com/bennibbelink?tab=packages)).  This workflow will **not** pass in a PR, because when it tries to push the images it tries to push to the Cyclus registry but is authenticated as the contributor, so authentication fails (@gonuke may have a better grasp on this, but this is my understanding)

A few clarifications on this story for this workflow: 

- When working from a fork, this action will be triggered by pull request to the upstream and fail on the attempt to push an image.  If we also trigger on a push to branches other than `main`, then the workflow will also happen on the fork itself each time the author pushes, and it will work there.  This allows the PR author to (a) see whether it works in their own fork, and (b) at least put a link to the action in the comments.

- If a PR comes from a branch on the upstream, then the whole workflow will run successfully (assuming the workflow is correct :) )

.github/workflows/build_test_publish.yml
```diff
@@ -7,6 +7,9 @@ on:
   push:
     branches:
       - main
+    paths:
```
gonuke, line 10: If we run this on push to branches other than `main` it will run on the fork with each push, and the author can see whether it works or not

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,62 @@
...
+    branches:
+      - main
+    paths-ignore:
+      - '.github/workflows/build_test_publish.yml'
+      - '.github/workflows/changelog_test.yml'
+      - 'docker/**'
+      - 'doc/**'
+  push:
+    branches:
+      - main
```
gonuke, line 16: If we trigger on push to all branches, authors can see this be tested when they push to their own branch on their own fork.

.github/workflows/build_test.yml
```diff
@@ -0,0 +1,62 @@
...
+        run: |
+          echo "LD_LIBRARY_PATH=${HOME}/.local/lib:${HOME}/.local/lib/cyclus:$LD_LIBRARY_PATH" >> "$GITHUB_ENV"
+
+      - name: Cyclus Unit Tests
+        run: |
+          cyclus_unit_tests
+
+      - name: Cyclus Python Tests
+        run: |
+          cd tests && python -m pytest --ignore test_main.py
```
gonuke, line 62: Do you think it's important that these are different tasks in the job, or could they be just multiple lines on the same task

gonuke (changes requested):
A few subtle suggestions...

.github/workflows/build_test.yml
bennibbelink: I think it's important that the tests are separate steps so it is clear where failures happen.  I will throw the PATH and LD_LIBRARY_PATH modification in with the "Building Cyclus" step, those being their own steps is a remnant of my debugging

.github/workflows/build_test_publish.yml
bennibbelink: You're right, I'll remove the branch requirement so the workflow is triggered on pushes to any branch (assuming the `path` requirement is met)

gonuke:
Thanks @bennibbelink !! The docker build is passing on your fork and the tests are passing here. Let's give it a shot.

gonuke (approved):

PR #1599, opened 2023-10-17
Add seed in simulation init and initialize PRNG

nuclearkatie:


src/context.cc
```diff
@@ -211,6 +221,11 @@ void Context::InitSim(SimInfo si) {
       ->AddVal("DurationSecs", static_cast<int>(si.dt))
       ->Record();
 
+  NewDatum("RNGInfo")
+      ->AddVal("Seed", static_cast<int>(si.seed))
+      ->AddVal("Stride", static_cast<int>(si.stride))
+      ->Record();
+
```
gonuke, line 228: I don't think these need to be added here as well as in the "Info" block above.  I'm not sure which is better, but I'm guessing the one above will be good enough.

You may still need the `static_cast` though.

src/context.cc
```diff
@@ -22,6 +22,8 @@ SimInfo::SimInfo()
       m0(0),
       dt(kDefaultTimeStepDur),
       decay("manual"),
+      seed(kDefaultSeed),
+      stride(kDefaultStride),
```
gonuke, line 26: These should be initialized in the order they are declared in the header file.  You can either change all of these, or move the location in which they are ordered in the header file.  Note: this is a best practice and is enforced with some compiler options.  It may not be true of all current Cyclus practice.

src/sim_init.cc
```diff
@@ -160,6 +160,10 @@ void SimInit::LoadInfo() {
   // should be updated to uint64_t.
   si_.dt = qr.GetVal<int>("DurationSecs");
 
+  qr = b_->Query("RNGInfo", NULL);
+  si_.seed = qr.GetVal<int>("Seed");
+  si_.stride = qr.GetVal<int>("Stride");
+
```
gonuke, line 166: Note above comment about where the data is added - it's in two places above and I advise one that is different from this one.

gonuke (changes requested):
Thanks @nuclearkatie - a few comments on consistency here.

I'm still not sure why the tests are failing.

I'm also not sure about your commit history... perhaps you've used a merge rather than rebase approach?

src/context.cc
nuclearkatie: Wasn't sure if you had an opinion about adding to Info vs their own data table. If you think the former is fine, I'll just do that

nuclearkatie:
Yeah I just pulled without thinking (or `--rebase`ing). Happy to undo and fix to be rebase-y

gonuke:
Give it a shot... it may be messy and not worth it.  Have you ever done an interactive rebase? That might be the cleanest way, otherwise, I think you'll have many meaningless conflicts to resolve.

src/random_number_generator.h
```diff
@@ -0,0 +1,33 @@
...
+/// A random number generator.
+class RandomNumberGenerator {
+    friend class ::SimInitTest;
+  public:
+    RandomNumberGenerator();
+
+    ~RandomNumberGenerator();
+
+    /// Initialize from seed
+    void Initialize(Context* ctx, SimInfo si);
```
gonuke, line 22: Do we need the `Context`? We don't save it/use it. There may be long term value in storing the `Context` in this class, but I don't see a reason right now.

gonuke (changes requested):
A little change in the RNG initialization.

We probably need to add some testing of this, as well.

gonuke:
@bennibbelink is going to look into the best way to address the conda failures....

bennibbelink:
> @bennibbelink is going to look into the best way to address the conda failures....

Needed to include the `<cstdint>` header in `random_number_generator.h`.  This [discussion](https://stackoverflow.com/questions/14883896/why-is-stduint32-t-different-from-uint32-t) does a good job of explaining some context.  I'm not sure exactly why it broke using the newer version of `g++`, but I'm guessing that the compiler implicitly included some `std` headers and `uint32_t` was declared in the global namespace somewhere.

Regardless, including the header fixes the problem and is better practice anyway.

nuclearkatie:
Thanks @bennibbelink!

src/agent.h
gonuke, line 1: I'm a little surprised that you need to `friend` the `RandomTest` class here... and thus forward declare it as well.  Maybe it's a sign that the `random_tests` are more complicated than they should be/need to be?

tests/random_tests.cc
gonuke, line 1: I think these test may be overkill.  It looks like you probably used the `sim_init_tests` as a template, but `SimInit` does more complicated stuff than you need to do...

You might not need the `RandomInver` class at all, and may only need to make a context with a timer and a recorder in your setup method.

gonuke (changes requested):
Probably can scale back the complexity of the tests.

tests/random_tests.cc
nuclearkatie: removed

gonuke:
This looks good now... planning to merge

gonuke (approved):
LGTM

PR #1601, opened 2023-10-18
limit PR conditions for running this workflow

gonuke:
Only run the publish workflow on PRs that have changes in specific files.

Fixes #1600

PR #1602, opened 2023-10-18
limit PR conditions for running this workflow

### LINKED ISSUE #1600: Action to publish docker images not limited enough
The path and branch restrictions for `build_test_publish.yml` are a little wonky.

I think this action should run on PR's to `main` and only with specific paths changing, and on push to any branch and only with specific paths changing.

gonuke:
Only run the publish workflow on PRs that have changes in specific files.

Fixes #1600 

(replaces #1601 with local branch)

### DIFF
--- a/.github/workflows/build_test_publish.yml
+++ b/.github/workflows/build_test_publish.yml
@@ -4,6 +4,9 @@ on:
   # allows us to run workflows manually
   workflow_dispatch:
   pull_request:
+    paths:
+      - '.github/workflows/build_test_publish.yml'
+      - 'docker/**'
   push:
     branches:
       - main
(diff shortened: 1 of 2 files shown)


bennibbelink:
Your changes look good.  However, the workflows are failing the cyclus_unit_tests (SymbolicFunctionTests.expfunc, specifically).  There must be a problem elsewhere, since your changes shouldn't affect whether the tests pass or fail.  

What is best practice here?

gonuke:
Best practice seems to be rerunning the tests (not really best practice!).  I noticed this when I was setting up these workflows and this particular test fails randomly and/or 1/2 the time... so that's another bug to track down?

PR #1603, opened 2023-10-19
Add Ubuntu 20.04 support

bennibbelink:
Added 20.04 to the testing matrix and modified the Dockerfile slightly to build cyclus on 20.04.  

The only apt dependency that isn't available on 20.04 is `python3-pprintpp`, but it looks as though this is an optional dependency in the first place (see `src/pyinfile.pyx`), so we could probably drop it altogether if we want.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,8 @@
 ARG pkg_mgr=apt
 ARG make_cores=2
+ARG ubuntu_version=22.04
 
-FROM ubuntu:22.04 as common-base
+FROM ubuntu:${ubuntu_version} as common-base
 
 ENV TZ=America/Chicago
 RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -120,6 +120,7 @@ WORKDIR /cyclus
 
 # You may add the option "--cmake-debug" to the following command
 # for further CMake debugging.
+RUN mkdir -p `python -m site --user-site`
 RUN python install.py -j ${make_cores} --build-type=Release --core-version 999999.999999 
 ENV PATH /root/.local/bin:$PATH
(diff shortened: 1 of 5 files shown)


bennibbelink:
The dependency image workflow passes on my fork, publishing 4 images for the matrix [[20.04, 22.04], [conda, apt]]

docker/Dockerfile
```diff
@@ -37,9 +38,9 @@ RUN apt install -y \
         python3-pandas \
         python3-jinja2 \
         cython3 \
-        libwebsockets-dev \
-        python3-pprintpp \
-    && apt clean -y all
+        libwebsockets-dev
+RUN apt install python3-pprintpp; exit 0
```
gonuke, line 42: Could this be handled with this `apt install` options?
```suggestion
RUN apt install -m python3-pprintpp
```

gonuke (changes requested):
This looks good to me.  One little comment about how to handle the potentially missing package more cleanly (??).

I notice that when this package is not available, we use `pprint` instead.  Is that part of the standard python or do we need to try to install that?

Also, won’t we need to update the PR testing workflow as well?

docker/Dockerfile
bennibbelink: This was still throwing an error for me.  I played around with some different `apt install` options but couldn't find a more graceful way to do this

bennibbelink:
`pprint` is part of the standard library so we shouldn't need to worry about installing it

All the workflows have the updated matrix now

docker/Dockerfile
bennibbelink: ** an error that stopped the docker build

bennibbelink:
Hmm... a sticky situation with the CI.  The PR workflows will fail because the 20.04 image doesn't exist yet, but the dependency image workflow won't run to create those images since this is coming from a fork.  How do you suggest we handle this?

bennibbelink:
Closing this PR in favor of #1605 .  Need the dependency changes to come from the upstream repo so the workflows have a chance to pass.

PR #1605, opened 2023-10-19
Add Ubuntu 20.04 support

bennibbelink:
Creating this PR from the upstream repo so the workflows pass (instead of #1603).  This adds 20.04 to the testing matrix and slightly modifies the Dockerfile to work on 20.04 and 22.04.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -51,3 +51,4 @@ jobs:
       - name: Cyclus Python Tests
         run: |
           cd tests && python -m pytest --ignore test_main.py
+          
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -51,4 +51,3 @@ jobs:
       - name: Cyclus Python Tests
         run: |
           cd tests && python -m pytest --ignore test_main.py
-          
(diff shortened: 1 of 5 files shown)


bennibbelink:
When the 20.04 images are being published they are not tagged as `latest`, so when the build/test workflows try to pull the image it can't find it

gonuke:
They will get tagged as `latest` once we merge this, right?

bennibbelink:
I think so 🤞 The workflow should be triggered again on the push to `main`, and in that workflow it should be tagged as `latest` since it would satisfy `${{ github.repository_owner == 'cyclus' && github.ref == 'refs/heads/main' }}`

bennibbelink:
OR, the better way to do this might be to only include changes to the `Dockerfile` and `build_test_publish.yml` in this PR (which should pass all tests since only the dep Image workflow will be triggered), then have a separate PR (can be from my fork) that updates the other workflows

bennibbelink:
This PR now should publish dependency images for [[20.04, 22.04], [apt, conda]] with the `latest` tag when it is merged

I will make a separate PR that updates the other workflows, it's workflows should all pass once this PR is merged and new images are pushed

gonuke (approved):
Looks good - thanks @bennibbelink

PR #1607, opened 2023-10-19
Pull cyclus image from cyclus ghcr

### LINKED ISSUE #1606: Build_test workflow pulls cyclus-deps image from owner repository
This is a problem if a contributor wants to test on their own fork, but hasn't run the dependency image workflow (thus they don't have dependency images in their ghcr)

The solution is to hardcode the image as coming from the cyclus ghcr.  The downside of this is that a contributor will have to do some modifications in their fork if they want to run workflows on their own image, but since any PRs that affect the dependencies should be coming from the upstream repo, they will likely have to merge dependency changes first in a different PR (this may require a more in-depth PR review for the two PRs)

bennibbelink:
Closes #1606

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -32,7 +32,7 @@ jobs:
         ]
     
     container:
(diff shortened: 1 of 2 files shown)


gonuke (approved):
Thanks @bennibbelink

PR #1608, opened 2023-10-20
Add 20.04 support for PR and Publish workflows

bennibbelink:
Relies on #1605  to be merged first.  The `build_test.yml` needs the dependency images for 20.04 to be published as `latest` before all the workflows will pass

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -39,6 +39,11 @@ jobs:
       - name: Checkout and Build Cyclus
         uses: actions/checkout@v3
 
+      - name: echo some stuff
+        run: |
+          python -m site --user-site
+          ls /github/home/.local/lib/python3.8/site-packages
+
       - name: Building Cyclus
         run: |
           python install.py -j 2 --build-type=Release --core-version 999999.999999 
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -39,11 +39,6 @@ jobs:
       - name: Checkout and Build Cyclus
         uses: actions/checkout@v3
 
-      - name: echo some stuff
-        run: |
-          python -m site --user-site
-          ls /github/home/.local/lib/python3.8/site-packages
-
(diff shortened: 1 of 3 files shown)


bennibbelink:
In doing this I found an issue with #1605.  In the Dockerfile the site-packages directory is made in the `cyclus` build stage instead of the `apt-deps` stage.  I added the `mkdir` command to `build_test.yml` to  confirm that this was the solution to get these workflows passing.  

It's probably better to update the Dockerfile directly instead of adding steps to the build_test workflow. (also the `PYTHONPATH` is hardcoded to python3.10, so that should be changed as well)

gonuke:
Happy to review merge a new PR fore this.

Is there an elegant way to determine the right package directory so that it can adapt to the python version?

gonuke:
I just made #1609 to handle it, I think

bennibbelink:
> Happy to review merge a new PR fore this.
> 
> Is there an elegant way to determine the right package directory so that it can adapt to the python version?

Yeah I think we should be able to something similar to the `mkdir` command I added... `python -m site --user-site`, I'll see if that will work when setting env vars in the Dockerfile

bennibbelink:
This is a little puzzling but I think I know what the problem is here.  When the workflow runs on the GitHub runner, the python site-packages directory is `/github/home/.local/lib/python3.8/site-packages/`. Even though we moved the `mkdir` command to the deps stage of the Dockerfile (where it is something along the lines of `/root/.local/lib/python...` , that created directory isn't where the install script is trying to install files.  

The solution I see to this is include a `mkdir` command in the `build_test.yml` workflow (so the workflows pass), on top of the one we already have in the Dockerfile (so users can build on that dependency image)

gonuke:
Hmmm - I remember seeing something about this - the workflow runs in a different home dir than the docker ...

gonuke:
Thanks @bennibbelink

gonuke (approved):

PR #1609, opened 2023-10-21
Fix pythonpath

### LINKED ISSUE #1610: Make sure python path is created and set in correct place
The PYTHONPATH variable should point to the user site location.  This location can be determined with:
`python -m site --user-site`

It was added recently to the `Dockerfile` but only in the `cyclus` stage and not the dependency state.

Two thinks should be done in each of the different dependency paths (apt, conda):
* [x] ensure the directory exists
* [x] set the PYTHONPATH variable

gonuke:
Ensure user-site path is created at right stage.

Fix PYTHONPATH to be not hardcoded

Fixes #1610

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -45,7 +45,6 @@ RUN apt clean -y all
 
 RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 10
 RUN mkdir -p `python -m site --user-site`
-ENV PYTHONPATH `python -m site --user-site`
 
 FROM common-base as conda-deps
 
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -44,6 +44,8 @@ RUN apt install -y python3-pprintpp; exit 0
 RUN apt clean -y all
 
 RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 10
+RUN mkdir -p `python -m site --user-site`
+ENV PYTHONPATH `python -m site --user-site`
```
bennibbelink, line 48: When I build this image I'm seeing `$PYTHONPATH` equal to "\`python -m site --user-site\`" and not the actual output of that command (confirmed by various [articles](https://forums.docker.com/t/how-do-i-send-runs-output-to-env-in-dockerfile/16106)).  Might need to figure out something more clever to set this dynamically

bennibbelink: The options I'm seeing are:
1. set it in the `CMD` (don't think this will work with multiple stages since the `RUN` steps will run in different shells
2. store it in a file? (doesn't really solve anything if we need to run a script to set the env var)
3. change the site-packages directory to something we know, so that we know where cyclus is installed and we can safely hardcode the PYTHONPATH (I'm trying to see if this is possible)

bennibbelink: Actually... do we need to explicitly set the `PYTHONPATH` env var?  When I omit that line and build images, the correct site-packages directory is in `sys.path` and all tests pass (besides test_main.py of course). The best solution might be to delete line 48 altogether

bennibbelink (approved):
Looks good, thanks @gonuke

PR #1611, opened 2023-10-21
Boost warnings

bennibbelink:
The Boost deprecation warnings are gone now.  

The most significant changes involve converting `<boost/detail/sp_typeinfo.hpp>` to `<boost/core/typeinfo.hpp` and its declarations.

I also cleaned up some unnecessary `#include`s and added some `#include`s to explicitly declare `<boost/shared_ptr.hpp>` where it had been declared implicitly through other header files (not necessary but might prevent issues in the future)

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -24,6 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         ubuntu_versions : [
+          20.04,
           22.04,
           ]
         pkg_mgr : [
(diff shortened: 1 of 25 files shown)


gonuke (approved):
This looks great!! I thought it might be straightforward, but they did make a clutter. Thanks @bennibbelink

gonuke:
I'll merge once this goes out of draft

bennibbelink:
Thanks! It was more straightforward than I was expecting thankfully.  I'll move this out of draft once we get #1608 merged, I haven't tested these boost changes on 20.04 so I'd like to see the new testing matrix run first

gonuke (approved):

PR #1613, opened 2023-11-06
Fix CBC usage in Cyclus

bennibbelink:
This PR currently contains fixes to the Cyclus build procedure (in CMakeLists.txt and Dockerfile/workflows), as well as changes to how the CbcSolver is being called in Cyclus.

I removed the different versions of OsiCbcSolverInterface that we had in Cyclus since the latest minor version is 2.10, and the oldest platform that we currently support (ubuntu 20.04) supports cbc 2.10.7.  May be able to clean this up in the codebase some more, I will take another look.

It also includes a slight change to one of the unit tests, which was failing like this:
```
/cyclus/tests/equality_helpers.h:80: Failure
Value of: act
  Actual: 0.6000000000000002
Expected: exp
Which is: 0.59999999999999964
soln arrays differ at index 6
 exp: 0.59999999999999964
 act: 0.6000000000000002
```
The old test checked for equality.  I modified it (and created a new helper method) to use `EXPECT_NEAR` with some margin of error instead of `EXPECT_DOUBLE_EQ`.  This seems like a reasonable change to make, but please let me know if not for some reason.

Since this PR includes changes to the dependency image workflow, we may need to split it up into 2 (one from an upstream branch to make workflow changes, one from my fork to make other changes).  I just wanted to get this up somewhere in one piece since it is in a good state on my fork.


bennibbelink:
Closing this in favor of #1614

PR #1616, opened 2023-11-12
Add Code Coverage using Coveralls

### LINKED ISSUE #1591: introduce code coverage measurement to CI
It will be valuable to have some measure of how much of our code we are covering in CI.  Through work on #1588 I got the impression that there are core methods implemented in C++ that are tested only in the python testing. I'm not sure that's a good state to be in.  It would be useful to know how much coverage we get from gtest alone vs gtest + pytest.

bennibbelink:
Adds a build option to enable code coverage and generates coverage reports in the `build_test.yml` workflow.

The workflows will generate html artifacts that can be downloaded and push coverage reports to [Coveralls](https://coveralls.io/github/cyclus/cyclus).  The settings in Coveralls allow us to define failure thresholds for PRs.

Fixes #1591

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -29,7 +29,7 @@ jobs:
           ]
         pkg_mgr : [
             apt,
-            conda
+            conda,
         ]
     
     container:
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -60,7 +60,7 @@ jobs:
 
       - name: Initial Coverage Report
         run: |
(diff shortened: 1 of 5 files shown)


gonuke (changes requested):
LGTM - I think we need to merge this to properly understand how it interacts with coveralls

gonuke:
I think we need to resolve the CHANGELOG merge conflict

bennibbelink:
I addressed the merge conflict. 

I agree, I'm not entirely sure what needs to happen for the chat-bot to work.  At the very least this puts the coverage infrastructure in place and we can play around with the different features that coveralls provides

gonuke (approved):
Merging now

PR #1629, opened 2023-11-27
Build new docker images on all pushes to `main` not just with specifi…

### LINKED ISSUE #1628: Build new cyclus docker image for downstream testing on push to `main`
We currently only push new docker images when we change the docker publish workflow or the Dockerfile.  However, to enable testing of downstream packages (E.g. Cycamore) we need to have new docker images build on any merge to `main`.

Review our testing strategy to determine the best combination of packages to have available.  Probably something marked `stable` which is the last released version and something marked `latest` that is the last push to `main`.

At the moment we may not get to have a `stable` until we get a release out the door - perhaps that makes it urgent?

gonuke:
Fixes #1628

We currently only pushed new docker images when pushes to `main` were caused by changes in specific files.  This means that there is no up-to-date image based on the most current state of `main` for downstream repos.

This fix causes all pushes to `main` to cause an update of the docker images tagged as `latest`.  In the future, once we have a working release, we can add to the images with a `stable` based on the latest release.

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -9,12 +9,14 @@ on:
       - '.github/workflows/changelog_test.yml'
       - 'docker/**'
       - 'doc/**'
+      - 'CHANGELOG.rst'
   push:
     paths-ignore:
       - '.github/workflows/build_test_publish.yml'
       - '.github/workflows/changelog_test.yml'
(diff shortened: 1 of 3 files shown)


bennibbelink (approved):
Looks good to me.

PR #1632, opened 2023-12-01
Cmake updates and python installation best practices

bennibbelink:
This includes lots of minor cmake changes and attempts to use modern python best practices for installing cyclus' python package

The apt-20.04 build_test.yml fails because the dependency images don't have pip installed, but it passes in the build_test_publish.yml.  Should I make a separate PR to add pip as a dependency, then address this PR so the workflows pass?


bennibbelink:
#1633 will add pip as a dependency, then these workflows should pass on the new images

gonuke (approved):
Thanks @bennibbelink

PR #1634, opened 2023-12-04
Implement randomness into active and dormant buying periods in toolkit/buy policy

### LINKED ISSUE #1622: Implement randomness into active and dormant buying periods in toolkit/buy policy

nuclearkatie:
Closes #1622 and #1641


gonuke:
I'm not sure why that conda test is failing... I'm investigating now

gonuke:
Now that I fixed the docker image this is passing all tests- happy to consider merging

nuclearkatie:
Oops, forgot it was still in draft status @gonuke, please merge or add more review if necessary. Thanks!

src/random_number_generator.cc
```diff
@@ -57,7 +57,54 @@ namespace cyclus{
...
         int rounded_val = std::lrint(val);
+        return rounded_val;
+    }
+
+    //
+    // Distributions
+    UniformDoubleDist::UniformDoubleDist(double min, double max) : dist(min, max) {}
+
+    double UniformDoubleDist::sample() { 
+        return UniformDoubleDist::dist(RandomNumberGenerator::gen_);
```
gonuke, line 68: Shouldn't this just be:

```suggestion
        return dist(RandomNumberGenerator::gen_); 
```

since `dist` is a member variable of this particular object?

src/random_number_generator.cc
```diff
@@ -57,7 +57,54 @@ namespace cyclus{
...
+    //
+    // Distributions
+    UniformDoubleDist::UniformDoubleDist(double min, double max) : dist(min, max) {}
+
+    double UniformDoubleDist::sample() { 
+        return UniformDoubleDist::dist(RandomNumberGenerator::gen_); 
+    }
+
+    double UniformDoubleDist::max() { 
+        return UniformDoubleDist::dist.max();
```
gonuke, line 72: same as above

gonuke (changes requested):
It seems that you are over-specifying the `dist` in many of your routines here?  I put a couple of comments, but not sure if I'm missing something.

gonuke:
> something from PR #1640 seems to have prevented the CI from being able to install dependencies

I think that's my fault for suggesting the `Dockerfile` change to force `libxmlpp-4.0` in this branch.  Any changes to the `Dockerfile` force it to try and rebuild the CI images, but the CI images can only be rebuilt from a branch of the upstream repo.

If you rebase on `main` that `Dockerfile` change will go away and then it should pass all tests.

src/toolkit/matl_buy_policy.cc
```diff
@@ -55,34 +53,45 @@ void MatlBuyPolicy::set_throughput(double x) {
...
 }
 
-void MatlBuyPolicy::set_active(int x) {
-  assert(x > 0);
-  active_ = x;
+void MatlBuyPolicy::set_default_distributions() {
+  active_dist_ = new FixedIntDist(1);
+  dormant_dist_ = new FixedIntDist(-1);
+  size_dist_ = new FixedDoubleDist(1.0);
+  init_active_dormant();
```
gonuke, line 60: Since this gets called explicitly in cases where you don't use default distributions, I think it helps readability to call it explicitly after calling `set_default_distributions` in all the places where you do use the defaults.

src/toolkit/matl_buy_policy.cc
```diff
@@ -55,34 +53,45 @@ void MatlBuyPolicy::set_throughput(double x) {
...
 }
 
-void MatlBuyPolicy::set_dormant(int x) {
-  assert(x >= 0);
-  dormant_ = x;
+void MatlBuyPolicy::init_active_dormant() {
+  next_active_end_ = active_dist_->sample();
```
gonuke, line 64: Doesn't this need to be added to `manager()->context()->time()` in case this policy gets initialized after time 0?

Or just call `SetNextActiveTime()`

gonuke: What happens if `next_active_end_ < 0`?

src/toolkit/matl_buy_policy.cc
```diff
@@ -199,10 +226,32 @@ void MatlBuyPolicy::AcceptMatlTrades(
...
 
+void MatlBuyPolicy::SetNextActiveTime() {
+  next_active_end_ = active_dist_->sample() + manager()->context()->time();
+  return;
+};
+
+void MatlBuyPolicy::SetNextDormantTime() {
+  if (next_dormant_end_ < 0) {}
+  else {
+    next_dormant_end_ = dormant_dist_->sample() + manager()->context()->time();
```
gonuke, line 242: doesn't this need to be
```suggestion
    next_dormant_end_ = dormant_dist_->sample() + next_active_end_;
```

src/toolkit/matl_buy_policy.h
```diff
@@ -128,15 +141,18 @@ class MatlBuyPolicy : public Trader {
...
   void Stop();
 
+  double sample_fraction_ = 1.0;
+
   /// the total amount requested
   inline double TotalQty() const {
-    return std::min(throughput_,
-                    fill_to_ * buf_->capacity() - buf_->quantity());
+    return std::min((throughput_* sample_fraction_),
+                    ((fill_to_ * buf_->capacity()) - buf_->quantity()) * sample_fraction_);
```
gonuke, line 149: Since 
* `sample_fraction` is only needed as part of `TotalQty()`, 
* it is set by `SetSampleSize()`, and
* `SetSampleSize()` must be called before `TotalQty()` and only then

why not just add the code from `SetSampleSize()` to `TotalQty()` and get rid of the former and the member variable.

gonuke: Also, perhaps multiply the result of the `std::min()` by `sample_fraction_` instead of multiplying both and then taking the min?

src/toolkit/matl_buy_policy.h
```diff
@@ -173,13 +193,22 @@ class MatlBuyPolicy : public Trader {
...
-  void set_dormant(int x);
+  void set_default_distributions();
+  void init_active_dormant();
 
   ResBuf<Material>* buf_;
   std::string name_;
   double fill_to_, req_when_under_, quantize_, throughput_;
-  int active_, dormant_;
+  
+  double cumulative_capacity_;
```
gonuke, line 203: This isn't used.

src/toolkit/matl_buy_policy.h
```diff
@@ -85,15 +86,27 @@ class MatlBuyPolicy : public Trader {
...
   /// Note that active and dormant periods are note currently compatible with
   /// (s, S) inventory management
   /// @{
   MatlBuyPolicy& Init(Agent* manager, ResBuf<Material>* buf, std::string name);
   MatlBuyPolicy& Init(Agent* manager, ResBuf<Material>* buf, std::string name,
-                      double throughput, int active = 1, int dormant = 0);
+                      double throughput, 
+                      IntDistribution* active_dist = new FixedIntDist(1),
+                      IntDistribution* dormant_dist = new FixedIntDist(0),
+                      DoubleDistribution* size_dist = new FixedDoubleDist(1.0));
```
gonuke, line 109: This may create a memory leak if these are not deleted somewhere.  However, we can't **always** delete them since they may be passed in from somewhere else and we don't have the "authority" to delete them.

I think the best solution is to use `NULL` as the default and then instantiate these in the constructor if the value is `NULL`.  Then, you'll also want these variables to use `boost::shared_ptr<>`.  This should automatically track whether there continue to be references to the objects in question.

src/toolkit/matl_buy_policy.cc
```diff
@@ -153,24 +164,40 @@ std::set<RequestPortfolio<Material>::Ptr> MatlBuyPolicy::GetMatlRequests() {
   std::set<RequestPortfolio<Material>::Ptr> ports;
   bool make_req = buf_->quantity() < req_when_under_ * buf_->capacity();
   double amt;
-  
-  if (dormant_ > 0 && manager()->context()->time() % (active_ + dormant_) < active_) {
+
+  int current_time_ = manager()->context()->time();
+
+  if (current_time_ < next_active_end_ || next_dormant_end_ < 0) {
```
gonuke, line 170: A little inline private function might make this more readable:
```
bool never_dormant() { return next_dormant_end_ < 0;};
```

gonuke (changes requested):
(I'm not done reviewing but need to move on to something else, so I'm just submitting these for now.)

I need to think through the usage of `next_active_end_` and `next_dormant_end_` to see if there is something simpler...

src/toolkit/matl_buy_policy.cc
nuclearkatie: I think these are functionally equivalent? `SetNextDormantTime` is only called at the end of an active period, so `manager()->context()->time()` is the same as `next_active_end_`. If you think using `next_active_end_` is better, 👍

src/toolkit/matl_buy_policy.h
nuclearkatie: Oops, left over from thinking about adding cumulative capacity (which is the next step)

src/random_number_generator.cc
```diff
@@ -57,7 +57,54 @@ namespace cyclus{
...
+    }
+
+    UniformIntDist::UniformIntDist(int min, int max) : dist(min, max) { }
+
+    int UniformIntDist::sample() {
+        return dist(RandomNumberGenerator::gen_);
+    }
+
+    NormalIntDist::NormalIntDist(double mean, double std_dev, int min, int max) : dist(mean, std_dev) {
+        min_ = std::min(min, 0);
```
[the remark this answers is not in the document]
nuclearkatie: Guess I was only thinking of applying distributions to things like time or size, which have physical constraints. Do you think there's an obvious use case for negative numbers?

src/toolkit/matl_buy_policy.h
nuclearkatie: But the defaults aren't overwritten until init, so the values would always be null at construction

src/toolkit/matl_buy_policy.cc
nuclearkatie: If I add a call to `manager()->context()->time()` here I get a seg fault when building

src/toolkit/matl_buy_policy.h
nuclearkatie: That's not exactly true. Yes, , `SetRequestSize()` is only called immediately before `TotalQty()` in `matl_buy_policy.cc` directly, but `TotalQty()` is also called through member functions `ReqQty()` or `NReq()`, which also happens in the same time step. If I move `sample_fraction` to be part of `TotalQty()`, then `sample_fraction` will be re-set multiple times in a single timestep, because it calls a new sample every time

docker/Dockerfile
gonuke, line 1: I think this change will collapse when you rebase on the current main branch, and then it won't try to do the CI-image-building tests that are failing.

src/random_number_generator.cc
```diff
@@ -57,7 +57,38 @@ namespace cyclus{
...
         int rounded_val = std::lrint(val);
+        return rounded_val;
+    }
+
+    //
+    // Distributions
+    NormalDoubleDist::NormalDoubleDist(double mean, double std_dev, double min, double max) : dist(mean, std_dev) {
+        min_ = min;
+        max_ = max;
+    }
```
gonuke, line 68: This can probably be inlined in the header.

src/random_number_generator.cc
```diff
@@ -57,7 +57,38 @@ namespace cyclus{
...
+        double val = dist(RandomNumberGenerator::gen_);
+        while (val < min_ || val > max_){
+            val = dist(RandomNumberGenerator::gen_);
+        }
         return val;
     }
 
+    double NormalDoubleDist::max() { 
+        return dist.max();
+    }
```
gonuke, line 80: This can be inlined in the header

src/random_number_generator.cc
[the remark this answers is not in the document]
gonuke: I don't think there is anything obvious, but I'm not sure there is a reason to disallow it.

src/toolkit/matl_buy_policy.cc
gonuke: It's odd that it seg faults when building???  I don't have the fine details of the order of initializing everything in my head, but conceptually, I could imagine a seg fault if an agent invokes this initialization before the context has initialized time - I'd have to dig to see if that's what's happening and then figure out what the best path forward would be.  I think the agent has to exist (?? and been added to the context ??) before you `Init()` the buy policy.

In general, it seems that the first time that `next_active_end_` is set does need to be relative to the time at which the agent enters the simulation.  Calling `SetNextActiveTime()` should take care of that (but it may also result in a segfault?)

Our tests may not catch this and may always have agents enter at time 0 - if that's true, it's not your problem to fix as part of this PR, but we should make an issue to improve testing.

src/toolkit/matl_buy_policy.cc
```diff
@@ -55,34 +53,46 @@ void MatlBuyPolicy::set_throughput(double x) {
...
+void MatlBuyPolicy::init_active_dormant() {
+  next_active_end_ = active_dist_->sample();
+ 
+  int dormant_len = dormant_dist_->sample();
+  if (dormant_len < 0) {
+    next_dormant_end_ = -1;
+  }
+  else {
+    next_dormant_end_ = dormant_len + next_active_end_;
+  }
```
gonuke, line 71: Assuming it makes sense to call `SetNextActiveEnd()` above, it will make this more readable if you just call `SetNextDormandEnd()` here as well.  As long as you set `next_dormant_end_` to be -1 before then, when appropriate, it will result in the same behavior, but waste a single random number.

src/toolkit/matl_buy_policy.cc
```diff
@@ -153,24 +167,40 @@ std::set<RequestPortfolio<Material>::Ptr> MatlBuyPolicy::GetMatlRequests() {
...
+    SetNextDormantTime();
+  }
+  // the following is an if rather than if-else statement because if dormant
+  // length is zero, buy policy should return to active immediately
+  if (current_time_ == next_dormant_end_) {
+    // finished dormant. starting buying and sample/set length of active period
+    SetRequestSize();
+    amt = TotalQty();
+    SetNextActiveTime();
   }
```
gonuke, line 195: This section may be more readable as (pseudo-code):
```
if (!never_dormant() || current_time_ < next_active_end_) {
       do something
}
else if (current_time_ == next_active_end) {
    do other things
}
else if (current_time_ < next_dormant_end_) {
   do still different things
}
else if (current_time_ == next_dormant_end_) {
  do last batch of things
}
```

src/toolkit/matl_buy_policy.h
gonuke: Oops! I keep getting `Init()` confused with the constructor... let me rephrase:

I think the best solution is to use `NULL` as the default and then instantiate these in `Init()` if the value is NULL. Then, you'll also want these variables to use boost::shared_ptr<>. This should automatically track whether there continue to be references to the objects in question.

src/toolkit/matl_buy_policy.h
```diff
@@ -128,15 +141,18 @@ class MatlBuyPolicy : public Trader {
...
   void Stop();
 
+  double sample_fraction_ = 1.0;
+
   /// the total amount requested
   inline double TotalQty() const {
-    return std::min(throughput_,
-                    fill_to_ * buf_->capacity() - buf_->quantity());
+    return sample_fraction_ * std::min(throughput_,
+           (fill_to_ * buf_->capacity()) - buf_->quantity());
```
gonuke, line 149: I think you just want to move the logic/code from `SetRequestSize()` into `TotalQty()` since that code is needed if and only if `TotalQty()` is called.

tests/toolkit/matl_buy_policy_tests.cc
```diff
@@ -181,12 +182,49 @@ TEST_F(MatlBuyPolicyTests, MultiReqQuantize) {
...
-TEST_F(MatlBuyPolicyTests, ActiveDormant) {
-  int active = 2;
-  int dormant = 3;
-  int dur = 4;
+TEST_F(MatlBuyPolicyTests, DefaultFixedActiveDormant) {
+  using cyclus::QueryResult;
+  
+  FixedIntDist a_dist = FixedIntDist(1);
+  FixedIntDist d_dist = FixedIntDist(-1);
+  FixedDoubleDist size_dist = FixedDoubleDist(1.0);
```
gonuke, line 190: Doesn't it make sense for this test to call `Init()` with no values for these and rely on the defaults defined in the header?  That will test both:

- that the defaults result in the default behavior
- the default behavior is to have no dormant phase

tests/toolkit/matl_buy_policy_tests.cc
```diff
@@ -181,12 +182,49 @@ TEST_F(MatlBuyPolicyTests, MultiReqQuantize) {
...
+  EXPECT_EQ(1, second_cycle);
+
+  delete a;
+}
+
+TEST_F(MatlBuyPolicyTests, FixedActiveDormant) {
+  using cyclus::QueryResult;
+  
+  FixedIntDist a_dist = FixedIntDist(1);
+  FixedIntDist d_dist = FixedIntDist(1);
```
gonuke, line 224: I'm not sure what our best practice should be, but you could re-use these distributions if their identical, especially if they are `FixedIntDist`s
```suggestion
  FixedIntDist a_d_dist = FixedIntDist(1);
```
and pass `a_d_dist` for both the active and dormant distribution.

tests/toolkit/matl_buy_policy_tests.cc
```diff
@@ -235,13 +320,187 @@ TEST_F(MatlBuyPolicyTests, ActiveDormantMultipleCycles) {
...
+  sim.context()->AddPrototype(a->prototype(), a);
+  sim.agent = sim.context()->CreateAgent<cyclus::Agent>(a->prototype());
+  sim.AddSource("commod1").Finalize();
+
+  TestFacility* fac = dynamic_cast<TestFacility*>(sim.agent);
+
+  cyclus::toolkit::ResBuf<cyclus::Material> inbuf;
+  cyclus::toolkit::MatlBuyPolicy policy;
+  policy.Init(fac, &inbuf, "inbuf", throughput, &a_dist, &d_dist, &size_dist)
+        .Set("commod1").Start();
```
gonuke, line 429: If your default was `NULL` you could also actively just pass `NULL` to get default behavior.
```suggestion
  policy.Init(fac, &inbuf, "inbuf", throughput, NULL, NULL, &size_dist)
        .Set("commod1").Start();
```

gonuke (changes requested):
This is looking pretty good.  I think we're nearly there - happy to discuss any of these requests/suggestions.

src/toolkit/matl_buy_policy.h
```diff
@@ -128,23 +142,26 @@ class MatlBuyPolicy : public Trader {
...
 
-  /// the total amount requested
-  inline double TotalQty() const {
+  double random_request_size_ = 1.0;
+
+  /// the total amount available to request
+  inline double TotalAvailable() const {
     return std::min(throughput_,
-                    fill_to_ * buf_->capacity() - buf_->quantity());
+           (fill_to_ * buf_->capacity()) - buf_->quantity());
```
gonuke, line 150: The original indentation is probably preferred

gonuke (changes requested):
I've left a couple of little suggestions here.

src/random_number_generator.h
```diff
@@ -76,6 +84,74 @@ class RandomNumberGenerator {
...
+
+class NormalDoubleDist : public DoubleDistribution {
+  private:
+    boost::random::normal_distribution<> dist;
+    double min_;
+    double max_;
+  public:
+    NormalDoubleDist(double mean, double std_dev, double min=0, double max=std::numeric_limits<double>::max()) : dist(mean, std_dev), min_(min), max_(max) {};
+    virtual double sample();
+    virtual double max() { return dist.max(); }
```
gonuke, line 119: Should this be `dist.max()` or `max_`?

src/random_number_generator.h
```diff
@@ -76,6 +84,74 @@ class RandomNumberGenerator {
...
+
+class NormalIntDist : public IntDistribution {
+  private:
+    boost::random::normal_distribution<> dist;
+    int min_;
+    int max_;
+  public:
+    NormalIntDist(double mean, double std_dev, int min=0, int max=std::numeric_limits<int>::max()) : dist(mean, std_dev), min_(min), max_(max) {};
+    virtual int sample();
+    virtual int max() { return dist.max(); }
```
gonuke, line 152: `dist.max()` or `max_`?

src/toolkit/matl_buy_policy.cc
```diff
@@ -153,24 +170,39 @@ std::set<RequestPortfolio<Material>::Ptr> MatlBuyPolicy::GetMatlRequests() {
...
+    amt = TotalAvailable() * SampleRequestSize();
   }
-  else if (dormant_ == 0)
-    amt = TotalQty();
-  else {
-    // in dormant part of cycle, return empty portfolio
+  else if (current_time_ == next_active_end_) {
+    // finished active. starting dormancy and sample/set length of dormant period
+    amt = 0;
+  }
```
gonuke, line 183: How is this condition different from the next one?  Do you need it?

src/toolkit/matl_buy_policy.cc
```diff
@@ -153,24 +170,39 @@ std::set<RequestPortfolio<Material>::Ptr> MatlBuyPolicy::GetMatlRequests() {
...
     amt = 0;
-    LGH(INFO3) << "in dormant period, no request";
+    LGH(INFO3) << "in dormant period, no request" << std::endl;
+  }
+  // the following is an if rather than if-else statement because if dormant
+  // length is zero, buy policy should return to active immediately
+  if (current_time_ == next_dormant_end_) {
+    // finished dormant. starting buying and sample/set length of active period
+    SetRequestSize();
+    amt = TotalAvailable() * random_request_size_;
```
gonuke, line 194: Update as above (assuming that `SetRequestSize()` returns a double as commented below)

src/toolkit/matl_buy_policy.cc
```diff
@@ -199,10 +231,32 @@ void MatlBuyPolicy::AcceptMatlTrades(
...
+  return;
+}
+
+void MatlBuyPolicy::SetRequestSize() {
+  random_request_size_ = size_dist_->sample();
+  if (random_request_size_ > 1) {
+    random_request_size_ = random_request_size_ / size_dist_->max();
+  }
+  return;
+}
```
gonuke, line 258: With the above update, this now should return a `double` and we can eliminate the member variable `random_request_size_` in favor of a local variable here.

tests/toolkit/matl_buy_policy_tests.cc
```diff
@@ -226,25 +277,239 @@ TEST_F(MatlBuyPolicyTests, ActiveDormantMultipleCycles) {
...
+  int second_cycle = qr.GetVal<int>("Time", 6);
+  EXPECT_EQ(9, second_cycle);
+  // third cycle should include time 19
+  int third_cycle = qr.GetVal<int>("Time", 13);
+  EXPECT_EQ(19, third_cycle);
+
+  delete a;
+}
+
+// TEST_F(MatlBuyPolicyTests, MixedActiveDormant) {
```
gonuke, line 378: Are these tests commented out on purpose?

gonuke (changes requested):
(Not sure if this was ready for review... or maybe only partly finished?) but here are some comments

src/random_number_generator.h
nuclearkatie: ahh, it should be `max_`

tests/toolkit/matl_buy_policy_tests.cc
nuclearkatie: oops, no. at least, not anymore

nuclearkatie:
@gonuke addressed all comments (I think) in the latest commits.

I think I figured out what's going on with the seg fault in testing (not build, that was my mistake). So, if a buy policy is initialized normally, with a manager, then a timer (and a time, I think) exist at init.

However, I realized that the test that was giving me the seg fault is `StartStop`, which tests for a `ValueError` when a policy is initialized with a NULL manager. NULL manager = no access to `time()` = seg fault.

I addressed this by checking if manager is NULL as a first thing during init (`set_manager`), and throwing a ValueError at that point. Then, I can call `SetNextActiveTime()` and `SetNextDormantTime()` without issue because anything that makes it to `init_active_dormant` should be confirmed to have a non-null manager, and thus be reasonably expected to access `time()` with no difficulty. This does affect previous tests because now you can't get to start (or stop) if you don't have a manager. I assume that's alright?

src/toolkit/matl_buy_policy.cc
```diff
@@ -153,24 +184,35 @@ std::set<RequestPortfolio<Material>::Ptr> MatlBuyPolicy::GetMatlRequests() {
...
+
+  if (never_dormant() || current_time_ < next_active_end_) {
+    // currently in the middle of active buying period
+    amt = TotalAvailable() * SampleRequestSize();
   }
-  else if (dormant_ == 0)
-    amt = TotalQty();
-  else {
-    // in dormant part of cycle, return empty portfolio
+  else if (current_time_ == next_active_end_ || current_time_ < next_dormant_end_) {
```
gonuke, line 194: The only way to get here is if `next_dormant_end_ > 0` and `next_active_end_ < current_time < next_dormant_end_ ` so I think you only need to test against `next_dormant_end_`:
```suggestion
  else if (current_time_ < next_dormant_end_) {
```

src/toolkit/matl_buy_policy.cc
```diff
@@ -199,10 +241,33 @@ void MatlBuyPolicy::AcceptMatlTrades(
...
+    next_dormant_end_ = dormant_dist_->sample() + next_active_end_;
+  }
+  return;
+}
+
+double MatlBuyPolicy::SampleRequestSize() {
+  double random_request_size_ = size_dist_->sample();
+  if (random_request_size_ > 1) {
+    random_request_size_ = random_request_size_ / size_dist_->max();
+  }
```
gonuke, line 266: This seems like it will give inconsistent behavior - if `size_dist_->max() > 1` it's possible to get a random sample that is less than 1 that will NOT be normalized by `size_dist_->max()` and then another sample that is greater than 1 that WILL be normalized.  

Maybe the condition should be to divide by `size_dist_->max()` if `size_dist_->max() > 1`??

tests/toolkit/matl_buy_policy_tests.cc
```diff
@@ -235,13 +324,182 @@ TEST_F(MatlBuyPolicyTests, ActiveDormantMultipleCycles) {
...
+  EXPECT_EQ(8, third_cycle);
+
+  delete a;
+}
+
+TEST_F(MatlBuyPolicyTests, NormalActiveDormant) {
+  using cyclus::QueryResult;
+
+  boost::shared_ptr<NormalIntDist> a_dist = 
+  boost::shared_ptr<NormalIntDist>(new NormalIntDist(5, 1, 2, 10));
```
gonuke, line 341: indent or combine with previous line

gonuke (changes requested):
Two minor comments and one more substantive one (`SampleRequestSize()`)

src/toolkit/matl_buy_policy.cc
nuclearkatie: You know, as I'm thinking about this, it may make sense for all reasonable user use cases (like myself, encountering this while writing the tests). You're right that my original version is inconsistent in any case. Your version makes sense In the context of uniform distribution, which is required to have a max, but less so for normal, which is not. For a normal dist, the max defaults to the numeric limit. Therefore, a default normal dist, which only sets mean and stddev, will constantly be scaled against the largest value possible, resulting in a zero request size. 
Instead of adding fancy handling here, it may be simplest to just say that if 
```suggestion
  if (random_request_size_ > 1) {
    random_request_size_ = 1.0;
  }
```
aka the simplest artificial max

gonuke: Two things:

1) This will skew the sampling of the distribution.  What if someone has a Normal Distribution with a max of 1.1?  The sampling routine will reject/resample until it gets a value less than 1.1, but this will simple change the value rather than resample.  That's an inconsistent handling of your stochastic variable.

2) Therefore, it might make more sense to require all distributions used for sample size to have a max that is less than or equal to 1?  This could be tested when the policy is initialized and cause an error (or automatic correction) when violated.

src/toolkit/matl_buy_policy.cc
nuclearkatie: 👍 added check of dist max in init, throw ValueError if >1. Removed check in SampleRequestSize

src/random_number_generator.h
```diff
@@ -76,6 +86,88 @@ class RandomNumberGenerator {
...
+    boost::random::normal_distribution<> dist;
+    double min_;
+    double max_;
+  public:
+    NormalDoubleDist(double mean, double std_dev, double min=0, double max=1) : dist(mean, std_dev), min_(min), max_(max) {
+      if (min_ == max_) {
+        throw ValueError("Min and max cannot be equal for a normal distribution. Either use FixedDoubleDist or change the min/max.");
+      }
+      if (max_ < (mean - 3*std_dev) || min_ > (mean + 3*std_dev)) {
+        Warn<VALUE_WARNING>("Dist is sampling from a truncated normal more than 3 standard deviations from the mean. Drawing sampling may be inefficient");
```
gonuke, line 124:
```suggestion
        Warn<VALUE_WARNING>("Dist is sampling from a tail of a truncated normal more than 3 standard deviations from the mean. Drawing sampling may be inefficient");
```

gonuke (approved):

Issue #1589, opened 2023-09-15
any attempt to run cyclus via python ends in segfault


gonuke:
In the updated build under development in #1588, most of the tests pass.  The only one that does not attempts to call `cyclus` via the python `cyclus.main.main()` method in `test_main.py`.

I've replicated the behavior by just invoking:
```
from cyclus.main import main

main()
```
It appears to segfault as the program attempts to shutdown, having completed any actual computing.  It may be related to unloading the dynamically loaded modules? (hope that's not a red herring)

Tracking this down will requires a deep dive into a few layers of python/C++ interface. 

See also #1512, #1535, and #1543

bennibbelink:
Been playing around with this in gdb.  The segfaults occur when unloading the dynamically loaded modules, but based based on the backtrace I think they may be due to invalid references to cyclib in pymodule.pyx (cyclib doesn't exist in the module ?)

Will do some more digging... its taking time to understand how the python wrappers interact with the C++

gonuke:
Thanks for the update.  This sounds like something that is hard to find but might have a quick fix once you do (🤞)

bennibbelink:
UPDATE: I found **a** fix in the dummy project, I will try to implement it in cyclus now.  I don't think that it is the best practice but it is working in the simple case.

Essentially what I've done is build each cython module in `src/` as its own shared library, then in `__init__.py` I import it (instead of calling `cyclus.lib.py_import_call_init`, it is this broken functionality in `pyhooks.cc` that I think should be the "correct" fix)

bennibbelink:
I found a different fix in the dummy repo which I think is "more correct" (commit [0729d6f](https://github.com/bennibbelink/cython-playground/commit/0729d6f972a7a6e29e41c6c2f44f79bab9310c1a)).

This is what (I think) is happening in cyclus:
### When run using the cli
- `PyStart()` is called appropriately (in `pyhooks.cc`)
- The cython modules are appended to the initialization table as built-in python modules
- Python is initialized
- the cython modules are imported, and everything works well
### When run from Python
- `PyImportCallInit()` is called in `__init__.py` which initializes the cython modules
- `PyStart()` is called appropriately from within `libcyclus`
[...]

PR #1636, opened 2023-12-06
Cython3 upgrade

bennibbelink:
This PR removes the pin on cython in cyclus, and we now see the same behavior with cython3.0.6 as we did in 0.29.x.  Unfortunately it did not magically solve the issues discussed in #1589, so there is still some work to be done there

This has been branched off of `cmake-updates`, so I propose we resolve any needed revisions in #1632 and get that merged first to make this PR more readable


bennibbelink:
A ton of warnings still from the cython compiler.  Looking into resolutions...

CHANGELOG.rst
```diff
@@ -23,6 +23,8 @@ Since last release
 * Added code coverage reporting to GitHub workflows (#1616)
 * Adds active and dormant buying cycles in buy policy (#1596)
 * Add random number generator (Mersenne Twister 19937, from boost) and the ability to set the seed in the simulation control block (#1599, #1639)
+* Add random number generator (Mersenne Twister 19937, from boost) and the ability to set the seed in the simulation control block (#1599)
```
gonuke, line 26: this line suggests that we might need a rebase?

gonuke (changes requested):
Just one minor thing in the `CHANGELOG` makes me wonder if we need a rebase?

gonuke (approved):
Thanks for moving us into the Cython 3 era @bennibbelink

Issue #190, opened 2012-06-18
Allow non-standard module locations?


katyhuff:
This issue is a sub-part of issue #66.

Ensure that when a module is separated from the main repository, it can be configured (with CMake) to link to the core library for testing, test runs, etc. This will include a capability in the main function (App.cpp) and *LoadConstr.h headers to find the modules in places defined by the environment (perhaps a $CYCLUS_MODULE_PATH environment variable?)

Alternatively, we might require that a user install the modules in the installation path. This has some downsides for users without admin rights on their computers, but is easier to implement. 

Next Action : Determine whether tests need to be dynamically linked. If so, Issue #80 blocks this issue. Else, demonstrate running tests added with the macro. 

Next Action : Review

PR #1637, opened 2023-12-09
CI Workflow Refactoring

bennibbelink:
This branch uses a custom action (currently in my [repo](https://github.com/bennibbelink/cyclus-ci-action)) to build docker images using the Dockerfiles we have in our repos. The action is as follows:
- optionally download an docker image artifact (as a tarball) and load the image 
- checks out a specified repository and a specified ref (defaults to the commit that triggered the workflow)
- builds a docker image via the Dockerfile specified and optionally tags it according to an input argument and uploads a tarball artifact

I separated the Dockerfile into one that just builds the dependency images and one that builds cyclus.  I also moved code coverage into its own workflow since it is more in-depth than just building a Dockerfile

Changes to the old workflows:
### build_test.yml 
- uses the custom action to build cyclus from `cyclus.dockerfile`
- includes downstream testing of cycamore and cymetric, which is only triggered on PR and "soft" fails

### build_test_publish.yml
I couldn't find a great way to implement this without redundancy (or the multi-stage-build action we used before), so I'm open to ideas (I did find an [action](https://github.com/dawidd6/action-download-artifact) that might allow us to share artifacts between workflows but I fear that may end up adding much more complexity than its worth) 
- builds the dependency image from `deps.dockerfile` using the custom action and uploads a tarball artifact
- builds cyclus using the custom action (after loading the tarball) and push image if the workflow was triggered on push (to `main`)
- IF the cyclus build succeeds, push the dependency image to ghcr

### publish_release.yml
- uses the custom action to build and push a cyclus image tagged as the `github.ref_name`

I'm sure there must be some details I'm forgetting.  @gonuke what thoughts do you have?

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -72,7 +72,7 @@ jobs:
         with:
           repository: cyclus/cycamore
           ref: tag-build-arg
-          path: ${{ runner.temp }}/cycamore
+          path: ${{ github.workspace }}/cycamore
 
       - name: Build and Test Cycamore
         id: build-cycamore
@@ -80,8 +80,8 @@ jobs:
         continue-on-error: true
         uses: docker/build-push-action@v5
         with:
-          context: ${{ runner.temp }}/cycamore
-          file: ${{ runner.temp }}/cycamore/docker/Dockerfile
+          context: ${{ github.workspace }}/cycamore
+          file: ${{ github.workspace }}/cycamore/docker/Dockerfile
           cache-from: type=registry,ref=ghcr.io/cyclus/cycamore_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cycamore:ci-layer-cache
           cache-to: type=registry,ref=ghcr.io/cyclus/cycamore_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cycamore:ci-layer-cache,mode=max
           tags: ghcr.io/cyclus/cycamore_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cycamore:ci-image-cache
@@ -97,7 +97,7 @@ jobs:
         with:
           repository: cyclus/cymetric
           ref: tag-build-arg
-          path: ${{ runner.temp }}/cymetric
+          path: ${{ github.workspace }}/cymetric
 
       - name: Build and Test Cymetric
         if: ${{ github.event_name == 'pull_request' && steps.build-cycamore.outcome == 'success' }}
@@ -105,8 +105,8 @@ jobs:
         continue-on-error: true
         uses: docker/build-push-action@v5
         with:
-          context: ${{ runner.temp }}/cymetric
-          file: ${{ runner.temp }}/cymetric/docker/Dockerfile
+          context: ${{ github.workspace }}/cymetric
+          file: ${{ github.workspace }}/cymetric/docker/Dockerfile
           cache-from: type=registry,ref=ghcr.io/cyclus/cymetric_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cymetric:ci-layer-cache
           cache-to: type=registry,ref=ghcr.io/cyclus/cymetric_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cymetric:ci-layer-cache,mode=max
           tags: ghcr.io/cyclus/cymetric_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cymetric:ci-image-cache
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -71,7 +71,6 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: cyclus/cycamore
-          ref: tag-build-arg
           path: ${{ github.workspace }}/cycamore
 
       - name: Build and Test Cycamore
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -96,7 +96,6 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: cyclus/cymetric
-          ref: tag-build-arg
           path: ${{ github.workspace }}/cymetric
 
       - name: Build and Test Cymetric
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -72,7 +72,7 @@ jobs:
         with:
           repository: cyclus/cycamore
           ref: tag-build-arg
-          path: $HOME/cycamore
+          path: ${{ runner.temp }}/cycamore
 
       - name: Build and Test Cycamore
         id: build-cycamore
@@ -80,8 +80,8 @@ jobs:
         continue-on-error: true
         uses: docker/build-push-action@v5
         with:
-          context: $HOME/cycamore
-          file: $HOME/cycamore/docker/Dockerfile
+          context: ${{ runner.temp }}/cycamore
+          file: ${{ runner.temp }}/cycamore/docker/Dockerfile
           cache-from: type=registry,ref=ghcr.io/cyclus/cycamore_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cycamore:ci-layer-cache
           cache-to: type=registry,ref=ghcr.io/cyclus/cycamore_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cycamore:ci-layer-cache,mode=max
           tags: ghcr.io/cyclus/cycamore_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cycamore:ci-image-cache
@@ -97,7 +97,7 @@ jobs:
         with:
           repository: cyclus/cymetric
           ref: tag-build-arg
-          path: $HOME/cymetric
+          path: ${{ runner.temp }}/cymetric
 
       - name: Build and Test Cymetric
         if: ${{ github.event_name == 'pull_request' && steps.build-cycamore.outcome == 'success' }}
@@ -105,8 +105,8 @@ jobs:
         continue-on-error: true
         uses: docker/build-push-action@v5
         with:
-          context: $HOME/cymetric
-          file: $HOME/cymetric/docker/Dockerfile
+          context: ${{ runner.temp }}/cymetric
+          file: ${{ runner.temp }}/cymetric/docker/Dockerfile
           cache-from: type=registry,ref=ghcr.io/cyclus/cymetric_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cymetric:ci-layer-cache
           cache-to: type=registry,ref=ghcr.io/cyclus/cymetric_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cymetric:ci-layer-cache,mode=max
           tags: ghcr.io/cyclus/cymetric_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cymetric:ci-image-cache
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -9,7 +9,6 @@ on:
       - '.github/workflows/publish_release.yml'
       - '.github/workflows/changelog_test.yml'
       - '.github/workflows/code_coverage.yml'
-      - 'docker/**'
       - 'doc/**'
       - 'CHANGELOG.rst'
   push:
@@ -18,7 +17,6 @@ on:
       - '.github/workflows/publish_release.yml'
       - '.github/workflows/changelog_test.yml'
       - '.github/workflows/code_coverage.yml'
-      - 'docker/**'
       - 'doc/**'
       - 'CHANGELOG.rst'
 
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -57,8 +57,8 @@ jobs:
             pkg_mgr=${{ matrix.pkg_mgr }}
             ubuntu_version=${{ matrix.ubuntu_versions }}
             ci_run_id=${{ github.run_id }}
-          secret-envs: |
-            "GITHUB_TOKEN=${{ secrets.GITHUB_TOKEN }}"
+          secrets: |
+            repo_token=${{ secrets.GITHUB_TOKEN }}
         
   finish-coveralls:
     needs: build-and-test
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -55,21 +55,29 @@ jobs:
           cache-to: type=registry,ref=ghcr.io/cyclus/cyclus_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cyclus-coverage:ci-layer-cache,mode=max
           file: docker/Dockerfile
           target: coverage-report
-          outputs: type=local,dest=${{ github.workspace }}/coverage
+          outputs: type=local,dest=${{ github.workspace }}
           build-args: |
             pkg_mgr=${{ matrix.pkg_mgr }}
             ubuntu_version=${{ matrix.ubuntu_versions }}
 
       - name: debug
         run: |
-          ls ${{ github.workspace }} && ls ${{ github.workspace }}/coverage
+          pwd
+          echo break
+          ls 
+          echo break
+          ls ${{ github.workspace }}
+          echo break
+          ls ${HOME}
+          echo break 
+          ls ~
 
       - name: Coveralls
         uses: coverallsapp/github-action@v2
         with:
           flag-name: ${{ join(matrix.*, '-') }}
           parallel: true
-          file: ${{ github.workspace }}/coverage/cyclus_built/total_coverage.info
+          file: ${{ github.workspace }}/cyclus_built/total_coverage.info
           format: lcov
           compare-ref: main
 
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -22,12 +22,12 @@ jobs:
       fail-fast: false
       matrix:
         ubuntu_versions : [
-          20.04,
+          # 20.04,
           22.04,
         ]
         pkg_mgr : [
           apt,
-          conda,
+          # conda,
         ]
     
     name: Build, Test, Report Coverage
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -52,20 +52,29 @@ jobs:
           cache-from: type=registry,ref=ghcr.io/cyclus/cyclus_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cyclus-coverage:ci-layer-cache
           cache-to: type=registry,ref=ghcr.io/cyclus/cyclus_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cyclus-coverage:ci-layer-cache,mode=max
           file: docker/Dockerfile
-          target: cyclus-coverage
+          target: coverage-report
+          outputs: type=local,dest=${HOME}
           build-args: |
             pkg_mgr=${{ matrix.pkg_mgr }}
             ubuntu_version=${{ matrix.ubuntu_versions }}
-          secrets: |
-            "coveralls_flags=--parallel --job-flag ${{ matrix.ubuntu_versions }}-${{ matrix.pkg_mgr }} 
-            --build-number ${{ github.run_id }} -r ${{ secrets.GITHUB_TOKEN }}"
+
+      - name: debug
+        run: ls ${HOME}
+
+      - name: Coveralls
+        uses: coverallsapp/github-action@v2
+        with:
+          flag-name: ${{ join(matrix.*, '-') }}
+          parallel: true
+          file: ${HOME}/src_coverage.info
+          format: lcov
+          compare-ref: main
         
   finish-coveralls:
     needs: build-and-test
     runs-on: ubuntu-latest
     steps:
-    - name: Coveralls Done
-      run: |
-        curl -L https://coveralls.io/coveralls-linux.tar.gz | tar -xz
-        ./coveralls done --build-number ${{ github.run_id }} -r ${{ secrets.GITHUB_TOKEN }}
-        
+    - name: Coveralls Finished
+      uses: coverallsapp/github-action@v2
+      with:
+        parallel-finished: true
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -55,15 +55,13 @@ jobs:
           cache-to: type=registry,ref=ghcr.io/cyclus/cyclus_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cyclus-coverage:ci-layer-cache,mode=max
           file: docker/Dockerfile
           target: coverage-report
-          outputs: type=local,dest=${{ runner.temp }}
+          outputs: type=local,dest=.
           build-args: |
             pkg_mgr=${{ matrix.pkg_mgr }}
             ubuntu_version=${{ matrix.ubuntu_versions }}
 
       - name: debug
         run: |
-          cd ${{ runner.temp }}/cyclus_built
-          echo BREAK
           ls -l
           echo BREAK
           pwd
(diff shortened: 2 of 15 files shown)


bennibbelink:
Ah shoot looks like I need to fix the default tag. Will find fix soon

gonuke:
I think this PR is prematurely pushing updated docker images to GHCR as `latest`.  Most of our workflows have only done this when pushing to `develop` (aka merging a PR).  I haven't been able to confirm how that logic works here, but I do know that there is a sudden failure due to a lack of a libxml library in #1634, and I think I've traced it to this.

gonuke:
Thanks for building this system for us to review.  I think there's a lot to discuss and so many different factors to consider.

bennibbelink:
> I think this PR is prematurely pushing updated docker images to GHCR as `latest`. Most of our workflows have only done this when pushing to `main` (aka merging a PR). I haven't been able to confirm how that logic works here, but I do know that there is a sudden failure due to a lack of a libxml library in #1634, and I think I've traced it to this.

My fault, sorry about that.

What is the backtrace you've found?  I'm seeing the image sha used by failing workflow in #1634 as `4481a2f20cc68136982538b3a2a8a263dc818f282fd9374c2d135d8dada4855e`, which is consistent with the version of `latest` in ghcr.  But the latest successful build/push workflow associated with this PR pushes a 22.04-conda-deps image with sha `190ff26d1e527a489547b2dbf088760e9beab6662315340a9aa9490ae6ec9083`.

If you want to close this PR until we fix the problem so that CI works for other PRs, that's fine with me.

gonuke:
I'm not sure if we need to close it, or just prevent it from pushing with the `latest` tag?

gonuke:
I retagged an older version with the `latest` tag - hopefully that fixes it for now

gonuke (changes requested):
I look forward to chatting about some of these questions.

docker/cyclus.dockerfile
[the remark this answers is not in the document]
bennibbelink: Re-combined dockerfile and added a `docker-build-target` option to the custom action

.github/workflows/build_test_publish.yml
```diff
@@ -26,35 +26,93 @@ jobs:
...
+        ubuntu_versions : [
+          20.04,
+          22.04,
+        ]
+        pkg_mgr : [
+          apt,
+          conda,
+        ]
+
+    name: Push Dependency Image
```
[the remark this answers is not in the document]
bennibbelink: Yes, good catch.  Right now this workflow only runs on PR and a push to `main`, do you want to finally introduce a `develop` branch workflow like we've discussed in the past?

CHANGELOG.rst
```diff
@@ -27,7 +27,7 @@ Since last release
 **Changed:**
 
 * Moved to unified CHANGELOG Entry and check them with GithubAction (#1571)
-* Major update and modernization of build (#1587, #1632)
+* Major update and modernization of build (#1587, #1632, #1640)
```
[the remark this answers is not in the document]
bennibbelink: Hmm to be honest I'm not sure, maybe a remnant of a rebase(?)

Regardless... it is removed now

bennibbelink:
I'm not convinced that my manual caching between jobs is working correctly, I need to dig into this more.

Also... just stumbled across [this](https://docs.docker.com/build/cache/backends/gha/#scope) which might allow us to simplify the custom action

bennibbelink:
Still working through this downstream testing PR, but it looks like #1634 might've broken Cycamore @gonuke @nuclearkatie

nuclearkatie:
Yes, it did, see [cycamore#569](https://github.com/cyclus/cycamore/issues/569). #1634 changed the behavior of material buy policy, including the `init` process so cycamore also has to change.

I have a PR open that will fix it, [cycamore#568](https://github.com/cyclus/cycamore/pull/568). Also, there's a stray unrelated (sink, vs that PR is about storage) failing test that I have [cycamore#570](https://github.com/cyclus/cycamore/pull/570) open for

bennibbelink:
Thanks or the explanation @nuclearkatie

This was helpful in demonstrating this PRs capabilities (or lack thereof).  Thanks to the failing cycamore build we now have an example of failing downstream tests.  I fixed the workflow so that downstream failures don't inhibit us from merging a Cyclus PR.  Unfortunately, the failing downstream tests dont present themselves in this PR discussion in any way, you have to view the workflow [summary](https://github.com/cyclus/cyclus/actions/runs/7634020447?pr=1637) to see the warnings. 

@gonuke do you have any thoughts about this? I wish it were more convenient to return warnings instead of success/failure but Github doesn't seem to have that functionality in place natively

gonuke:
Hmmm.... that's annoying.  What about something like this: https://github.com/marketplace/actions/add-pr-comment

bennibbelink:
Please see cycamore PR [#573](https://github.com/cyclus/cycamore/pull/573) for notes on some of the changes that I've made (also involved is cymetric PR [#190](https://github.com/cyclus/cymetric/pull/190)).  I am finally confident that the workflows are caching layers and images appropriately, and downstream testing is using correct images to build from.

We are still seeing cycamore build failures but this is expected until cycamore PR [#568](https://github.com/cyclus/cycamore/pull/568) is merged (and it currently gives us a good example of what happens when downstream testing fails).  

Also, since this PR now utilizes build arguments in the cycamore and cymetric Dockerfiles, those two PRs must be merged first before this can be merged.  Currently the workflow checks out the `tag-build-arg` ref instead of `main` to allow it to work.  This should be fixed before this PR gets approved

bennibbelink:
Also... thanks to using the registry as a cache backend, we no longer need to maintain <pkg_mgr>-deps images to speed up our workflows.  Thus, I renamed the `build_test_publish.yml` workflow to `publish_latest.yml` in all 3 repos.  It is nearly identical to the `publish_release.yml` workflow except - `publish_latest.yml` publishes the `latest` image on a push to `main` and `publish_release.yml` publishes the `stable` image on a push to `main` with a tag (I think this will work for a release, we can test it out prior to a release by creating an arbitrary tag.  There are release triggers but the different types seem somewhat convoluted and this solution feels simpler yet still correct).

gonuke:
Did you meant to put this back in draft @bennibbelink ?

bennibbelink:
The downstream docker builds are failing when they attempt to push images to ghcr.  In order for this to work @gonuke you might need to poke around in the package settings.  By default a repository's packages inherit permissions from the repository - thus I don't think the GITHUB_TOKEN credential in a cyclus workflow holds permission to push to the cycamore package registry.  [This page](https://docs.github.com/en/packages/learn-github-packages/configuring-a-packages-access-control-and-visibility) might be helpful

That being said... feel free to review/discuss the PR(s) before changing any settings

gonuke:
I added permission for the cyclus repo to write to the packages initially created by the cycamore repo (and thus inheriting their permisison) - and it seems to have worked for cycamore.  I've just added it for cymetric as well.

What matrix do you think is necessary?  Which repo's need to be able to build which packages?  Clearly, cyclus repo needs to write the cycamore and cymetric packages.  Do we need cycamore to be able to write cyclus packages? cymetric packages?

bennibbelink:
Cycamore needs to be able to write cymetric packages so that CI in cycamore works.  I don't think we should go the other way (i.e. cycamore writing cyclus packages). 

In the past we may have discussed testing a new cycamore feature agains stable cyclus and latest cyclus, but I think we should force cycamore contributors to introduce features using `cyclus:latest` (i.e. `main` branch).  In theory any commits to cyclus `main` branch have been vetted and will be included in the next `stable` release anyway, so a contribution to cycamore's `main` branch should have to align with the latest cyclus library.  I think this is the simplest way to ensure that the respective `main` branches stay mostly in-sync, and releases act as checkpoints where they are confirmed to be in-sync.

As for the case where a new cyclus feature breaks cycamore:
1. there should be a corresponding cycamore PR that brings cycamore up to date with cyclus (ideally very shortly after the cyclus PR)
2. we should encourage users to use the ghcr images, since the `latest` images will always be from a point in time where the `main` branches were in-sync (and... maybe we should have the cyclus `publish_latest.yml` build and push `cycamore:latest` too so that the cyclus library is always up-to-date in the `cycamore:latest` image - assuming its a non-breaking change)

Let me know your thoughts, I may be forgetting some cases

.github/workflows/build_test.yml
```diff
@@ -30,77 +33,97 @@ jobs:
...
+        continue-on-error: true
+        uses: docker/build-push-action@v5
         with:
-          name: cyclus-coverage-report-${{ matrix.ubuntu_versions }}-${{ matrix.pkg_mgr }}
-          path: ${{ env.artifactPath }}
+          context: $HOME/cycamore
+          file: $HOME/cycamore/docker/Dockerfile
+          cache-from: type=registry,ref=ghcr.io/cyclus/cycamore_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cycamore:ci-layer-cache
+          cache-to: type=registry,ref=ghcr.io/cyclus/cycamore_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cycamore:ci-layer-cache,mode=max
+          tags: ghcr.io/cyclus/cycamore_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cycamore:ci-image-cache
```
gonuke, line 85: It's not clear to me what's in the cache at this point?  Is the most recent successful cyclus build still in the cache?

.github/workflows/code_coverage.yml
```diff
@@ -0,0 +1,98 @@
...
+          20.04,
+          22.04,
+        ]
+        pkg_mgr : [
+          apt,
+          conda,
+        ]
+    
+    container:
+      image: ghcr.io/cyclus/cyclus_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/${{ matrix.pkg_mgr }}-deps
```
gonuke, line 30: do we still maintain a `-deps` image with this workflow?

gonuke (changes requested):
Thanks @bennibbelink - here are a few comments as we approach the switch this plan

gonuke:
I feel as though I will ultimately want some diagram or listing of which versions of which images with which tags are in all the caching locations at which points in time.

I see the following "caching" locations:
* GHCR
* GH action cache
* Docker local image cache

and I see the following images:
* cyclus/cyclus_*
* cyclus/cycamore_*
* cyclus/cymetric_*

and I see the following tags:
* ci-layer-cache
* ci-image-cache
* latest

Also - how will these interact if/when there simultaneous unrelated PR's pushing/pulling/caching against all of these?

.github/workflows/publish_latest.yml
```diff
@@ -0,0 +1,66 @@
...
+  push:
+    branches:
+      - main
+
+jobs:
+  build-dependency-img:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
```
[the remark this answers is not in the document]
bennibbelink: Pretty sure we need them.  Tried [removing them](https://github.com/cyclus/cyclus/actions/runs/7802117776) and got 403 forbidden

.github/workflows/build_test.yml
```diff
@@ -19,9 +19,12 @@ on:
...
 
 jobs:
-  build-and-test:
+  build-test-cyclus:
     runs-on: ubuntu-latest
-
+    permissions:
+      contents: read
+      packages: write
+      pull-requests: write
```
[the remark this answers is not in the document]
bennibbelink: See above.  `pull-requests: write` is necessary for the bot to comment on our PR at the end of the workflow

.github/workflows/publish_latest.yml
```diff
@@ -0,0 +1,66 @@
+name: Publish Latest Cyclus
+
+on:
+  # allows us to run workflows manually
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '.github/workflows/build_test_publish.yml'
```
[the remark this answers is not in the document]
bennibbelink: Fixed these typos and updated the ignores on all workflows to reflect new filenames

docker/Dockerfile
```diff
@@ -11,7 +10,8 @@ RUN apt update --fix-missing
 
 FROM common-base as apt-deps
 
-RUN apt install -y \
+RUN apt update --fix-missing && \
```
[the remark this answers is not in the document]
bennibbelink: Good catch. That was a remnant of experimentation

.github/workflows/code_coverage.yml
bennibbelink: No we do not.  Working on a solution now

bennibbelink:
> I feel as though I will ultimately want some diagram or listing of which versions of which images with which tags are in all the caching locations at which points in time.
> 

I will make something for our docs

> I see the following "caching" locations:
> 
> * GHCR
> * GH action cache
> * Docker local image cache

We only technically cache layers using GHCR as a [backend](https://docs.docker.com/build/cache/backends/registry/).  All "images" tagged with `:ci-layer-cache` act as a place to store layers from the build.  I put "images" in quotes because they cannot actually be pulled and run as an image, you can think of them more as a repository for layers.  This is acting in place of the `<pkg_mgr>-deps` images that we pushed before, except now instead of manually specifying a deps image to build from, our builds will automatically reuse whatever layers they can find.  This appears to be working well... if you inspect the workflow logs you can see which layers are being found in the cache.

GH action cache isn't feasible for the amount of layers we are producing given the storage limit of 10GB.

> and I see the following images:
> 
> * cyclus/cyclus_*
> * cyclus/cycamore_*
> * cyclus/cymetric_*
> 
> and I see the following tags:
> 
> * ci-layer-cache
> * ci-image-cache
> * latest

Each respective image has it's own tag set aside for a layer cache (described above) and image cache.  The `ci-image-cache` tags are used to store images that are built from CI.  It's not really a "cache" per se but it is a mechanism to pass images downstream for cycamore to build from.  Since all downstream testing is performed in one job we don't need to push these images, we could just load them onto the runner machine and perform downstream testing as usual.  However, by pushing them to the registry we can pull images locally and inspect them if we feel the need to.

`latest` hasn't changed, it is still the tag used on a commit to `main`

> Also - how will these interact if/when there simultaneous unrelated PR's pushing/pulling/caching against all of these?

The `:ci-layer-cache` tag will be overwritten by unrelated PR's caching against it. However, since its main purpose is to store the layers up through the `<pkg-mgr>-deps` stage this shouldn't be an issue unless one of the simultaneous PRs is changing the dependencies in the Dockerfile.  If that will happen frequently we can reconsider this strategy.  In almost all cases the cache is invalidated by `COPY . /cyclus` since some files in the repo have changed, so we shouldn't consider the following layers being overwritten as a major fault of this strategy.

The `:ci-layer-cache` tag will be overwritten by unrelated PR's as well but it is no issue at all.  When performing downstream testing we reference the cyclus image by digest (`:ci-image-cache@<sha>`) which we get from the previous steps in the workflow.  I was running into a lot of issues trying to get cycamore to use the exact image from an upstream cyclus build, so I am fairly confident that specifying by digest would be necessary even if we kept the images local to the github runner.

gonuke:
Note that merging #1666 has introduced a little merge conflict

gonuke:
Is this drop in coverage real? Do we know why?

bennibbelink:
Ok I'm pretty sure I solved the issue.  This is what I think was happening(?):
- cyclus was being built inside of the docker image and a coverage report was generated
- I was copying the coverage report back out to the github runner machine so we can utilize the coveralls github action
- The cyclus repo on the github runner didn't have any of the files generated during build time. Without that context coveralls wasn't processing the entire coverage report(?)

This explains why I was seeing consistent coverage values in the docker image but they weren't being pushed to coveralls.

I fixed the issue by copying the whole project back out to the github runner in the `cyclus_built` subdirectory.

---------------------------------------------------------------------
Another thing to note: Previously the workflow just reported on the `src/` subdirectory of cyclus, but I changed it to report on the whole project.  That's why coverage dropped to ~45%.  If you go to coveralls and look at the [build](https://coveralls.io/builds/65596929) you can see that the `src/` directory is still up around ~48-49% which is consistent with previous builds.

bennibbelink:
Here is a basic diagram of what's happening in the `build_test.yml`:
```mermaid
sequenceDiagram
	
	participant G as GitHub Container Registry
	participant R as Github Runner
	
	G->>R: Fetch cyclus:ci-layer-cache
	Note right of R: Build cyclus and tag as cyclus:ci-image-cache
	R->>G: Export build cache to cyclus:ci-layer-cache
	R->>G: Push cyclus:ci-image-cache to GHCR
	Note right of R: End of cyclus build

	G->>R: Fetch cycamore:ci-layer-cache
	Note right of R: Build cycamore and tag as cycamore:ci-image-cache, using <br/>cyclus:ci-image-cache@<cyclus-sha> as the base image
	R->>G: Export build cache to cycamore:ci-layer-cache
	R->>G: Push cycamore:ci-image-cache to GHCR
	Note right of R: End of cyamore build

	G->>R: Fetch cymetric:ci-layer-cache
	Note right of R: Build cymetric and tag as cymetric:ci-image-cache, using <br/>cycamore:ci-image-cache@<cycamore-sha> as the base image
	R->>G: Export build cache to cymetric:ci-layer-cache
	R->>G: Push cymetric:ci-image-cache to GHCR
	Note right of R: End of cymetric build

	Note right of R: Comment build statuses on pull request
```

bennibbelink:
With that - marking this as ready for review.  Some things that I changed since we met on Wednesday:

- I changed some triggers... mostly to do with the Dockerfile.  I may be forgetting something but I think a big reason we had special cases for Dockerfile changes were because we were pushing updated dependency images.  Now the workflows are triggered as follows:
  -  `publish_latest.yml` runs on a PR that changes `publish_latest.yml` or on a push to `main`
  - `build_test.yml` runs on a PR that changes pretty much anything relevant, i.e. everything except the other workflows, docs, and changelog. 
  - the others are pretty straightforward
- I modified `publish_latest.yml` to build and push downstream images.  This way if there is a cyclus update the downstream `latest` images will get updated with the latest cyclus core library.  It will tag them as `latest` according to the usual conditions and as `ci-image-cache` otherwise (i.e. when someone has a PR that modifies the workflow).  The downstream image builds fail softly


**Note: I marked this ready for review but as we discussed we must merge cymetric [#573](https://github.com/cyclus/cycamore/pull/573) first, update cycamore PR [#190](https://github.com/cyclus/cymetric/pull/190) and merge, then update this PR and merge**

.github/workflows/build_test.yml
```diff
@@ -30,77 +35,97 @@ jobs:
...
 
-      - name: Cyclus Python Tests
-        run: |
-          cd tests && python -m pytest
+      - name: Checkout Cycamore
+        if: ${{ github.event_name == 'pull_request' && steps.build-cyclus.outcome == 'success' }}
+        uses: actions/checkout@v4
+        with:
+          repository: cyclus/cycamore
+          ref: tag-build-arg
```
gonuke, line 74: cleanup after merge on cycamore

.github/workflows/build_test.yml
```diff
@@ -30,77 +35,97 @@ jobs:
...
+        if: ${{ github.event_name == 'pull_request' && steps.build-cycamore.outcome == 'success' }}
+        uses: actions/checkout@v4
         with:
-          flag-name: ${{ join(matrix.*, '-') }}
-          parallel: true
-          file: ${HOME}/src_coverage.info
-          format: lcov
-          compare-ref: main
+          repository: cyclus/cymetric
+          ref: tag-build-arg
```
gonuke, line 99: cleanup on cymetric merge

.github/workflows/code_coverage.yml
```diff
@@ -0,0 +1,82 @@
...
+      fail-fast: false
+      matrix:
+        ubuntu_versions : [
+          20.04,
+          22.04,
+        ]
+        pkg_mgr : [
+          apt,
+          conda,
+        ]
```
gonuke, line 31: Do you think we need to do coverage across all these variations?  What do we miss if we don't?

.github/workflows/publish_latest.yml
```diff
@@ -0,0 +1,110 @@
...
+          tags: ghcr.io/cyclus/cyclus_${{ matrix.ubuntu_versions }}_${{ matrix.pkg_mgr }}/cyclus:${{ env.tag }}
+          build-args: |
+            pkg_mgr=${{ matrix.pkg_mgr }}
+            ubuntu_version=${{ matrix.ubuntu_versions }}
+
+      - name: Checkout Cycamore
+        uses: actions/checkout@v4
+        with:
+          repository: cyclus/cycamore
+          ref: tag-build-arg
```
gonuke, line 71: cleanup after cycamore merge

.github/workflows/publish_latest.yml
```diff
@@ -0,0 +1,110 @@
...
+          build-args: |
+            pkg_mgr=${{ matrix.pkg_mgr }}
+            ubuntu_version=${{ matrix.ubuntu_versions }}
+            cyclus_tag=${{ env.tag }}@${{ steps.build-cyclus.outputs.digest }}
+
+      - name: Checkout Cymetric
+        uses: actions/checkout@v4
+        with:
+          repository: cyclus/cymetric
+          ref: tag-build-arg
```
gonuke, line 94: cleanup after cymetric merge

.github/workflows/publish_release.yml
```diff
@@ -1,12 +1,18 @@
...
+name: Publish Stable Cyclus
 
 on:
-  release:
-    types: [published, created, edited]
+  push:
+    branches:
+      - main
+    tags:
+      - '*'
```
gonuke, line 8: repeating comment: `release = released`?

docker/Dockerfile
```diff
@@ -112,6 +111,33 @@ ENV CC /opt/conda/bin/x86_64-conda_cos6-linux-gnu-gcc
...
+RUN mamba install -y lcov curl
+ENV GCOV /opt/conda/bin/x86_64-conda-linux-gnu-gcov
+
+FROM ${pkg_mgr}-coverage-deps as cyclus-coverage
+ARG make_cores=2
+
+COPY . /cyclus
+WORKDIR /cyclus
+
+RUN python install.py -j 2 --build-type=Release --core-version 999999.999999 --allow-milps --code_coverage
```
gonuke, line 128: Is there a reason not to parallelize this further?
```suggestion
RUN python install.py -j ${make_cores} --build-type=Release --core-version 999999.999999 --allow-milps --code_coverage
```

gonuke (changes requested):
Some repeated some new comments... mostly minor

.github/workflows/code_coverage.yml
bennibbelink: I guess we probably don't.  We ignore external dependencies so the coverage should be the same across the build matrix.  They mostly are... [except](https://coveralls.io/builds/65607725) for the fact that the `20.04-apt` is covering 264 files compared to 266 files in the other 3 builds so it is slightly different 🧐.

The only downside I see to performing coverage on the entire matrix is resource usage, but we don't have any limits since our repos are public.  Do you think we should slim down the testing matrix for this workflow?

docker/Dockerfile
bennibbelink: Nope just oversight on my part.  Updated Dockerfile accordingly

.github/workflows/code_coverage.yml
gonuke: It could be confusing to get 4 coverage results and then decide which matter?

bennibbelink: Hmm... I may have been misinterpreting what Coveralls meant by a [parallel build](https://docs.coveralls.io/parallel-builds).  I thought it was intended for a testing matrix like ours, but it appears that it might be intended for multiple stages of a build rather than multiple versions of the same build.  That, along with the results I'm seeing in the coverage reports (ach of the 4 builds is around 32% but the collective coverage is 37% ???) leads me to agree with you.  Which build do you want to keep?

.github/workflows/publish_release.yml
bennibbelink: Just realized we need to change the image tag since `github.ref_name` doesnt exist in the context of a release https://github.com/orgs/community/discussions/64528.  I will make PRs to update the other two repos

bennibbelink: In opening the other PRs I found a flaw(?) in the triggers for `publish_release.yml`.  Should we add a trigger so that `publish_release.yml` runs on a PR that edits `publish_release.yml`?  (and tags images accordingly)

.github/workflows/code_coverage.yml
gonuke: I don't think it matters which we keep - flip a coin or two?

gonuke: I see you did already

gonuke (approved):

gonuke:
Merging because the code coverage reduction is an artifact of increased testing scope

PR #1640, opened 2023-12-23
force libxmlpp 4.0 in conda

gonuke:
Some changes in the conda dependency tree have resulted in the `libxmlpp` resolving as version 2.40.x and then having an incompatibility with the newest `glibmm`.  This fixes it in local testing.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,7 +27,7 @@ Since last release
 **Changed:**
 
 * Moved to unified CHANGELOG Entry and check them with GithubAction (#1571)
-* Major update and modernization of build (#1587, #1632, #)
+* Major update and modernization of build (#1587, #1632, #1640)
 * Changed Json formatting for compatibility with current python standards (#1587)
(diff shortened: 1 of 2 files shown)


gonuke:
@bennibbelink - I think this will fix the issue that we're seeing in testing.  It is probably due to changes upstream in the conda packages, and this fix should help.  It is likely that if this fix were appleid to your PR #1637 it would mean that PR was less intrusive at this time, as well

bennibbelink:
Thanks for tracking this down @gonuke.  Can you confirm that we are okay "pinning" our build to libxmlpp-4.0? (I would assume yes because libxmlpp-4.0 is it's own package and not a package version number, but want to double check)

bennibbelink:
I have rebased #1637 on top of these commits so it is less problematic for the time being

gonuke:
Yes - it's effectively a "pinning" but without really pinning

bennibbelink (approved):
Sounds good to me, thanks @gonuke!

PR #1646, opened 2024-01-24
TotalInvTracker to be an inventory check on whole agents

### LINKED ISSUE #1647: Design (R,Q) inventory policy

### LINKED ISSUE #1648: Implement (R,Q) inventory policy in buypol

### LINKED ISSUE #1649: Test (R,Q) inventory policy

### LINKED ISSUE #1650: Design (s,S) inventory policy

### LINKED ISSUE #1651: Implement (s,S) inventory policy in buypol

### LINKED ISSUE #1652: Test (s,S) inventory policy

nuclearkatie:
`TotalInvTracker` keeps track of multiple ResBufs. 

- Has its own maximum inventory. 
- Checking for the space available in a `TotalInvTracker` returns the min of its own inventory and the space remaining in all ResBufs
-  Its perfectly acceptable to build a tracker with only one ResBuf
- `MatlBuyPolicy` modified to require a `TotalInvTracker`. This allows there to be a total agent cap in addition to any restrictions placed on the receiving buffer connected to the buy policy
    - Importantly, this will allow for cycamore:Storage to be fixed (see [cycamore#554](https://github.com/cyclus/cycamore/issues/554)). Currently it abuses its receiver ResBuf `inventory` by constantly modifying its total allowable inventory. The new process will be to set a receiving ResBuf with a max inventory (and throughput, etc) and then set up a `TotalInvTracker` that is aware of all ResBufs in the archetype and has its own maximum inventory. This way, the capacity of `inventory` will never have to change, but material present in other ResBufs will restrict the demand against the total allowable inventory

(R,Q): Closes #1647, closes #1648, closes #1649
(s:S): closes #1650, closes #1651, closes  #1652


abachma2 (commented):
This looks pretty good to me. I have a few small questions for you. Would any of this information show up in the output database anywhere? Or is this more for internal cyclus record keeping? I don't think I saw anything about writing to the database.

tests/toolkit/matl_buy_policy_tests.cc
```diff
@@ -42,71 +42,125 @@ class MatlBuyPolicyTests: public ::testing::Test {
 };
 
 TEST_F(MatlBuyPolicyTests, Init) {
-  double cap = 5;
+  double cap = 10;
```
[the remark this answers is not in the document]
nuclearkatie: No, not anymore. I was previously adding all the sS and RQ tests into Init, but realized that 1. re-initializing a buy policy with a different buying strategy leaves the old parameters hanging around and probably isn't good practice so I separated the tests and 2. refactored the test after making that change so the cap isn't really important anymore. I'll roll it back since there's no reason for it any more

abachma2:
I'm not sure if I would necessarily need something to be in the output, I feel like `explicit_inventory` would capture this some since the total inventory is limited. I was mostly doing a sanity check based on your code.

gonuke:
I don't think anything about the current Storage class shows up in the database either, other than the total storage size as an agent state variable

src/toolkit/matl_buy_policy.cc
```diff
@@ -43,18 +45,44 @@ void MatlBuyPolicy::set_manager(Agent* m) {
   }
 }
 
+void MatlBuyPolicy::set_total_inv_tracker(TotalInvTracker* t) {
+  if (t == NULL){
```
gonuke, line 49: Should `t` have a default value of `NULL` in the header?  (I notice it doesn't)

src/toolkit/matl_buy_policy.cc
```diff
@@ -43,18 +45,44 @@ void MatlBuyPolicy::set_manager(Agent* m) {
...
+    // maximum amount that an RQ policy could achieve is req_at + fill
+    set_fill_to(req_at + fill);
+  }
+  else {
+    throw ValueError("Invalid inventory policy");
+  }
+}
+
 void MatlBuyPolicy::set_fill_to(double x) {
-  if (x > 1)
```
gonuke, line 47: This seems like an important difference - can you explain why we no longer want to constrain this to [0,1]?

src/toolkit/matl_buy_policy.cc
```diff
@@ -43,18 +45,44 @@ void MatlBuyPolicy::set_manager(Agent* m) {
...
-    x /= buf_->capacity();
-  assert(x > 0 && x <= 1.);
+  assert(x > 0);
   fill_to_ = x;
 }
 
-void MatlBuyPolicy::set_req_when_under(double x) {
```
gonuke, line 53: Same comment as above about why we no longer limit to [0,1].

src/toolkit/matl_buy_policy.h
```diff
@@ -191,17 +207,22 @@ class MatlBuyPolicy : public Trader {
...
   void set_quantize(double x);
   void set_throughput(double x);
   void init_active_dormant();
 
   ResBuf<Material>* buf_;
-  std::string name_;
-  double fill_to_, req_when_under_, quantize_, throughput_;
+  TotalInvTracker* buf_tracker_;
+  std::string name_, inv_policy;
+  double fill_to_, req_at_, quantize_, throughput_, ccap_, cycle_total_inv_;
```
gonuke, line 225: Is there a substantive difference between `req_at_` and `req_when_under_`?

gonuke: Also, can we use a longer name for `ccap_` - perhaps `cumulative_capacity_`? Long variable names can really help readability, esp. for new developers and me in 6 months.

src/toolkit/matl_buy_policy.cc
```diff
@@ -43,18 +45,44 @@ void MatlBuyPolicy::set_manager(Agent* m) {
...
+    buf_tracker_ = t;
+  }
+}
+
+void MatlBuyPolicy::set_inv_policy(std::string inv_policy, double fill, double req_at) {
+  set_req_at(req_at);
+  if ((inv_policy == "sS") || (inv_policy == "Ss")) {
+    set_fill_to(fill);
+  }
+  else if ((inv_policy == "RQ") || (inv_policy == "QR")) {
```
gonuke, line 63: Do you want to handle lower case?
```suggestion
  std::transform(inv_policy.begin(), inv_policy.end(), inv_policy.begin(), ::tolower);
  if ((inv_policy == "ss")) {
    set_fill_to(fill);
  }
  else if ((inv_policy == "rq") || (inv_policy == "qr")) {
```

src/toolkit/matl_buy_policy.cc
```diff
@@ -186,11 +253,17 @@ void MatlBuyPolicy::Stop() {
 std::set<RequestPortfolio<Material>::Ptr> MatlBuyPolicy::GetMatlRequests() {
   rsrc_commods_.clear();
   std::set<RequestPortfolio<Material>::Ptr> ports;
-  bool make_req = buf_->quantity() < req_when_under_ * buf_->capacity();
+
+  // (s,S)/(R,Q) inventory policy handling
+  if (!MakeReq()) {
+    return ports;
+  }
```
gonuke, line 260: Is there a specific code logic reason this moved up here?

src/toolkit/matl_buy_policy.cc
```diff
@@ -206,11 +279,18 @@ std::set<RequestPortfolio<Material>::Ptr> MatlBuyPolicy::GetMatlRequests() {
...
+      SetNextDormantTime();
+      LGH(INFO4) << "end of dormant period, next active time end: " << next_active_end_ << ", and next dormant time end: " << next_dormant_end_ << std::endl;
+    }
+    else {next_dormant_end_ = -1;}
   }
 
-  if (!make_req || amt < eps())
+  if ((ccap_ != -1) && ((cycle_total_inv_ + amt) > ccap_)) {
+    amt = ccap_ - cycle_total_inv_;
+  }
```
gonuke, line 291:
```suggestion
  if ( ccap_ != -1 ) {
    amt = std::min(amt, ccap_ - cycle_total_inv_);
  }
```

src/toolkit/matl_buy_policy.cc
```diff
@@ -247,6 +327,17 @@ void MatlBuyPolicy::AcceptMatlTrades(
     LGH(INFO3) << "got " << it->second->quantity() << " kg of "
                << it->first.request->commodity()  << std::endl;
     buf_->Push(it->second);
+    // ccap handling
+    if (ccap_ > 0) {
```
gonuke, line 331: Above you used the condition `ccap_ != -1`. Do these mean the same thing? If so, can we pick one for clarity/readability?

gonuke: Of we can even make a simple inline function:
`include bool use_cumulative_capacity() { return ccap_ > 0; };`

src/toolkit/matl_buy_policy.cc
```diff
@@ -186,11 +253,17 @@ void MatlBuyPolicy::Stop() {
...
+  if (!MakeReq()) {
+    return ports;
+  }
+
   double amt;
 
   int current_time_ = manager()->context()->time();
 
+  // period inventory handling
   if (never_dormant() || current_time_ < next_active_end_) {
```
gonuke, line 267: As all this logic has become more complicated, I'm thinking about a system that separates out the different purposes into different chunks or code:
* determine if the inventory policy requires/allows a request to be made
* determine if the time requires/allows a request to be made
* update the next active/dormant times

It may cause some redundancy in the checks that occur, but it will be easier to know what is happening everyhere.

gonuke (changes requested):
This has a lot of interesting things going on @nuclearkatie - thanks for this effort.

Some of these comments connect to larger questions about the code design of the `MatBuyPolicy` as it gets more complex.  In particular, we should discuss the different initializations that are possible and whether some of them are replaced by explicit calls to setup policy parameters outside of an `Init()` call?

This review covers much of the changes, but I'm holding off on reviewing tests until we resolve some of these other questions.

cli/cycpp.py
```diff
@@ -102,11 +102,14 @@
            ('{0}::toolkit::ResBuf'.format(CYCNS), CYCNS + '::Product'),
            ('{0}::toolkit::ResBuf'.format(CYCNS), CYCNS + '::Material'),
            '{0}::toolkit::ResMap'.format(CYCNS),
+           '{0}::toolkit::TotalInvTracker'.format(CYCNS),
```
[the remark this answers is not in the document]
nuclearkatie: Uh oh... I honestly just added this stuff one by one in cycpp until the errors stopped coming when I tried to build cycamore. I don't really understand this type stuff myself

src/toolkit/matl_buy_policy.cc
nuclearkatie: Sure. `fill_to` and `req_at`, as they were originally imagined, were set as a fraction of the capacity. Through re-factoring the code and using them in my own demonstrations, I found it very annoying to work with these as fractions and felt that they would be better off as actual mass values

src/toolkit/matl_buy_policy.cc
nuclearkatie: I personally like them *as a user* a lot more as values rather than fractions. Maybe an argument can be made that it makes sense to set them up as [0,1] and then storage (which the user is engaging when writing input files) passes them as values that get scaled back to fractions.

But also, when you're setting req_under and fill_to as values, then you don't really need to set the buffer capacity, because those values alone dictate how the policy acts. Versus when you're using them as fractions, you have to set three values (and in my experience, that means setting the capacity and then pulling up the calculator to figure out what req_under and fill_to should be)

src/toolkit/matl_buy_policy.cc
nuclearkatie: Yeah, you don't need to calculate anything below if MakeReq is false, so it saves executing useless code if that's the case

src/toolkit/matl_buy_policy.h
```diff
@@ -102,17 +102,27 @@ class MatlBuyPolicy : public Trader {
   /// Note that active and dormant periods are note currently compatible with
   /// (s, S) inventory management
   /// @{
-  MatlBuyPolicy& Init(Agent* manager, ResBuf<Material>* buf, std::string name);
   MatlBuyPolicy& Init(Agent* manager, ResBuf<Material>* buf, std::string name,
-                      double throughput, 
+                      TotalInvTracker* buf_tracker);
```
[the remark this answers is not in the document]
nuclearkatie: No, and the reason I did it this way is that I don't really want to replicate the solution where sometimes the agent starts/manages the tracker, but other times they don't and then the buy policy has to and then shared pointers have to be used everywhere to paper over this

src/toolkit/matl_buy_policy.h
```diff
@@ -144,12 +154,18 @@ class MatlBuyPolicy : public Trader {
 
   /// the total amount available to request
   inline double TotalAvailable() const {
-    return std::min(throughput_,
-                   (fill_to_ * buf_->capacity()) - buf_->quantity());
+    return std::min({throughput_,
+                    fill_to_ - buf_->quantity(),
```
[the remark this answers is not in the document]
nuclearkatie: Because fill_to is now a value and not a fraction

src/toolkit/matl_buy_policy.h
```diff
@@ -144,12 +154,18 @@ class MatlBuyPolicy : public Trader {
...
+                    buf_->space(),
+                    buf_tracker_->space()});
```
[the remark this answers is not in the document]
nuclearkatie: Yes, do you think I should do it a different way?

src/toolkit/matl_buy_policy.h
```diff
@@ -144,12 +154,18 @@ class MatlBuyPolicy : public Trader {
...
   }
 
+  /// whether a request can be made
+  inline bool MakeReq() const { return buf_tracker_->quantity() <= req_at_; }
+
   /// whether trades will be denoted as exclusive or not
-  inline bool Excl() const { return quantize_ > 0; }
+  inline bool Excl() const { 
+    return quantize_ > 0; }
```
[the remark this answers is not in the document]
nuclearkatie: not sure, probably not intentional. I'll switch back

src/toolkit/matl_buy_policy.h
nuclearkatie: No, I should probably change everything to `req_at` because that's the more precise term (the other one is the previously-existing term)

src/toolkit/matl_buy_policy.cc
nuclearkatie: Is this more clear? 
https://github.com/cyclus/cyclus/blob/c4ac399d2236f34cf79848a180c626d2d82fd88a/src/toolkit/matl_buy_policy.cc#L265-L307

src/toolkit/matl_buy_policy.h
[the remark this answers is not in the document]
gonuke: OK - I have some thoughts on this for a follow-on PR, perhaps.

src/toolkit/matl_buy_policy.h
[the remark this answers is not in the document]
gonuke: No.  It was a new pattern to me, and may not be obvious to all.  Perhaps a comment that says something like
`// adding all options to a std::set to allow std::min to consider more than 2 choices`

gonuke (changes requested):
Nearly there.... I think the logic is more clear now, but still very confusing to follow when certain conditions constrain things.  

I'm thinking it might make sense to have some local state variables:
```
// pseudo-code
double amt = 0;
bool request_allowed = True;
double max_request_size = max_double;

if ( !MakeReq() || dormant() ) request_allowed = False;

if (request_allowed && using_cummulative_capacity()) max_request_size = cumulative_capacity_ - cycle_total_inv_;

if (request_allowed) amt = std::min( TotalAvailable() * SampleRequestSize(), max_request_size);

resetActiveDormant();
```

src/toolkit/matl_buy_policy.cc
```diff
@@ -43,18 +45,46 @@ void MatlBuyPolicy::set_manager(Agent* m) {
...
 }
 
-void MatlBuyPolicy::set_req_when_under(double x) {
-  if (x > 1)
-    x /= buf_->capacity();
-  assert(x > 0 && x <= 1.);
-  req_when_under_ = x;
+void MatlBuyPolicy::set_req_at(double x) {
+  assert(x >= 0);
+  req_at_ = x;
```
gonuke, line 82: Don't we need to check this against the buffer capacity, too?  Alternatively, we could just set it to the buffer capacity if too large?
```suggestion
  req_at_ = std::min(x, buf_->capacity());
```

src/toolkit/matl_buy_policy.cc
```diff
@@ -43,18 +45,46 @@ void MatlBuyPolicy::set_manager(Agent* m) {
...
+}
+
+void MatlBuyPolicy::set_cumulative_cap(double x) {
+  assert(x > 0);
+  cumulative_cap_ = x;
```
gonuke, line 87: check against buffer capacity? or set to min of x and buffer capacity?

src/toolkit/matl_buy_policy.cc
```diff
@@ -43,18 +45,46 @@ void MatlBuyPolicy::set_manager(Agent* m) {
   }
 }
 
+void MatlBuyPolicy::set_total_inv_tracker(TotalInvTracker* t) {
+  if (t == NULL){
+    std::vector<ResBuf<Material>*> bufs = {buf_};
+    buf_tracker_->Init(bufs, buf_->capacity());
+  }
+  else {
+    buf_tracker_ = t;
```
gonuke, line 54: should we make sure that `buf_` is part of `buf_tracker_`?  That is, make sure that the `buf_tracker_` includes `buf_` as one of it's buffers?

src/toolkit/matl_buy_policy.cc
```diff
@@ -256,9 +327,15 @@ void MatlBuyPolicy::SetNextActiveTime() {
...
 
 void MatlBuyPolicy::SetNextDormantTime() {
-  if (next_dormant_end_ < 0) {}
+  if (use_cumulative_capacity()) {
+    // need the +1 when not using next_active_end_ 
+    next_dormant_end_ = (
+      dormant_dist_->sample() + manager()->context()->time() + 1);
+  }
+  else if (next_dormant_end_ < 0) {}
   else {
```
gonuke, line 336:
```suggestion
  else if (next_dormant_end_ >= 0) {
```

src/toolkit/matl_buy_policy.cc
```diff
@@ -267,6 +344,17 @@ double MatlBuyPolicy::SampleRequestSize() {
   return size_dist_->sample();
 }
 
+void MatlBuyPolicy::CheckActiveDormantCumulativeTimes() {
+  if (manager()->context()->time() == next_dormant_end_) {
+    SetNextActiveTime();
+    if (use_cumulative_capacity()) { next_dormant_end_ = -1; }
+    else {
```
gonuke, line 351: `next_active_time_` is irrelevant when using cumulative capacity, so only set it when not.
```suggestion
    if (use_cumulative_capacity()) { next_dormant_end_ = -1; }
    else { 
        SetNextActiveTime();

```

gonuke (changes requested):
A few more tweaks for robustness and long-term readability

gonuke:
Merging now because the cycamore update is close to ready

PR #1666, opened 2024-02-07
Fix Python segmentation fault behavior

bennibbelink:
This prevents segmentation faults described in #1589 and removes the `--ignore test_main.py` flag from the python tests in the Dockerfile.  

There are three cython modules in the cython core code - `pymodule.pyx`, `eventhooks.pyx`, and `pyinfile.pyx`.  There is functionality in `pyhooks.cc` that manages the python module initialization.  When cyclus is called from python, the python interpreter is already initialized so these three modules must be imported explicitly (previously "handled" by `py_import_call_init()` in `cyclus/__init__.py`).  This wasn't working correctly, so the cython modules are not appropriately initialized and this is where the segmentation faults occur.  

The current fix builds these three cython modules as their own shared libraries so that python can import them.  I modified `cyclus/__init__.py` to import these modules explicitly instead of using `py_import_call_init()`.  I was able to fix my dummy repo by calling `py_import_call_init()` followed by `py_import_init()` to actually import the modules, but have struggled to get this working in cyclus (I think it is the "best practice" fix, I can continue to work at implementing this in cyclus).

Another significant change to the CMake build (that can be up for debate) is that these three shared libraries are installed to both `/root/.local/lib` and python site packages.  This is so cyclus can find the shared libraries on LD_LIBRARY_PATH and python can find the modules in its path.  This was a design choice I made, but we can choose to install them to one location and modify our environment variables accordingly.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -45,7 +45,7 @@ Since last release
   quantize, bids that one was not able to fullfill and caused cyclus to crash. (#1552)
 * Resolve deprecation warnings involving <boost/detail/sp_typeinfo.hpp> (#1611)
 * Resolve segmentation faults when calling Cbc (#1614)
-* Resolve segmentation faults wehn using cyclus via Python (#1666)
+* Resolve segmentation faults when using cyclus via Python (#1666)
 
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -45,6 +45,7 @@ Since last release
   quantize, bids that one was not able to fullfill and caused cyclus to crash. (#1552)
 * Resolve deprecation warnings involving <boost/detail/sp_typeinfo.hpp> (#1611)
 * Resolve segmentation faults when calling Cbc (#1614)
+* Resolve segmentation faults wehn using cyclus via Python (#1666)
 
 
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -130,22 +130,23 @@ if(Cython_FOUND)
     foreach(file ${CYCLUS_CORE_CYTHON_FILES})
         message(STATUS "Cython Compiling ${file}")
         get_filename_component(name ${file} NAME_WE)
-        set(_generated_h "${CMAKE_CURRENT_SOURCE_DIR}/${name}.cpp.h")
-        set(_generated_cpp "${CMAKE_CURRENT_SOURCE_DIR}/${name}.cpp.gen")
+        set(_generated_h "${CMAKE_CURRENT_SOURCE_DIR}/${name}.cc.h")
+        set(_generated_cc "${CMAKE_CURRENT_SOURCE_DIR}/${name}.cc.gen")
         set(_h_file "${CMAKE_CURRENT_SOURCE_DIR}/${name}.h")
-        set(_cpp_file "${CMAKE_CURRENT_SOURCE_DIR}/${name}.cpp")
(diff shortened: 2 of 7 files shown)


CHANGELOG.rst
```diff
@@ -45,6 +45,7 @@ Since last release
   quantize, bids that one was not able to fullfill and caused cyclus to crash. (#1552)
 * Resolve deprecation warnings involving <boost/detail/sp_typeinfo.hpp> (#1611)
 * Resolve segmentation faults when calling Cbc (#1614)
+* Resolve segmentation faults wehn using cyclus via Python (#1666)
```
gonuke, line 48:
```suggestion
* Resolve segmentation faults when using cyclus via Python (#1666)
```

src/CMakeLists.txt
```diff
@@ -129,22 +130,43 @@ if(Cython_FOUND)
     foreach(file ${CYCLUS_CORE_CYTHON_FILES})
         message(STATUS "Cython Compiling ${file}")
         get_filename_component(name ${file} NAME_WE)
-        set(_generated_h "${CMAKE_CURRENT_SOURCE_DIR}/${name}.cc.h")
-        set(_generated_cc "${CMAKE_CURRENT_SOURCE_DIR}/${name}.cc.gen")
+        set(_generated_h "${CMAKE_CURRENT_SOURCE_DIR}/${name}.cpp.h")
+        set(_generated_cpp "${CMAKE_CURRENT_SOURCE_DIR}/${name}.cpp.gen")
```
gonuke, line 134: Are these changes substantive or just aesthetic?

gonuke (changes requested):
Thanks for figuring this out @bennibbelink - great sleuthing!

src/CMakeLists.txt
bennibbelink: Substantive.  CMake grabs all of the  `.cc` files in `src/` and compiles them into `libcyclus.so` but I don't want it to include the cython modules since those are built as their own respective shared libraries.  Renaming them to `*.cpp` is one way to do this, though I can just modify the GLOB that grabs the files to compile if we want to keep consistent file extensions.

gonuke: Where is that GLOB?  I feel that modifying the GLOB may be more explicit and leave a better record of its importance.  It may be difficult for future me (or others) to recognize the importance of this file extension difference.

src/CMakeLists.txt
```diff
@@ -175,6 +198,7 @@ foreach(ccfile ${CYCLUS_COIN_SRC})
 endforeach()
 list(REMOVE_ITEM cc_files ${CMAKE_CURRENT_SOURCE_DIR}/coin_helpers.cc)
 list(REMOVE_ITEM cc_files ${CMAKE_CURRENT_SOURCE_DIR}/prog_translator.cc)
+list(REMOVE_ITEM cc_files ${CythonModuleSrc})
```
gonuke, line 201: Can you add a comment to indicate why we need to remove these, for posterity?

gonuke (changes requested):
Just one more comment as a breadcrumb for the future.

src/CMakeLists.txt
bennibbelink: Comment added

bam241:
Just wanted to say `Whooo!`.
Great job @bennibbelink ! 
I just wished, you had time 5y ago :)

PR #1667, opened 2024-02-13
Fix CI Refactor

bennibbelink:
Updates the workflows to push images/caches to the ghcr of `github.repository_owner` instead of hardcoded cyclus.

Also adds the PR comment workflow along the same lines as cymetric and cycamore.

[Test PR](https://github.com/bennibbelink/cyclus/pull/10) in my fork

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -23,9 +23,6 @@ on:
 jobs:
   build-test-cyclus:
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      packages: write
     strategy:
       fail-fast: false
       matrix:
@@ -38,6 +35,12 @@ jobs:
           conda,
         ]
 
+    services:
+      registry:
+        image: registry:2
+        ports:
+          - 5000:5000
+
     steps:
       # - name: Log in to the Container registry
       #   uses: docker/login-action@v3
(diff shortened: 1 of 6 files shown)


gonuke:
Don't we expect more of these tests to work because they should be invoking the newer workflow files?

bennibbelink:
I don't think they will invoke the newer workflow files until after this gets merged (?)

In cycamore this was the case.  For example - `publish_latest.yml` failed in the [PR](https://github.com/cyclus/cycamore/actions/runs/7880840245) but succeeded on [merge](https://github.com/cyclus/cycamore/actions/runs/7893269055)

gonuke:
Hmmm... then how do we test workflows?

gonuke:
I just realized that maybe `repository_owner` is not right b/c I think that resolves to `cyclus` for all PRs.  It may need to be `actor`.

bennibbelink:
Good catch... I thought I explored that in the past but I'm not sure why I didn't follow through.  I'll give it a shot and we can see if these workflows start passing

gonuke:
Scratch all that…. This appears to be using the updated workflow, but `github.repository_owner` is resolving to `cyclus`?

bennibbelink:
Yes I think because it is triggered by `pull_request` it runs in the context of the base repository.  `pull_request_target` would cause the workflow to run in the target repository (on my fork) but that is not best practice for security reasons (it would give contributors access to secrets they shouldn't have access to)

bennibbelink:
And because the PR is coming from a fork the GITHUB_TOKEN only has read permission - https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

I think your suggestion of using the triggering actor instead of repository_owner might be a good solution

bennibbelink:
closing in favor of #1668

Issue #583, opened 2013-09-12
add tests for matrix-related classes


gidden:
unfortunately, no such tests exists. the classes in question are:

- [ ] LMatrix
- [ ] UniformTaylor

I'm not sure what level of priority this should take, and perhaps would be a good exercise for a new entrant to the dev team, because:
- matrix operations are well-known linear-programming techniques
- the implementation is already written
- it is purely an exercise in writing tests
- we could dovetail documentation issues into this (i.e., make a separate issue to document these classes using doxygen, which both are currently lacking)
- issue #495 could be tackled at the same time

I'll assign to @scopatz for now and let him unassign or reassign.

gidden:
moved to 1.4

rwcarlsen:
closing in favor of #1134

Issue #195, opened 2012-06-18
Add Building Management Utility


gidden:
For agents that wish to build objects, a utility to manage that building process is needed. This manager should solve a simple optimization problem:

   \min \sum_{i=1}^{N}n_i_c_i  
   s.t. \sum_{i=1}^{N}n_i_\phi_i \ge \Phi  
        n_i \in [0,\infty) \forall i \in I, n_i integer               

Where n_i is the number of objects of type i to build, c_i is the cost to build the object of type i, \phi_i is the nameplate capacity of the object, and \Phi is the capacity demand. Here the set I corresponds to all producers of a given product.  

The BuildingManager would take as an argument a SupplyDemandManager and then make build decisions given a product and an unmet demand for that product.

PR #1668, opened 2024-02-14
CI Updates

bennibbelink:
This fixes the CI permissions problems we're seeing (along with cymetric [#195](https://github.com/cyclus/cymetric/pull/195) and cycamore [#583](https://github.com/cyclus/cycamore/pull/583)).  Sorry that this hasn't gone as smoothly as possible, hopefully this is the last major refactor.

Changes are as follows:
- `build_test.yml` does not push any images to the cyclus GHCR.  Instead it uses a local docker registry on the GitHub runner.  This way we don't experience permissions errors with PRs from forks.  This adds a tiny bit of complexity that I wanted to avoid, but honestly I think the workflow might be more intuitive now than in previous iterations.  I tried implementing this without a local registry service but docker automatically tries to pull metadata so I was seeing lots of permission errors.
  - we still use build-contexts to "redefine" the cyclus image being used in downstream testing but now it references the image in our local registry
- `publish_latest.yml` is the only workflow that updates the layer cache.  This way all workflows essentially pull the layers that build the dependency image in `cyclus:latest`.  Most workflows pull from this cache
- `code_coverage.yml` has its own cache since its dependencies are slightly different, but I've included the `ignore-error` option so that cache-to won't stop a workflow from passing in a PR
- all workflows that publish images (`publish_latest.yml` and `publish_release.yml`) need to be updated from a branch on the upstream repository.  This is similar to our previous strategy
- in the cycamore and cymetric PRs I changed the cyclus_tag/cycamore_tag in the Dockerfile to `latest`.  This is to eliminate some verbosity in the workflow files and ensure users are using the latest versions when building from the Dockerfile

Once again I've reference branches in cycamore and cymetric corresponding to their respective PRs with these changes

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -68,6 +68,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: cyclus/cycamore
+          ref: ci-updates
           path: ${{ github.workspace }}/cycamore
 
       - name: Build and Test Cycamore
@@ -92,6 +93,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: cyclus/cymetric
+          ref: ci-updates
           path: ${{ github.workspace }}/cymetric
 
       - name: Build and Test Cymetric
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -23,10 +23,7 @@ on:
 jobs:
   build-test-cyclus:
     runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      packages: write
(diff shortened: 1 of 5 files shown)


bennibbelink:
Something this PR also does is add a trigger for `code_coverage.yml` on push to main.  This will update the coverage reference in coveralls on a push to main.  This did not exist before which I think is why we still see a coverage drop from 49%, we never ran the coverage workflow after merge to update the coverage of `main`

gonuke (approved):
Hopefully we are done for now?

bennibbelink:
Hopefully 🤞

PR #1671, opened 2024-02-18
Resolve compilation warnings in Cyclus build

bennibbelink:
Summary of the changes:

- `src/any.hpp` I found resolution from [newer boost version](https://www.boost.org/doc/libs/1_84_0/boost/spirit/home/support/detail/hold_any.hpp)
- `src/cyc_std.h std::binary_function` is deprecated.  We can just use `std::function` now - https://stackoverflow.com/questions/33114656/replacement-for-stdbinary-function
- `src/greedy_solver.cc` and `src/resource_exchange.h` - `std::bind1st` is deprecated. We can just use `std::bind` now https://stackoverflow.com/questions/52769059/replacement-for-removed-bind1st-in-c17
- `src/discovery.cc` a couple deprecated methods - I followed the suggestions that the warnings listed
- `src/pyhooks.[cc,h]`, `cyclus/cpp_cyclus.pxd`, `cyclus/lib.pyx` - removed a function from pyhooks that we no longer use (and was bad practice given the warnings)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,6 +34,7 @@ Since last release
 * Major update and modernization of build (#1587, #1632)
 * Changed Json formatting for compatibility with current python standards (#1587)
 * Changed README.rst installation instructions, tested on fresh Ubuntu-22.04 system with Python 3.11 (#1617, #1644)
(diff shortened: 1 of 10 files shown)


gonuke (approved):
Thanks for this great cleanup @bennibbelink

Issue #1620, opened 2023-11-21
Implement packaging of bulk to discrete items


gonuke:

nuclearkatie:
Closed by core packaging PRs #1673, #1675, #1683, #1729, and #1749

PR #1673, opened 2024-02-19
Package class and declaration of packages in input file

### LINKED ISSUE #1654: Design and implement `Package` class
Define member variables and methods for class

### LINKED ISSUE #1655: Read in set of Packages and store in Context
Following the model of `Recipe` load available packages from input into context.

Always populate Context with a default implicit packaging instance.

nuclearkatie:
- creates Package class
- packages can be declared in input file (optional)

Closes #1654, closes #1655 
Part of #1620

### DIFF
--- a/src/context.h
+++ b/src/context.h
@@ -351,8 +351,7 @@ class Context {
 
   std::map<std::string, Agent*> protos_;
   std::map<std::string, Composition::Ptr> recipes_;
-  //std::map<std::string, Package::Ptr> packages_;
-  std::map<std::string, boost::shared_ptr<Package>> packages_;
+  std::map<std::string, Package::Ptr> packages_;
   std::set<Agent*> agent_list_;
   std::set<Trader*> traders_;
   std::map<std::string, int> n_prototypes_;
--- a/src/context.h
+++ b/src/context.h
@@ -18,6 +18,7 @@
 #include "greedy_solver.h"
 #include "pyhooks.h"
 #include "recorder.h"
+#include "package.h"
 
 const uint64_t kDefaultTimeStepDur = 2629846;
 
@@ -248,6 +249,15 @@ class Context {
   /// Returns the current simulation timestep.
   virtual int time();
 
+  /// Adds a package type to a simulation-wide accessible list.
+  /// Agents should NOT add their own packages.
+  void AddPackage(std::string name, double fill_min = 0,
+                  double fill_max = std::numeric_limits<double>::max(),
+                  std::string strategy = "first");
+
+  // Retrieve a registered package. 
+  Package::Ptr GetPackage(std::string name);
+
   int random();
 
   /// Generates a random number on the range [0,1)]
@@ -341,6 +351,8 @@ class Context {
 
   std::map<std::string, Agent*> protos_;
   std::map<std::string, Composition::Ptr> recipes_;
+  //std::map<std::string, Package::Ptr> packages_;
+  std::map<std::string, boost::shared_ptr<Package>> packages_;
   std::set<Agent*> agent_list_;
   std::set<Trader*> traders_;
   std::map<std::string, int> n_prototypes_;
--- a/src/package.cc
+++ b/src/package.cc
@@ -1,0 +1,32 @@
+#include "package.h"
+#include "error.h"
+
+namespace cyclus {
+
+int Package::next_id_ = 1;
+
+Package::Ptr Package::Create() {
+  Ptr p(new Package());
+  return p;
+}
+
+Package::Ptr Package::Create(std::string name, double fill_min, double fill_max, std::string strategy) {
+  if (fill_min < 0 || fill_max < 0) {
+    throw ValueError("fill_min and fill_max must be non-negative");
+  }
+  else if (fill_min > fill_max) {
+    throw ValueError("fill_min must be less than or equal to fill_max");
+  }
+  Ptr p(new Package(name, fill_min, fill_max, strategy));
+  return p;
+}
+
+int Package::id() {
+  return id_;
+}
+  
+Package::Package() : id_(next_id_++), fill_min_(0), fill_max_(std::numeric_limits<double>::max()), strategy_("first") {}
+
+Package::Package(std::string name, double fill_min, double fill_max, std::string strategy) : name_(name), id_(next_id_++), fill_min_(fill_min), fill_max_(fill_max), strategy_(strategy) {}
+
+} // namespace cyclus
--- a/src/package.h
+++ b/src/package.h
@@ -1,6 +1,7 @@
 #ifndef CYCLUS_SRC_PACKAGE_H_
 #define CYCLUS_SRC_PACKAGE_H_
 
+#include <limits>
 #include <string>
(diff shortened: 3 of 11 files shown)


src/context.h
```diff
@@ -341,6 +351,8 @@ class Context {
 
   std::map<std::string, Agent*> protos_;
   std::map<std::string, Composition::Ptr> recipes_;
+  //std::map<std::string, Package::Ptr> packages_;
+  std::map<std::string, boost::shared_ptr<Package>> packages_;
```
gonuke, line 355: The version with `Package::Ptr` should work, right?

src/xml_file_loader.cc
```diff
@@ -326,6 +327,28 @@ void XMLFileLoader::LoadRecipes() {
...
+    InfileTree* qe = xqe.SubTree(query, i);
+    std::string name = cyclus::OptionalQuery<std::string>(qe, "name", "default");
+    CLOG(LEV_DEBUG3) << "loading package: " << name;
+    
+    double fill_min = cyclus::OptionalQuery<double>(qe, "fill_min", 0);
+    double fill_max = cyclus::OptionalQuery<double>(qe, "fill_max", std::numeric_limits<double>::max());
+    
+    std::string strategy = cyclus::OptionalQuery<std::string>(qe, "strategy", "first");
+
+    boost::shared_ptr<Package> p = Package::Create(name, fill_min, fill_max, strategy);
```
gonuke, line 347: It might make sense to put this in a `ReadPackage()` method analogous to `ReadRecipe()` but this one is a lot simpler, so could go either way.

src/package.h
```diff
@@ -0,0 +1,38 @@
+#ifndef CYCLUS_SRC_PACKAGE_H_
+#define CYCLUS_SRC_PACKAGE_H_
+
+#include <string>
+#include <boost/shared_ptr.hpp>
```
gonuke, line 5: This should fix your build error

```suggestion
#include <limits>
#include <string>
#include <boost/shared_ptr.hpp>
```

gonuke (changes requested):
This looks pretty good - I think I found the source of your build error.

src/xml_file_loader.cc
nuclearkatie: Yeah my thought was the simplicity negated the need for an extra method at this point

src/package.cc
```diff
@@ -0,0 +1,32 @@
...
+  }
+  Ptr p(new Package(name, fill_min, fill_max, strategy));
+  return p;
+}
+
+int Package::id() {
+  return id_;
+}
+  
+Package::Package() : id_(next_id_++), fill_min_(0), fill_max_(std::numeric_limits<double>::max()), strategy_("first") {}
```
gonuke, line 28: What does `strategy_ == "first"` imply/mean?  Should this just be blank?  Also, shouldn't you initialize the `name` as something, including an empty string?

gonuke (commented):
Another question, but I think this is nearly ready to go...

gonuke:
> * How to manage default? Right now I have the packages managed by context, including creating a default regardless if any are declared in the input file. But, packages isn't context-aware, so when I get around to packaging strategies, not sure how best to call the default value

I think the `default` will always have `id=0` and `name=default`.  You could pick a different default name that is more likely to be unique/rare, but I think it should be easy to access the default package through the context.

src/package.cc
nuclearkatie: This is me starting to think about the fill strategies. For now, just planning to do fill first `first` and fill equally `equal`. Fill first would be the default. But it's not set in stone, happy to use another strategy

nuclearkatie:
> > * How to manage default? Right now I have the packages managed by context, including creating a default regardless if any are declared in the input file. But, packages isn't context-aware, so when I get around to packaging strategies, not sure how best to call the default value
> 
> I think the `default` will always have `id=0` and `name=default`. You could pick a different default name that is more likely to be unique/rare, but I think it should be easy to access the default package through the context.

So you think package class should have the context as a member? As written, package doesn't have a way to access the context, because it doesn't have a manager or anything

gonuke:
> > > * How to manage default? Right now I have the packages managed by context, including creating a default regardless if any are declared in the input file. But, packages isn't context-aware, so when I get around to packaging strategies, not sure how best to call the default value
> 
> > 
> 
> > I think the `default` will always have `id=0` and `name=default`. You could pick a different default name that is more likely to be unique/rare, but I think it should be easy to access the default package through the context.
> 
> 
> 
> So you think package class should have the context as a member? As written, package doesn't have a way to access the context, because it doesn't have a manager or anything

No - I'm not sure the package needs to know about the context.  I think a resource will access info about available packages via the context and then store a direct link to their current package.  Am I missing something about why/how a package would reference its context?

nuclearkatie:
> > > > * How to manage default? Right now I have the packages managed by context, including creating a default regardless if any are declared in the input file. But, packages isn't context-aware, so when I get around to packaging strategies, not sure how best to call the default value
> > 
> > 
> > > 
> > 
> > 
> > > I think the `default` will always have `id=0` and `name=default`. You could pick a different default name that is more likely to be unique/rare, but I think it should be easy to access the default package through the context.
> > 
> > 
> > So you think package class should have the context as a member? As written, package doesn't have a way to access the context, because it doesn't have a manager or anything
> 
> No - I'm not sure the package needs to know about the context. I think a resource will access info about available packages via the context and then store a direct link to their current package. Am I missing something about why/how a package would reference its context?

Ah ok, so the default will always be passed through the resource. I was thinking of having a version of package where no package type was passed, and then the package would have to get access to the default

gonuke:
> > > > > * How to manage default? Right now I have the packages managed by context, including creating a default regardless if any are declared in the input file. But, packages isn't context-aware, so when I get around to packaging strategies, not sure how best to call the default value
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > I think the `default` will always have `id=0` and `name=default`. You could pick a different default name that is more likely to be unique/rare, but I think it should be easy to access the default package through the context.
> > > 
> > > 
> > > So you think package class should have the context as a member? As written, package doesn't have a way to access the context, because it doesn't have a manager or anything
> > 
> > 
> > No - I'm not sure the package needs to know about the context. I think a resource will access info about available packages via the context and then store a direct link to their current package. Am I missing something about why/how a package would reference its context?
> 
> Ah ok, so the default will always be passed through the resource. I was thinking of having a version of package where no package type was passed, and then the package would have to get access to the default

I think that any time we need to know about a package, we'll have access to the context.  I'm not sure I can think of a time when a package needs to know about the context without some other code-context to get the info it needs.  If we identify something, we'll have to rethink.

nuclearkatie:
Looking ahead, resources don't know about the context except when recording, where it is passed to record. So, somewhere is going to need to be context-aware in order to handle the default case. I'll do some more thinking about this...

gonuke (approved):
This is a great first step on Packaging @nuclearkatie

gonuke:
Sometime soon, coverage will settle down a little and I'll worry about that test, but I'll go ahead now

Issue #1669, opened 2024-02-14
Add workflow to check code style with clang-format


gonuke:
To impose coding style we should check each PR with clang-format and include a style file.  This is done now in [svalinn/DAGMC](https://github.com/svalinn/DAGMC/blob/develop/.github/workflows/housekeeping.yml)

ahnaf-tahmid-chowdhury:
Housekeeping of DAGMC is very complicated now. And it seems in the docker file, we are just installing clang-format through apt. Also, I noticed the docker publish file is changed and there is no way to rebuild a new docker image for housekeeping.

gonuke:
@dean-krueger noted that `clang-format` does not know how to properly deal with our `#pragma cycpp` code.  Some work may be necessary to add rules for `clang-format` to be useable.

PR #1674, opened 2024-02-19
Add Housekeeping to Check Code Style with `clang-format`

ahnaf-tahmid-chowdhury:
## Description

This PR adds a new GitHub Actions workflow named "Housekeeping Checks". The workflow is triggered on `workflow_dispatch` (manual runs) and `pull_request` events targeting the `develop` branch.

## Changes Made

- Added a new workflow file named `housekeeping.yml` to the `.github/workflows` directory.
- The workflow runs on `ubuntu-latest` and consists of the following steps:
  - Checkout the repository
  - Install `clang-format` dependencies
  - Setup environment variables
  - Execute the housekeeping script that runs `clang-format` on the source files and checks for style guide violations. If any are found, the script will fail, and the PR will not be merged.

## Related Issue

This PR addresses issue #1669.

## Additional Notes

- The housekeeping script uses `git status --porcelain` to check for changes in the working directory after running `clang-format`.
- If any changes are found, the script will print the differences and fail the workflow with an error message.
- This workflow helps maintain code style consistency and ensures that all changes adhere to the project's coding standards.


ahnaf-tahmid-chowdhury:
Here is my [Workflow](https://github.com/ahnaf-tahmid-chowdhury/cyclus/actions/runs/7963395049/job/21738857816)

ahnaf-tahmid-chowdhury:
I'm having trouble with housekeeping. Is it advisable to execute `clang-format` and create a new commit with the already available [.clang-format](https://github.com/cyclus/cyclus/blob/main/.clang-format) file?

ahnaf-tahmid-chowdhury:
Yes. I have noticed the coverage is 30%

dean-krueger:
> We (esp. @dean-krueger )have discovered that this is currently dangerous b/c of some features of Cyclus and may need more careful consideration.

It appears that now is the time to start digging this PR up. Sorry @ahnaf-tahmid-chowdhury for taking so long to get around to this.

It appears that the issue I was running into was because pragmas (`#pragma var { ...` specifically in this case) REQUIRES the space between var and {  as specified by the regex:

```
regex = re.compile(r"\s*#\s*pragma\s+cyclus\s+var\s+(.*)")
```

on line 475 of cycpp (in conjunction with the regex on line 398 of the same file). The `\s+var\s+` mandates that there be exactly 1 space before and after `var` in a `#pragma` statement. If this was changed to `\s*var\s*` it would probably stop flagging this error, but it seems like this "exactly one space" is a requirement because something like `string.strip().split()` is happening to actually parse these pragmas. That being said, the thing to do to fix this is to dig into where exactly these `.split()`s are happening, and somehow modify them to allow `var{` instead of `var {`. This would likely be quite difficult, though, since I'm not certain that all pragma decorators take dicts. 

One other option is that we could simply run some sort of script along the lines of `cyclus-formatter.py` that ran clang-format on the code, and then went back in, found all the pragmas, and then added the spaces back in. This would potentially not fit clang-format specs in that there may be 81 chars on a line instead of ≤ 80 (or whatever the number is), but that may be better than trying to fix the, honestly, very confusing cycpp file and risk breaking everything massively. This is also a slightly bulkier, less elegant fix, but it would basically replace users needing to run clang-format on the code with having to run a formatting script, so not too much bulkier.

I'd be interested to see what thoughts @gonuke has on these options.

gonuke:
The best option is to convince clang-format to keep a space between `var` and its opening curly brace... let me know if anyone can figure out how to do that?

I suppose another option is that on the first pass of cycpp, we could do a regex replacement for `var{` => `var {`?

dean-krueger:
One thing that's been bugging me is that in the Google Style guide it demands that there be a space between pretty much everything and the "{". ie.

```
void Foo(double bar) {
```

Leaving this comment mostly for me to try something later, but I also noticed that our clang-format file is 9 years old and specifies Cpp11 (I think the current version of C++ we use in cyclus is certainly newer than that. Perhaps this file is just super out of date and if we use a new one it will work?

gonuke:
> One thing that's been bugging me is that in the Google Style guide it demands that there be a space between pretty much everything and the "{". ie.
> 
> ```
> void Foo(double bar) {
> ```

I think clang-format needs to be able to determine the full context to know how to handle braces (see [details on brace wrapping](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#bracewrapping)).  Because our case is in the `#pragma` and not a standard C++ construct, I don't think it can figure out enough context to make a determination of how to handle the braces.

gonuke:
> It appears that the issue I was running into was because pragmas (`#pragma var { ...` specifically in this case) REQUIRES the space between var and { as specified by the regex:
> 
> ```
> regex = re.compile(r"\s*#\s*pragma\s+cyclus\s+var\s+(.*)")
> ```
> 
> on line 475 of cycpp (in conjunction with the regex on line 398 of the same file). The `\s+var\s+` mandates that there be exactly 1 space before and after `var` in a `#pragma` statement. If this was changed to `\s*var\s*` it would probably stop flagging this error, but it seems like this "exactly one space" is a requirement because something like `string.strip().split()` is happening to actually parse these pragmas. That being said, the thing to do to fix this is to dig into where exactly these `.split()`s are happening, and somehow modify them to allow `var{` instead of `var {`. This would likely be quite difficult, though, since I'm not certain that all pragma decorators take dicts.

I don't think we rely on `split()` anywhere - the regex above captures all the things that match after `var\s+` in the `(.*)`.  So I think we only need to get the last part of this regex correct. 

Also, the regex `\s+var\s+(.*)` does not require exactly one space, but requires at least one space.  Switching to `\s+var\s*(.*)` would allow zero spaces, but might cause other problems.... It may be better to try `\s+var\s*({.*)` so that it looks specifically for the opening brace and includes it in the captured regex match.  There may need to be something special to escape that curly brace since it probably also has a semantic meaning in the regex.

dean-krueger:
Update on this head-scratcher. It seems like, while the problem LOOKS like it comes from the regular expression on L398 of `cycpp.py`, the real issue is deeper than that, and has to do with the way clang-format breaks up the JSON formatting of the pragmas themselves by adding spaces and stuff like that, which throws off `Filter.transform()`, and means we **CANNOT** fix it simply by changing the regex on L398. If we want to fix this, the best option is to do something like adding 

```
// clang-format off
#pragma cyclus var { ... }
// clang-format on
``` 

around EVERY pragma block. So far as my searching (and ChatGPT, which is actually the more convincing source in this case) is concerned, there is NO way to modify clang-format to not only ignore the `...var {` vs `..var{` thing I identified earlier AND the correct JSON formatting. It's just not capable of that.

I am currently working on a python script that will go through, find all the `#pragma` bits, add `//clang-format off` and `//clang-format on` around them, format the code, then remove all the extra lines I added, so if that ends up working we should be able to just change the test that's being run to whether or not that script makes any substantive changes to the files. This feels like (again, if it works) a fine middle ground, but interested in your-all's thoughts.

## TLDR: 
clang-format is breaking things in more ways than one, and there isn't a simple fix we can do on the clang-format configuration side or the cycpp.py side to fix it. I am working on a third option that "brute force" ignores all the pragmas, but we'll see if it ends up being viable or not.

dean-krueger:
I don't know if there is, but my understanding is that clang-format just thinks everything is some variety of C code and sort of puts its horse blinders on and turns the crank. I tried to see if there was a way to get it to do both, but it looks like there isn't, and then I tried to see if there were any tools outside of clang-format that would work and it seemed like there weren't any clearly-great options.

dean-krueger:
I tried to mess around with cycpp.py this morning and my conclusion is that this is not the right path. That file is so dense and the rabbit hole goes so deep that I think the right thing to do is to just advise developers to manually put `//clang-format off ... //clang-format on` around all their pragma blocks, whether that be all at once or individually, and to just be careful with the way they format the blocks. I'm drafting a page for the C++ tutorial on `fuel cycle.org` to walk through writing pragma blocks, and will go through and make an issue in Cycamore to add this guard to all the files in there and then we can hopefully move forward with this.

dean-krueger:
With #1880 and #1881 merged, now, I think that this SHOULD be good for another look at least, and perhaps just a straight merge. I wonder if there's any additional testing that should be done, after #1880 and #1881, but I'll toss that over the wall at @gonuke for him to decide whether to toss back over at me or @ahnaf-tahmid-chowdhury . Either way, I think a lot of the ground work for this to get merged has been done now. Thanks for being patient with me everyone!

PR #1675, opened 2024-02-21
Package ID for resources (materials/products)

### LINKED ISSUE #1656: Extend Resource to require Package
Ensure all resources have a reference to it's packaging, using an implicit packaging as a default

nuclearkatie:
Part of #1620 
Closes #1656

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -203,11 +203,21 @@ void Context::AddPackage(std::string name, double fill_min, double fill_max,
     ->Record();
 }
 
-Package::Ptr Context::GetPackage(std::string name) {
+Package::Ptr Context::GetPackageByName(std::string name) {
   if (packages_.count(name) == 0) {
     throw KeyError("Invalid package name " + name);
   }
   return packages_[name];
+}
+
+Package::Ptr Context::GetPackageById(int id) {
+  // iterate through the list of packages to get the one package with the correct id
+  std::map<std::string, Package::Ptr>::iterator it;
+  for (it = packages_.begin(); it != packages_.end(); ++it) {
+    if (it->second->id() == id) {
+      return it->second;
+    }
+  }
 }
 
 void Context::InitSim(SimInfo si) {
--- a/src/material.cc
+++ b/src/material.cc
@@ -12,8 +12,6 @@ namespace cyclus {
 
 const ResourceType Material::kType = "Material";
 
-static int default_package_id_ = 1;
-
 Material::~Material() {}
 
 Material::Ptr Material::Create(Agent* creator, double quantity,
--- a/src/material.cc
+++ b/src/material.cc
@@ -12,6 +12,8 @@ namespace cyclus {
 
 const ResourceType Material::kType = "Material";
 
+static int default_package_id_ = 1;
+
 Material::~Material() {}
 
 Material::Ptr Material::Create(Agent* creator, double quantity,
@@ -23,9 +25,7 @@ Material::Ptr Material::Create(Agent* creator, double quantity,
 
 Material::Ptr Material::CreateUntracked(double quantity,
                                         Composition::Ptr c) {
-  // default package id for untracked
-  int package_id = 1;
-  Material::Ptr m(new Material(NULL, quantity, c, package_id));
+  Material::Ptr m(new Material(NULL, quantity, c, default_package_id_));
   return m;
 }
 
@@ -90,9 +90,7 @@ Material::Ptr Material::ExtractComp(double qty, Composition::Ptr c,
   }
 
   qty_ -= qty;
-  // extracted material has default package
-  int package_id = 1;
-  Material::Ptr other(new Material(ctx_, qty, c, package_id));
+  Material::Ptr other(new Material(ctx_, qty, c, default_package_id_));
 
   // Decay called on the extracted material should have the same dt as for
   // this material regardless of composition.
--- a/src/material.cc
+++ b/src/material.cc
@@ -154,7 +154,7 @@ void Material::ChangePackageId(int new_package_id) {
     package_id_ = new_package_id;
     return;
   }
-  
+ 
   Package::Ptr p = ctx_->GetPackageById(package_id_);
   double min = p->fill_min();
   double max = p->fill_max();
--- a/src/material.cc
+++ b/src/material.cc
@@ -15,15 +15,17 @@ const ResourceType Material::kType = "Material";
 Material::~Material() {}
 
 Material::Ptr Material::Create(Agent* creator, double quantity,
-                               Composition::Ptr c) {
-  Material::Ptr m(new Material(creator->context(), quantity, c));
+                               Composition::Ptr c, int package_id) {
+  Material::Ptr m(new Material(creator->context(), quantity, c, package_id));
   m->tracker_.Create(creator);
   return m;
 }
 
 Material::Ptr Material::CreateUntracked(double quantity,
                                         Composition::Ptr c) {
-  Material::Ptr m(new Material(NULL, quantity, c));
+  // default package id for untracked
+  int package_id = 1;
+  Material::Ptr m(new Material(NULL, quantity, c, package_id));
   return m;
 }
 
@@ -48,6 +50,7 @@ void Material::Record(Context* ctx) const {
   ctx_->NewDatum("MaterialInfo")
       ->AddVal("ResourceId", state_id())
       ->AddVal("PrevDecayTime", prev_decay_time_)
+      ->AddVal("PackageId", package_id_)
       ->Record();
 
   comp_->Record(ctx);
@@ -87,8 +90,9 @@ Material::Ptr Material::ExtractComp(double qty, Composition::Ptr c,
   }
 
   qty_ -= qty;
-
-  Material::Ptr other(new Material(ctx_, qty, c));
+  // extracted material has default package
+  int package_id = 1;
+  Material::Ptr other(new Material(ctx_, qty, c, package_id));
 
   // Decay called on the extracted material should have the same dt as for
   // this material regardless of composition.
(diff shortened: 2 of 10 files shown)


src/resource.h
```diff
@@ -5,6 +5,8 @@
 #include <vector>
 #include <boost/shared_ptr.hpp>
 
+#include "package.h"
+
```
nuclearkatie, line 9: Don't currently use this, should remove

src/material.cc
```diff
@@ -15,15 +15,17 @@ const ResourceType Material::kType = "Material";
...
+  Material::Ptr m(new Material(creator->context(), quantity, c, package_id));
   m->tracker_.Create(creator);
   return m;
 }
 
 Material::Ptr Material::CreateUntracked(double quantity,
                                         Composition::Ptr c) {
-  Material::Ptr m(new Material(NULL, quantity, c));
+  // default package id for untracked
+  int package_id = 1;
```
gonuke, line 27: We should probably create a variable or a function with the default ID, because I thought it was 0, not 1

src/material.cc
```diff
@@ -15,15 +15,17 @@ const ResourceType Material::kType = "Material";
...
+  Material::Ptr m(new Material(NULL, quantity, c, package_id));
```
gonuke, line 28: Probably don't need this change if you add a default in the constructor declaration.

src/material.cc
```diff
@@ -87,8 +90,9 @@ Material::Ptr Material::ExtractComp(double qty, Composition::Ptr c,
   }
 
   qty_ -= qty;
-
-  Material::Ptr other(new Material(ctx_, qty, c));
+  // extracted material has default package
+  int package_id = 1;
+  Material::Ptr other(new Material(ctx_, qty, c, package_id));
```
gonuke, line 95: This change may not be necessary with default in constructor.

src/material.cc
```diff
@@ -236,4 +241,12 @@ Material::Ptr NewBlankMaterial(double quantity) {
   return Material::CreateUntracked(quantity, comp);
 }
 
+int Material::package_id() {
+  return package_id_;
+}
+
+void Material::ChangePackageId(int new_package_id) {
```
gonuke, line 248: Do we want it to be this easy to change the package ID?

src/material.h
```diff
@@ -153,15 +153,22 @@ class Material: public Resource {
...
 
+  /// Returns the package id.
+  virtual int package_id();
+
+  /// Change the package id
+  virtual void ChangePackageId(int new_package_id);
+
  protected:
-  Material(Context* ctx, double quantity, Composition::Ptr c);
+  Material(Context* ctx, double quantity, Composition::Ptr c, int package_id);
```
gonuke, line 163: Define a default package_id here?

src/product.cc
```diff
@@ -12,25 +12,28 @@ int Product::next_qualid_ = 1;
...
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 Product::Ptr Product::Create(Agent* creator, double quantity,
-                             std::string quality) {
+                             std::string quality, int package_id) {
   if (qualids_.count(quality) == 0) {
     qualids_[quality] = next_qualid_++;
     creator->context()->NewDatum("Products")
         ->AddVal("QualId", qualids_[quality])
         ->AddVal("Quality", quality)
+        ->AddVal("PackageId", 1)
```
gonuke, line 21:
```suggestion
        ->AddVal("PackageId", package_id)
```

gonuke (commented):
Some early thoughts/feedback

src/material.cc
nuclearkatie: Things like compositions and recipes start at 1

gonuke: OK - then we should make sure the packages start at 1 as well

src/material.cc
```diff
@@ -12,18 +12,20 @@ namespace cyclus {
 
 const ResourceType Material::kType = "Material";
 
+static int default_package_id_ = 1;
```
gonuke, line 15: This can only exist in once place, so this may not be the best place.  There are other examples in the code base.

It may also make sense to make it a static member variable of the `Package` class (which could mean that the one place it exists is the `package.cc` file.

src/material.h
```diff
@@ -81,7 +81,7 @@ class Material: public Resource {
   /// pointer to the agent creating the resource (usually will be the caller's
   /// "this" pointer). All future output data recorded will be done using the
   /// creator's context.
-  static Ptr Create(Agent* creator, double quantity, Composition::Ptr c);
+  static Ptr Create(Agent* creator, double quantity, Composition::Ptr c, int package_id = 1);
```
gonuke, line 84:
```suggestion
  static Ptr Create(Agent* creator, double quantity, Composition::Ptr c, int package_id = default_package_id_);
```

src/material.h
```diff
@@ -153,15 +153,22 @@ class Material: public Resource {
...
 
+  /// Returns the package id.
+  virtual int package_id();
+
+  /// Change the package id
+  virtual void ChangePackageId(int new_package_id);
+
  protected:
-  Material(Context* ctx, double quantity, Composition::Ptr c);
+  Material(Context* ctx, double quantity, Composition::Ptr c, int package_id = 1);
```
gonuke, line 163:
```suggestion
  Material(Context* ctx, double quantity, Composition::Ptr c, int package_id = default_package_id_);
```

src/product.h
```diff
@@ -28,7 +28,7 @@ class Product : public Resource {
   /// pointer to the agent creating the resource (usually will be the caller's
   /// "this" pointer). All future output data recorded will be done using the
   /// creator's context.
-  static Ptr Create(Agent* creator, double quantity, std::string quality);
+  static Ptr Create(Agent* creator, double quantity, std::string quality, int package_id = 1);
```
gonuke, line 31:
```suggestion
  static Ptr Create(Agent* creator, double quantity, std::string quality, int package_id = default_package_id_);
```

src/product.h
```diff
@@ -75,7 +75,7 @@ class Product : public Resource {
   /// @param ctx the simulation context
   /// @param quantity is a double indicating the quantity
   /// @param quality the resource quality
-  Product(Context* ctx, double quantity, std::string quality);
+  Product(Context* ctx, double quantity, std::string quality, int package_id = 1);
```
gonuke, line 78:
```suggestion
  Product(Context* ctx, double quantity, std::string quality, int package_id = default_package_id_);
```

gonuke (changes requested):
Some cleanup on the default package id usage

src/product.h
```diff
@@ -71,6 +71,9 @@ class Product : public Resource {
   /// @throws ValueError 'other' resource is of different quality
   void Absorb(Product::Ptr other);
 
+  /// Changes the product's package id
+  void ChangePackageId(int package_id);
```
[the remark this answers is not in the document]
gonuke: See above comment, but this is marked `virtual` in `Material` and not here...

gonuke (changes requested):
A few changes requested here - let me know if you want to chat about any of them

src/material.cc
```diff
@@ -236,4 +237,12 @@ Material::Ptr NewBlankMaterial(double quantity) {
   return Material::CreateUntracked(quantity, comp);
 }
 
+int Material::package_id() {
+  return package_id_;
+}
+
+void Material::ChangePackageId(int new_package_id) {
```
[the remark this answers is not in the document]
nuclearkatie: I forgot the meeting was cancelled this week, I was going to go over this there, but my proposal was that this function should be called by the resource buffer only after doing the actual partitioning step. Aka, this is like the last step in doing the change, where you just do one last check that the resource has been shaped to be in compliance but not any of the work to do the shaping (which will be done by the resource buffer and res_manip helpers)

src/context.cc
```diff
@@ -203,13 +203,23 @@ void Context::AddPackage(std::string name, double fill_min, double fill_max,
...
 
-Package::Ptr Context::GetPackage(std::string name) {
+Package::Ptr Context::GetPackageByName(std::string name) {
   if (packages_.count(name) == 0) {
     throw KeyError("Invalid package name " + name);
   }
   return packages_[name];
 }
 
+Package::Ptr Context::GetPackageById(int id) {
```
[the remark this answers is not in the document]
nuclearkatie: Archetypes call `GetRecipe` to figure out the composition they need. Otherwise, Cyclus doesn't call `GetRecipe` itself

src/material.h
```diff
@@ -156,8 +156,9 @@ class Material: public Resource {
   /// Returns the package id.
   virtual int package_id();
 
-  /// Change the package id
-  virtual void ChangePackageId(int new_package_id);
+  /// Changes the package id. Checks that the resource fits the package 
+  /// type minimum and maximum mass criteria.
+  virtual void ChangePackageId(int new_package_id = default_package_id_);
```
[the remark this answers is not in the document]
nuclearkatie: I wasn't sure if it was better practice to put it in `Resource` or `Material` and `Product`. They both use `quantity` so there shouldn't be any difference in implementation

src/context.cc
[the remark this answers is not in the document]
gonuke: But they store the recipe "name" not the "ID", right?

src/material.h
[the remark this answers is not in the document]
gonuke: So long as their implementation is identical, you can put it in `Resource`.  If the implementation diverges at some point then we can discuss what that might look like.  Sometimes, the child class version will call the parent class version first and then do a few additional things.  Other times, the child class version will completely replace the parent class version.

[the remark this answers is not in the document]
gonuke: This also makes me wonder if the `package_id_` should be a property of `Resource` that is inherited by both...?

src/context.cc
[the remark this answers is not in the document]
nuclearkatie: Yes, the archetypes generally only deal with the name. I feel like the numeric ids are generally only used for things internally, but there's no internal use of `GetRecipe`. I guess we could use the name, but then that seems like it negates the usefulness of having an id?

src/material.h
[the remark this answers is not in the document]
nuclearkatie: Wait, I remember. `Resource` doesn't know about the context. Only `Material` and `Product` do

gonuke (approved):

PR #1679, opened 2024-02-28
Remove compare-ref param

bennibbelink:
I think this will solve the base branch problem we've see with code coverage and PRs.  It seemed to work on my [fork](https://github.com/bennibbelink/cyclus/pull/16).  For some reason coveralls is referencing an [old build](https://coveralls.io/builds/64235833) on `main` which you can see in the coveralls UI.  Removing the explicit reference in the coveralls GitHub action seems to alleviate this issue

### DIFF
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -64,7 +64,6 @@ jobs:
         with:
           file: ${{ github.workspace }}/total_coverage.info
           format: lcov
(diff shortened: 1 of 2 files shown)


gonuke (approved):
Not sure how to test this, but it can't get worse ??

bennibbelink:
I think its working correctly... That comment shows a change of 0.0% 👍

PR #1683, opened 2024-03-01
Packaging with Resource Buffer Pop and Push

### LINKED ISSUE #1657: Add default behavior to Resource Buffer handling of packages
* [ ] strip package type on Push (becomes implicit type)
* [ ] apply package type on Pop for default package type = implicit packaging

### LINKED ISSUE #1658: Perform repackaging upon Resource Buffer pop
Repackage from implicit packaging to explicit packaging on pop

nuclearkatie:
- First pass at packaging interface with resource buffer

Closes #1657 and closes #1658
Linked to #1620

### DIFF
--- a/src/package.cc
+++ b/src/package.cc
@@ -21,6 +21,31 @@ Package::Ptr Package::Create(std::string name, double fill_min, double fill_max,
   return p;
 }
 
+double GetFillMass(Resource::Ptr r, Package::Ptr pkg) {
+  if (r->quantity() < pkg->fill_min()) {
+    // less than one pkg of material available
+    return 0;
+  }
+
+  std::vector<Resource::Ptr> rs;
+  Resource::Ptr r_pkgd;
+  double fill_mass;
+  if (pkg->strategy() == "first") {
+    fill_mass = pkg->fill_max();
+  } else if (pkg->strategy() == "equal") {
+    int num_min_fill = std::floor(r->quantity() / pkg->fill_min());
+    int num_max_fill = std::ceil(r->quantity() / pkg->fill_max());
+    if (num_min_fill >= num_max_fill) {
+      // all material can fit in a package
+      double fill_mass = r->quantity() / num_max_fill;
+    } else {
+      // some material will remain unpackaged, fill up as many max packages as possible
+      fill_mass = pkg->fill_max();
+    }
+  }
+  return fill_mass;
+}
+
 std::vector<typename T::Ptr> Package::Repackage(typename T::Ptr r,
                                                 Package::Ptr pkg) {
     std::vector<typename T::Ptr> rs_pkgd;
--- a/src/package.cc
+++ b/src/package.cc
@@ -20,6 +20,25 @@ Package::Ptr Package::Create(std::string name, double fill_min, double fill_max,
   Ptr p(new Package(name, fill_min, fill_max, strategy));
   return p;
 }
+
+std::vector<typename T::Ptr> Package::Repackage(typename T::Ptr r,
+                                                Package::Ptr pkg) {
+    std::vector<typename T::Ptr> rs_pkgd;
+    typename T::Ptr r_pkgd;
+    
+    double fill_mass = GetFillMass(r, pkg);
+    if (fill_mass ==0) {
+      return rs_pkgd;
+    }
+
+    while (r->quantity() > pkg->fill_min()) {
+      double pkg_fill = std::min(r->quantity(), fill_mass);
+      r_pkgd = boost::dynamic_pointer_cast<T>(r->ExtractRes(pkg_fill));
+      r_pkgd->ChangePackageId(pkg->id());
+      rs_pkgd.push_back(r_pkgd);
+    }
+    return rs_pkgd;
+  }
   
 Package::Package() : id_(next_id_++), fill_min_(0), fill_max_(std::numeric_limits<double>::max()) {}
 
--- a/src/package.cc
+++ b/src/package.cc
@@ -20,6 +20,10 @@ Package::Ptr Package::Create(std::string name, double fill_min, double fill_max,
   Ptr p(new Package(name, fill_min, fill_max, strategy));
   return p;
 }
+  
+Package::Package() : id_(next_id_++), fill_min_(0), fill_max_(std::numeric_limits<double>::max()) {}
+
+Package::Package(std::string name, double fill_min, double fill_max, std::string strategy) : name_(name), id_(next_id_++), fill_min_(fill_min), fill_max_(fill_max), strategy_(strategy) {}
 
 double GetFillMass(Resource::Ptr r, Package::Ptr pkg) {
   if (r->quantity() < pkg->fill_min()) {
@@ -45,28 +49,5 @@ double GetFillMass(Resource::Ptr r, Package::Ptr pkg) {
   }
   return fill_mass;
 }
-
-std::vector<typename T::Ptr> Package::Repackage(typename T::Ptr r,
-                                                Package::Ptr pkg) {
-    std::vector<typename T::Ptr> rs_pkgd;
-    typename T::Ptr r_pkgd;
-    
-    double fill_mass = GetFillMass(r, pkg);
-    if (fill_mass ==0) {
-      return rs_pkgd;
-    }
-
-    while (r->quantity() > pkg->fill_min()) {
-      double pkg_fill = std::min(r->quantity(), fill_mass);
-      r_pkgd = boost::dynamic_pointer_cast<T>(r->ExtractRes(pkg_fill));
-      r_pkgd->ChangePackageId(pkg->id());
-      rs_pkgd.push_back(r_pkgd);
-    }
-    return rs_pkgd;
-  }
-  
-Package::Package() : id_(next_id_++), fill_min_(0), fill_max_(std::numeric_limits<double>::max()) {}
-
-Package::Package(std::string name, double fill_min, double fill_max, std::string strategy) : name_(name), id_(next_id_++), fill_min_(fill_min), fill_max_(fill_max), strategy_(strategy) {}
 
 } // namespace cyclus
--- a/src/package.cc
+++ b/src/package.cc
@@ -5,6 +5,31 @@ namespace cyclus {
 
 int Package::next_id_ = 1;
 
+double Package::GetFillMass(Resource::Ptr r) {
+  if (r->quantity() < fill_min_) {
+    // less than one pkg of material available
+    return 0;
+  }
+
+  std::vector<Resource::Ptr> rs;
+  Resource::Ptr r_pkgd;
+  double fill_mass;
+  if (strategy_ == "first") {
+    fill_mass = fill_max_;
+  } else if (strategy_ == "equal") {
+    int num_min_fill = std::floor(r->quantity() / fill_min_);
+    int num_max_fill = std::ceil(r->quantity() / fill_max_);
+    if (num_min_fill >= num_max_fill) {
+      // all material can fit in a package
+      double fill_mass = r->quantity() / num_max_fill;
+    } else {
+      // some material will remain unpackaged, fill up as many max packages as possible
+      fill_mass = fill_max_;
+    }
+  }
+  return fill_mass;
+}
+
 Package::Ptr Package::Create() {
   Ptr p(new Package());
   return p;
@@ -24,30 +49,5 @@ Package::Ptr Package::Create(std::string name, double fill_min, double fill_max,
 Package::Package() : id_(next_id_++), fill_min_(0), fill_max_(std::numeric_limits<double>::max()) {}
 
 Package::Package(std::string name, double fill_min, double fill_max, std::string strategy) : name_(name), id_(next_id_++), fill_min_(fill_min), fill_max_(fill_max), strategy_(strategy) {}
-
-double GetFillMass(Resource::Ptr r, Package::Ptr pkg) {
-  if (r->quantity() < pkg->fill_min()) {
-    // less than one pkg of material available
-    return 0;
-  }
-
-  std::vector<Resource::Ptr> rs;
-  Resource::Ptr r_pkgd;
-  double fill_mass;
-  if (pkg->strategy() == "first") {
-    fill_mass = pkg->fill_max();
-  } else if (pkg->strategy() == "equal") {
-    int num_min_fill = std::floor(r->quantity() / pkg->fill_min());
-    int num_max_fill = std::ceil(r->quantity() / pkg->fill_max());
-    if (num_min_fill >= num_max_fill) {
-      // all material can fit in a package
-      double fill_mass = r->quantity() / num_max_fill;
-    } else {
-      // some material will remain unpackaged, fill up as many max packages as possible
-      fill_mass = pkg->fill_max();
-    }
-  }
-  return fill_mass;
-}
 
 } // namespace cyclus
--- a/src/package.cc
+++ b/src/package.cc
@@ -30,6 +30,25 @@ double Package::GetFillMass(Resource::Ptr r) {
   return fill_mass;
 }
 
+template <class T>
+std::vector<typename T::Ptr> Package::PackageResource(typename T::Ptr r) {
+  std::vector<typename T::Ptr> rs_pkgd;
+  typename T::Ptr r_pkgd;
+  
+  double fill_mass = GetFillMass(r);
+  if (fill_mass ==0) {
+    return rs_pkgd;
+  }
+
+  while (r->quantity() > fill_min_) {
+    double pkg_fill = std::min(r->quantity(), fill_mass);
+    r_pkgd = boost::dynamic_pointer_cast<T>(r->ExtractRes(pkg_fill));
+    r_pkgd->ChangePackageId(id_);
+    rs_pkgd.push_back(r_pkgd);
+  }
+  return rs_pkgd;
+}
+
 Package::Ptr Package::Create() {
   Ptr p(new Package());
   return p;
--- a/src/package.h
+++ b/src/package.h
@@ -37,23 +37,7 @@ class Package {
     /// resource cannot be packaged using the given packaging strategy and
     /// restrictions, the remainder is left in the resource object. 
     template <class T>
-    std::vector<typename T::Ptr> PackageResource(typename T::Ptr r) {
-      std::vector<typename T::Ptr> rs_pkgd;
-      typename T::Ptr r_pkgd;
-      
-      double fill_mass = GetFillMass(r);
-      if (fill_mass ==0) {
-        return rs_pkgd;
-      }
-
-      while (r->quantity() > fill_min_) {
-        double pkg_fill = std::min(r->quantity(), fill_mass);
-        r_pkgd = boost::dynamic_pointer_cast<T>(r->ExtractRes(pkg_fill));
-        r_pkgd->ChangePackageId(id_);
-        rs_pkgd.push_back(r_pkgd);
-      }
-      return rs_pkgd;
-    };
+    std::vector<typename T::Ptr> PackageResource(typename T::Ptr r);
 
     // create a new package type with default values 
     static Ptr Create();
--- a/src/package.h
+++ b/src/package.h
@@ -48,7 +48,14 @@ class Package {
 
 /// Returns optimal fill mass for a resource to be packaged. Can be used
 /// to determine how to respond to requests for material, and to actually
-/// package and send off trades
+/// package and send off trades.
+/// Packaging strategy "first" simply fills the packages one by one to the
+/// maximum fill. Therefore, it should always try to max fill.
+/// Packaging strategy "equal" tries to fill all packages to the same mass.
+/// This tries to find the optimal number and fill mass of packages given the
+/// packaging limitations. It does this by calculating bounding fills, 
+/// floor(quantity/fill_min) and ceiling(quantity/fill_max). 
+/// There might be a scenario w
 double GetFillMass(Resource::Ptr r, Package::Ptr pkg);
 
 /// Repackages a single resource into a package. If some quantity of the 
--- a/src/package.h
+++ b/src/package.h
@@ -55,7 +55,11 @@ class Package {
 /// This tries to find the optimal number and fill mass of packages given the
 /// packaging limitations. It does this by calculating bounding fills, 
 /// floor(quantity/fill_min) and ceiling(quantity/fill_max). 
-/// There might be a scenario w
+/// There might be a scenario where there is no solution, i.e. an integer
+/// number of packages cannot be filled with no remainder. In this case,
+/// the most effective fill strategy is to fill to the max. Numeric example:
+/// quantity = 5, fill_min = 3, fill_max = 4. num_min_fill = floor(5/3) = 1,
+/// num_max_fill = ceil(5/4) = 2. num_min_fill < num_max_fill, so fill to max.
 double GetFillMass(Resource::Ptr r, Package::Ptr pkg);
 
 /// Repackages a single resource into a package. If some quantity of the 
--- a/src/package.h
+++ b/src/package.h
@@ -5,6 +5,7 @@
 #include <string>
 #include <vector>
 #include <boost/shared_ptr.hpp>
+#include "resource.h"
 
 namespace cyclus {
 
@@ -19,6 +20,12 @@ class Package {
 
     // create a new package type
     static Ptr Create(std::string name, double fill_min, double fill_max, std::string strategy);
+
+    /// Given a single resource and a package type, returns optimal fill mass
+    /// for the resource to be packaged. Can be used to determine how to   
+    /// respond to requests for material, and to actually package and send off
+    /// trades
+    double GetFillMass(Resource::Ptr r, Package::Ptr pkg);
 
     /// Repackages a single resource into a package. If some quantity of the 
     /// resource cannot be packaged, the remainder is left in the resource
--- a/src/package.h
+++ b/src/package.h
@@ -3,6 +3,7 @@
 
 #include <limits>
 #include <string>
+#include <vector>
 #include <boost/shared_ptr.hpp>
 
 namespace cyclus {
@@ -18,6 +19,11 @@ class Package {
 
     // create a new package type
     static Ptr Create(std::string name, double fill_min, double fill_max, std::string strategy);
+
+    /// Repackages a single resource into a package. If some quantity of the 
+    /// resource cannot be packaged, the remainder is left in the resource
+    /// object. 
+    std::vector<typename T::Ptr> Repackage(typename T::Ptr r, Package::Ptr pkg);
 
     // returns package id
     int id() const { return id_; }
--- a/src/package.h
+++ b/src/package.h
@@ -16,6 +16,45 @@ class Package {
   public:
     typedef boost::shared_ptr<Package> Ptr;
 
+    /// Returns optimal fill mass for a resource to be packaged. Can be used
+    /// to determine how to respond to requests for material, and to actually
+    /// package and send off trades.
+    /// Packaging strategy "first" simply fills the packages one by one to the
+    /// maximum fill. Therefore, it should always try to max fill.
+    /// Packaging strategy "equal" tries to fill all packages to the same mass.
+    /// This tries to find the optimal number and fill mass of packages given
+    /// the packaging limitations. It does this by calculating bounding fills, 
+    /// floor(quantity/fill_min) and ceiling(quantity/fill_max). 
+    /// There might be a scenario where there is no solution, i.e. an integer
+    /// number of packages cannot be filled with no remainder. In this case,
+    /// the most effective fill strategy is to fill to the max. Numeric example:
+    /// quantity = 5, fill_min = 3, fill_max = 4. num_min_fill = floor(5/3) = 1,
+    /// num_max_fill = ceil(5/4) = 2. num_min_fill < num_max_fill, so fill to
+    /// the max.
+    double GetFillMass(Resource::Ptr r);
+
+    /// Repackages a single resource into a package. If some quantity of the 
+    /// resource cannot be packaged using the given packaging strategy and
+    /// restrictions, the remainder is left in the resource object. 
+    template <class T>
+    std::vector<typename T::Ptr> PackageResource(typename T::Ptr r) {
+      std::vector<typename T::Ptr> rs_pkgd;
+      typename T::Ptr r_pkgd;
+      
+      double fill_mass = GetFillMass(r);
+      if (fill_mass ==0) {
+        return rs_pkgd;
+      }
+
+      while (r->quantity() > fill_min_) {
+        double pkg_fill = std::min(r->quantity(), fill_mass);
+        r_pkgd = boost::dynamic_pointer_cast<T>(r->ExtractRes(pkg_fill));
+        r_pkgd->ChangePackageId(id_);
+        rs_pkgd.push_back(r_pkgd);
+      }
+      return rs_pkgd;
+    };
+
     // create a new package type with default values 
     static Ptr Create();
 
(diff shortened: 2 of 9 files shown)


src/toolkit/res_manip.h
```diff
@@ -30,6 +30,8 @@ std::vector<Resource::Ptr> ResCast(std::vector<Product::Ptr> rs);
 /// basically a no-op that enables ResBuf to work for plain Resource types.
 std::vector<Resource::Ptr> ResCast(std::vector<Resource::Ptr> rs);
 
+double GetFillMass(Resource::Ptr r, Package::Ptr pkg);
```
gonuke, line 33: Can you add some docs about what this is meant to do?

src/toolkit/res_manip.cc
```diff
@@ -65,6 +68,34 @@ std::vector<Resource::Ptr> ResCast(std::vector<Resource::Ptr> rs) {
...
 }
 
+double GetFillMass(Resource::Ptr r, Package::Ptr pkg) {
+  if (r->quantity() == 0) {
+    throw Error("cannot fill zero resources into a package");
+  }
+  else if (r->quantity() < pkg->fill_min()) {
+    // less than one pkg of material available
+    return 0;
+  }
```
gonuke, line 78: I'm guessing we need to treat these two cases differently because we can have `fill_min == 0`.

thought: If we force `fill_min` to always bigger than `eps` than we can treat `quantity == 0` as the same as `quantity < fill_min`?

I'm not sure if that helps/hurts us down the line with material accountancy?

Alternatively, we can just return 0 in both cases and not throw an error?

src/toolkit/res_manip.cc
```diff
@@ -65,6 +68,34 @@ std::vector<Resource::Ptr> ResCast(std::vector<Resource::Ptr> rs) {
...
+
+  std::vector<Resource::Ptr> rs;
+  Resource::Ptr r_pkgd;
+  double fill_mass;
+  if (pkg->strategy() == "first") {
+    fill_mass = pkg->fill_max();
+  } else if (pkg->strategy() == "equal") {
+    int num_min_fill = std::floor(r->quantity() / pkg->fill_min());
+    int num_max_fill = std::ceil(r->quantity() / pkg->fill_max());
+    if (num_min_fill >= num_max_fill) {
```
gonuke, line 88: This is really slick and compact logic, but I confess that it's not obvious to me when this will NOT be true? Maybe we should put an example in a comment somewhere?

src/toolkit/res_buf.h
```diff
@@ -255,6 +284,8 @@ class ResBuf {
     } else if (rs_present_.count(m) == 1) {
       throw KeyError("duplicate resource push attempted");
     }
+    // strip package id and set as default
+    m->ChangePackageId();
```
gonuke, line 288: Don't we want to allow packaging to be preserved in some ResBuf's. Storage should often keep things packaged?  Is this a property of the ResBuf?

gonuke (changes requested):
A few thoughts, changes, questions.  We should discuss more at our group meeting.

src/toolkit/res_manip.cc
nuclearkatie: Yeah, it occurs when there's leftover material that cannot be packaged even keeping every package at `fill_min`

Example: 
`quantity` = 5
`fill_min` = 3
`fill_max` = 4

`num_min_fill` = floor(5/3) = 1
`num_max_fill` = ceil(5/4) = 2

Basically, there's not enough to fill two packages at min fill mas, but you have more than one max filled package worth of material. The solution is you fill one package, and then have to return the remaining (in this case, 1) to the buffer-- it cannot be packaged given the packaging strategy

src/toolkit/res_buf.h
nuclearkatie: That's fair, will remove

nuclearkatie: Though we did agree (from #1657 ) "strip package type on Push (becomes implicit type)"

So maybe there should be a `keep_packaging` bool?

gonuke: Or the opposite `bulk_storage` bool?

bennibbelink:
@nuclearkatie this is what I've seen playing around with the tests.  If you open a python shell and execute the commands from the test everything looks good and would pass.  Additionally - if you modify line 1190 of `cycpp_test.py` to `stderr=None` the test passes.  It appears as if all the SyntaxWarnings from the earlier tests in the file are being piped to the Popen object which is causing the assertion to fail.  

I will try to find a fix, and I'll try to resolve the other warnings if possible.  Stay tuned for a PR

Not sure why this isn't happening in the `apt` builds 🤔 maybe has to do with the python versions they use

bennibbelink:
Pretty sure the failures were simply due to the SyntaxWarnings from running `cycpp.py`.  #1684  resolves the warnings.  These warnings have been around for a while... I'm not sure why it is just now causing failure

gonuke:
This is passing tests now @nuclearkatie

gonuke:
If you convert this from draft, I can do a final review to merge

src/toolkit/res_buf.h
```diff
@@ -163,6 +163,35 @@ class ResBuf {
     return Pop(qty);
   }
 
+  /// Pops the specified amount of material, re-packed into the the 
+  /// package type provided. Returns any remaining material that doesn't
+  /// fit in a package per the packaging strategy and limits as a 
+  /// separate vector with default packaging
+  std::vector<typename T::Ptr> PopPackaged(double qty, Package::Ptr pkg) {
+    typename T::Ptr r = Pop(qty);
```
gonuke, line 171: Conceptual: this `Pop` will squash resources to get this `qty`.  What if it's already packaged in this `pkg` type and we could just `Pop` multiple items with `PopN`?  Is there a way to test that?

gonuke: What if we always `PopN` to get enough `qty` and then have a `repackage` method that takes some number of packaged materials and puts them into a different number of different packaged materials?  The `repackage` method could be something that we look to reuse in other places

src/toolkit/res_buf.h
```diff
@@ -243,7 +272,7 @@ class ResBuf {
   ///
   /// @throws KeyError the resource object to be pushed is already present
   /// in the buffer.
-  void Push(Resource::Ptr r) {
+  void Push(Resource::Ptr r, bool bulk_storage = true) {
```
gonuke, line 275: Given thoughts about a true bulk storage `ResBuf` that squashes everything upon `Push`, should we call this `unpacakged` instead of `bulk_storage`?

src/toolkit/res_manip.cc
```diff
@@ -25,6 +27,7 @@ Material::Ptr Squash(std::vector<Material::Ptr> ms) {
   for (int i = 1; i < ms.size(); ++i) {
     m->Absorb(ms[i]);
   }
+  m->ChangePackageId();
```
gonuke, line 30: Is it possible to squash something into a package?  What if you have mat A that is filling the min of a its package and then you squash mat B that will fit in that package?  Can we/should we avoid removing the package?

gonuke (changes requested):
A few more comments, including a big conceptual one

src/toolkit/res_manip.cc
nuclearkatie: The way I have it is that squash always removes packaging. I don't feel strongly that we should keep the packaging, but I will change it if you think that's important

src/toolkit/res_buf.h
```diff
@@ -101,16 +101,15 @@ class ResBuf {
...
   /// Resources are split if necessary in order to pop the exact quantity
   /// requested (within eps_rsrc()).  Resources are retrieved in the order they
-  /// were pushed (i.e. oldest first) and are squashed into a single object
-  /// when returned.
+  /// were pushed (i.e. oldest first).
   ///
   /// @throws ValueError the specified pop quantity is larger than the
   /// buffer's current inventory.
-  typename T::Ptr Pop(double qty) {
+  std::vector<typename T::Ptr> PopVector(double qty) {
```
gonuke, line 112: Do you think it would confusing to name this `PopN(double qty)`.  It behaves a lot like `PopN` except it chooses how many items by `qty` instead of `count`?

src/toolkit/res_buf.h
```diff
@@ -163,6 +172,35 @@ class ResBuf {
     return Pop(qty);
   }
 
+  /// Pops the specified amount of material, re-packed into the the 
+  /// package type provided. Returns any remaining material that doesn't
+  /// fit in a package per the packaging strategy and limits as a 
+  /// separate vector with default packaging
```
gonuke, line 178:
```suggestion
  /// separate vector with default packaging
  /// If it's not possible to fill the package
  /// it returns an empty vector.
```

src/toolkit/res_buf.h
```diff
@@ -163,6 +163,35 @@ class ResBuf {
...
+      double pkg_fill = std::min(r->quantity(), fill_mass);
+      r_pkgd = boost::dynamic_pointer_cast<T>(r->ExtractRes(pkg_fill));
+      r_pkgd->ChangePackageId(pkg->id());
+      rs_pkgd.push_back(r_pkgd);
+    }
+    // push leftover material back into the buffer
+    double remaining = r->quantity() - r_pkgd->quantity();
+    if (remaining > 0) {
+      rs_.push_front(boost::dynamic_pointer_cast<T>(r->ExtractRes(remaining)));
+    }
```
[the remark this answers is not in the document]
gonuke: ```std::vector<typename T::Ptr> Package::Repackage(typename T::Ptr r, Package::Ptr pkg)```

Would take a pointer to a resource, r, and return a vector of resources appropriately packaged resources and modify r to leave any excess behind.

gonuke (changes requested):
A few comments before the plane takes off....

src/toolkit/res_buf.h
```diff
@@ -189,6 +227,12 @@ class ResBuf {
     return rs;
   }
 
+  /// Pops the specified number of resource objects from the buffer and returns
+  /// them as a single, squashed object.
+  typename T::Ptr PopNSquash(int n) {
+    return Squash(PopN(n));
+  }
+
```
gonuke, line 235: We discussed leaving this to the user/developer - they could just call `PopN` and take charge of squashing of they want to??

src/toolkit/res_buf.h
```diff
@@ -243,7 +287,7 @@ class ResBuf {
   ///
   /// @throws KeyError the resource object to be pushed is already present
   /// in the buffer.
-  void Push(Resource::Ptr r) {
+  void Push(Resource::Ptr r, bool unpackaged = true) {
```
gonuke, line 290: Should this be a property of the `Push` or of the `ResBuf` itself?

gonuke (changes requested):
A couple of additional comments

src/toolkit/res_buf.h
ahnaf-tahmid-chowdhury: A clearer and less confusing name could be something like `PopByQuantity(double qty)` or `PopAmount(double qty)` to make it clear that it operates based on a quantity rather than a count.

nuclearkatie: @gonuke I personally think re-using `PopN` with a quantity would be confusing. 

@ahnaf-tahmid-chowdhury I also don't want it to confuse people with regular `Pop`, which is in heavy use already. I expect that will continue to be the most frequently used way to pop off of a ResBuf. To me, `PopByQuantity` or `PopAmount` doesn't distinguish it from `Pop`, it actually makes it more likely that people get confused. While the code is now written that this function is the workhorse of popping, in my mind it's a sub-type of Pop, just pop without squash.

src/toolkit/res_buf.h
nuclearkatie: Got it, will remove

src/toolkit/res_buf.h
ahnaf-tahmid-chowdhury: `PopWithoutSquash`/`PopNonSquashed`/`SimplePop`/`BasicPop`? Just some suggestions.

Ultimately, the goal is to choose a name that accurately describes the function's behavior while minimizing confusion with existing methods like Pop.

src/toolkit/res_manip.cc
```diff
@@ -65,6 +68,31 @@ std::vector<Resource::Ptr> ResCast(std::vector<Resource::Ptr> rs) {
...
+    int num_max_fill = std::ceil(r->quantity() / pkg->fill_max());
+    if (num_min_fill >= num_max_fill) {
+      // all material can fit in a package
+      double fill_mass = r->quantity() / num_max_fill;
+    } else {
+      // some material will remain unpackaged, fill up as many max packages as possible
+      fill_mass = pkg->fill_max();
+    }
+  }
+  return fill_mass;
```
ahnaf-tahmid-chowdhury, line 93: I suppose, this comment will help clarify the scenario when the condition num_min_fill >= num_max_fill is not true.

```suggestion
  // Handle the scenario when there's leftover material that cannot be packaged
  // due to not enough material for minimum fill, but more than one max filled package worth of material.
  // In this case, fill one package up all the way and return the remaining to the buffer.
  return fill_mass;
```

src/toolkit/res_buf.h
nuclearkatie: I'll merge in your new `is_bulk` bool from #1687 and use that

src/toolkit/res_manip.cc
nuclearkatie: Hi @ahnaf-tahmid-chowdhury, thanks for your comment! This method doesn't actually handle the repackaging, it only calculates how much material will go in the package(s), so this description isn't quite right for what this part of the code is doing. Also, a comment describing the functionality should be in the header file at the top, rather than at the bottom of the source. 

I'm going to commit your suggestion because I appreciate you trying help clarify, and I'll make the changes I just described in my next PR so there's better description in the header for `GetFillMass` and `Repackage`

src/toolkit/res_buf.h
gonuke: I think this is maybe slightly different. I can imagine one mode that forces items to become unpackaged but keeps them discrete, and another that squashes them all into a single unpackaged material. Mine is the latter.

gonuke: Rereading your comment, maybe you were already planning on both...

src/toolkit/res_buf.h
```diff
@@ -163,6 +172,37 @@ class ResBuf {
...
+    }
+    // push leftover material back into the buffer
+    double remaining = r->quantity() - r_pkgd->quantity();
+    if (remaining > 0) {
+      rs_.push_front(boost::dynamic_pointer_cast<T>(r->ExtractRes(remaining)));
+    }
+
+    return rs_pkgd;
+  }
+
```
ahnaf-tahmid-chowdhury, line 205:
```suggestion
class Package {
public:

  /**
   * @brief Repackages a given resource into the specified package.
   *
   * This function takes a resource and a package as input, and repackages the resource into the package. 
   * The function returns a vector of packaged resources. If there is any remaining quantity after repackaging, 
   * it is left in the original resource. If the resource is fully repackaged, the original resource is set to null.
   *
   * @param r A shared pointer to the resource to be repackaged.
   * @param pkg A shared pointer to the package into which the resource will be repackaged.
   * @return A vector of packaged resources.
   */

  std::vector<typename T::Ptr> repackage(typename T::Ptr r, Package::Ptr pkg) {
    std::vector<typename T::Ptr> rs_pkgd;
    typename T::Ptr r_pkgd;

    double fill_mass = GetFillMass(r, pkg);
    if (fill_mass == 0) {
      return rs_pkgd;
    }

    while (r->quantity() > pkg->fill_min()) {
      double pkg_fill = std::min(r->quantity(), fill_mass);
      r_pkgd = boost::dynamic_pointer_cast<T>(r->ExtractRes(pkg_fill));
      r_pkgd->ChangePackageId(pkg->id());
      rs_pkgd.push_back(r_pkgd);
    }
    // push leftover material back into the buffer
    double remaining = r->quantity() - r_pkgd->quantity();
    if (remaining > 0) {
      rs_.push_front(boost::dynamic_pointer_cast<T>(r->ExtractRes(remaining)));
    }

    return rs_pkgd;
  }
};


/**
 * @brief Pops a specified amount of material and repackages it.
 *
 * This function pops a specified amount of material from a container and repackages it into the provided package type.
 * The function returns a vector of packaged resources. If there is any remaining quantity after repackaging, it is left in the original resource.
 * If the resource is fully repackaged, the original resource is set to null.
 *
 * @param qty The quantity of material to be popped and repackaged.
 * @param pkg A shared pointer to the package into which the material will be repackaged.
 * @return A vector of packaged resources.
 */
std::vector<typename T::Ptr> PopPackaged(double qty, Package::Ptr pkg) {
  typename T::Ptr r = Pop(qty);
  
  // Use the repackage function to do the packaging
  std::vector<typename T::Ptr> rs_pkgd = pkg->repackage(r, pkg);
  
  return rs_pkgd;
}
```

ahnaf-tahmid-chowdhury (commented):
This repackage function can now be reused elsewhere in the code by creating an instance of the Package class and calling the repackage method on it.

src/toolkit/res_buf.h
gonuke: Also, the `Package` class already exists elsewhere in the project as part of @nuclearkatie 's PhD research

ahnaf-tahmid-chowdhury: And, sorry, @nuclearkatie. I truly respect your efforts and believe that you may have your own ideas in progress. I just added this suggestion because I've found that I've already done similar things before. nvm

src/toolkit/res_buf.h
nuclearkatie: Yeah, i meant doing both!

src/package.h
```diff
@@ -43,6 +46,33 @@ class Package {
...
+/// Returns optimal fill mass for a resource to be packaged. Can be used
+/// to determine how to respond to requests for material, and to actually
+/// package and send off trades
+double GetFillMass(Resource::Ptr r, Package::Ptr pkg);
+
+/// Repackages a single resource into a package. If some quantity of the 
+/// resource cannot be packaged using the given packaging strategy and
+/// restrictions, the remainder is left in the resource object. 
+template <class T>
+std::vector<typename T::Ptr> Repackage(typename T::Ptr r, Package::Ptr pkg) {
```
abachma2, line 58: Is there any reason you have this in the .h file instead of the .cc?

src/package.cc
```diff
@@ -25,4 +25,29 @@ Package::Package() : id_(next_id_++), fill_min_(0), fill_max_(std::numeric_limit
...
+  Resource::Ptr r_pkgd;
+  double fill_mass;
+  if (pkg->strategy() == "first") {
+    fill_mass = pkg->fill_max();
+  } else if (pkg->strategy() == "equal") {
+    int num_min_fill = std::floor(r->quantity() / pkg->fill_min());
+    int num_max_fill = std::ceil(r->quantity() / pkg->fill_max());
+    if (num_min_fill >= num_max_fill) {
+      // all material can fit in a package
+      double fill_mass = r->quantity() / num_max_fill;
```
abachma2, line 44: Can you walk me through the logic of this one? Starting with what the `equal` strategy is supposed to be.

src/resource.h
```diff
@@ -79,6 +81,16 @@ class Resource {
   /// @return a new resource object with same state id and quantity == quantity
   virtual Ptr ExtractRes(double quantity) = 0;
 
+  /// To enable the Decay method to be called on any child resource, define
+  /// a null op Decay method here.
+  /// @param curr_time the current time for the decay oepration
+  virtual void Decay(int curr_time) { throw Error("cannot decay resource type " + this->type()); };
```
abachma2, line 87: In `material.h` you applied a default value of -1. Did you mean to do that in this file too?

src/resource.h
```diff
@@ -79,6 +81,16 @@ class Resource {
...
+
+  /// To enable the Absorb method to be called on any child resource, define
+  /// a null op Absorb method here.
```
abachma2, line 90: `op`? Is that a typo in the documentation?

src/toolkit/res_buf.h
```diff
@@ -101,16 +101,15 @@ class ResBuf {
   /// Returns true if there are no resources in the buffer.
   inline bool empty() const { return rs_.empty(); }
 
-  /// Pops and returns the specified quantity from the buffer as a single
-  /// resource object.
+  /// Pops and returns the specified quantity from the buffer as vector of
```
abachma2, line 104:
```suggestion
  /// Pops and returns the specified quantity from the buffer as a vector of 
```

src/toolkit/res_buf.h
```diff
@@ -233,13 +250,13 @@ class ResBuf {
...
   }
 
-  /// Pushes a single resource object to the buffer.
-  /// Resource objects are never combined in the buffer; they are stored as
-  /// unique objects. The resource object is only pushed to the buffer if it
-  /// does not cause the buffer to exceed its capacity.
+  /// Pushes a single resource object to the buffer. If not classified as a bulk
+  /// storage buffer, resource objects are not combined in the buffer; they
+  /// are stored as unique objects. The resource object is only pushed to the
+  /// buffer if it does not cause the buffer to exceed its capacity.
```
abachma2, line 256:
```suggestion
  /// Pushes a single resource object to the buffer. Unless classified as a 
  // bulk storage buffer, the objects
  /// are stored as unique objects. If classified as a bulk storage buffer, then the 
  // resource objects are combined. The resource object is only pushed to the
  /// buffer if it does not cause the buffer to exceed its capacity.
```
There were quite a few negatives in the same sentence that may get confusing.

src/toolkit/res_buf.h
```diff
@@ -256,23 +273,33 @@ class ResBuf {
...
   }
 
-  /// Pushes one or more resource objects (as a std::vector) to the buffer.
-  /// Resource objects are never squashed in the buffer; they are stored as
-  /// unique objects. The resource objects are only pushed to the buffer if
-  /// they do not cause the buffer to exceed its capacity; otherwise none of the
-  /// given resource objects are added to the buffer.
+  /// Pushes one or more resource objects (as a std::vector) to the buffer. If
+  /// not classified as a bulk storage buffer, resource objects are not
+  /// squashed in the buffer; they are stored as unique objects. The resource
```
abachma2, line 293: See my previous suggested change about too many negatives.

src/xml_file_loader.cc
```diff
@@ -339,7 +339,7 @@ void XMLFileLoader::LoadPackages() {
     std::string name = cyclus::OptionalQuery<std::string>(qe, "name", "default");
     CLOG(LEV_DEBUG3) << "loading package: " << name;
     
-    double fill_min = cyclus::OptionalQuery<double>(qe, "fill_min", 0);
+    double fill_min = cyclus::OptionalQuery<double>(qe, "fill_min", eps());
```
abachma2, line 342: What is this `eps()` method?

tests/toolkit/res_buf_tests.cc
```diff
@@ -12,20 +13,20 @@ namespace toolkit {
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 // - - - - - - Getters, Setters, and Property changers - - - - - - - - - - - -
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-TEST_F(ResBufTest, set_capacity_ExceptionsEmpty) {
+TEST_F(ProductBufTest, set_capacity_ExceptionsEmpty) {
```
abachma2, line 16: So in this file you renamed all of the tests to Product instead of Res? Why? Do we still need those ResBuf tests? And if not, it probably makes sense to rename the file to `product_buf_tests`.

abachma2 (changes requested):
This seems like an interesting PR, allowing some bulk material handling in Cyclus. I have quite a few comments, either to clarify things in the code or to clarify them for me.

nuclearkatie:
Hi @abachma2, thanks for your comments. Last week I tried to rebase on @gonuke 's PR #1687, which made some messy merges and now some of his code from that PR is showing up here as changes I made. I'm going to tag him on the comments on his changes, which are all the bulk stuff.

Also, I'm on vacation today and tomorrow so I won't be fixing this on my end until later this week

tests/toolkit/res_buf_tests.cc
nuclearkatie: @gonuke this is from #1687

src/package.h
nuclearkatie: Not really

src/material.h
```diff
@@ -136,7 +136,9 @@ class Material: public Resource {
   /// not result in an updated material composition.  Does nothing if the
   /// simulation decay mode is set to "never" or none of the nuclides' decay
   /// constants are significant with respect to the time delta.
-  void Decay(int curr_time);
+  /// @param curr_time current time to use for the decay calculation 
+  ///        (default: -1 forces the decay to the context's current time)
```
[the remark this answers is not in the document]
nuclearkatie: @gonuke this is from #1687

src/material.h
```diff
@@ -121,7 +121,7 @@ class Material: public Resource {
                   double threshold = eps_rsrc());
 
   /// Combines material mat with this one.  mat's quantity becomes zero.
-  void Absorb(Ptr mat);
+  virtual void Absorb(Ptr mat);
```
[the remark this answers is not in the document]
nuclearkatie: @gonuke this is from #1687

src/context.cc
```diff
@@ -211,13 +211,17 @@ Package::Ptr Context::GetPackageByName(std::string name) {
...
+    throw ValueError("Invalid package id " + std::to_string(id));
+  }
   // iterate through the list of packages to get the one package with the correct id
   std::map<std::string, Package::Ptr>::iterator it;
   for (it = packages_.begin(); it != packages_.end(); ++it) {
     if (it->second->id() == id) {
       return it->second;
     }
   }
+  throw ValueError("Invalid package id " + std::to_string(id));
```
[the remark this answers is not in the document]
nuclearkatie: The facility doesn't hold package ids, the context does. But yes this throws an error if an agent asks for the packaging info for a package that doesn't exist

src/resource.h
nuclearkatie: @gonuke this is from #1687

src/resource.h
nuclearkatie: @gonuke this is from #1687

src/toolkit/res_buf.h
nuclearkatie: @gonuke this is from #1687

src/xml_file_loader.cc
nuclearkatie: This is the epsilon that either a user defines in their input file or a default is applied

src/material.h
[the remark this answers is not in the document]
gonuke: (We'll have to investigate how to unclog this rebase tangle)

`virtual` declares that this is a function that can/will be overwritten by a child class from a parent class - in this case in the child class.

src/material.h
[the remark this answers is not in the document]
gonuke: Currently, a user/developer can request a decay to any time of interest.  If they give no time, it will be set to -1 and interpreted as the current time.

src/package.cc
gonuke: `equal` is a strategy that attempts to fill as many packages as necessary with an equal inventory.  This strategy does have the case where distributing a certain quantity equally can result in underfilling the packages and using one fewer packages is not enough inventory for all the quantity requested.  In response to a similar comment by me @nuclearkatie gave an example.  It may be useful to have that example in a comment.

src/resource.h
gonuke: It doesn't matter here, because if the virtual function is not overwritten by a child class, it will invoke this version and throw an exception.

src/resource.h
gonuke: `null op` is a common term to describe a function that does nothing.  That said, I wrote this comment when the function really did do nothing (it just returned) so this comment should probably be udpated.

src/package.cc
```diff
@@ -25,4 +25,29 @@ Package::Package() : id_(next_id_++), fill_min_(0), fill_max_(std::numeric_limit
...
+  if (r->quantity() < pkg->fill_min()) {
+    // less than one pkg of material available
+    return 0;
+  }
+
+  std::vector<Resource::Ptr> rs;
+  Resource::Ptr r_pkgd;
+  double fill_mass;
+  if (pkg->strategy() == "first") {
+    fill_mass = pkg->fill_max();
```
gonuke, line 38: What if the requested quantity is more than N packages and the extra increment is less than the minimum fill?

nuclearkatie: Right now `first` strategy will just return N full packages. When designing this I did think that maybe we want a first-relaxed strategy that then goes back and tries to remove some material (up to `max`-`min`) from packages one by one until it can fill N+1 packages, but right now that's not implemented

src/toolkit/res_buf.h
```diff
@@ -163,6 +172,14 @@ class ResBuf {
     return Pop(qty);
   }
 
+  /// Pops the specified amount of material, re-packed into the the
```
gonuke, line 175: I'm not sure what this comment refers to now? Is it in the wrong place or did we delete the function that this was about?

src/toolkit/res_buf.h
```diff
@@ -255,9 +272,15 @@ class ResBuf {
...
       throw KeyError("duplicate resource push attempted");
     }
+
     if (!is_bulk_  || rs_.size() == 0) {
+      // strip package id and set as default
+      m->ChangePackageId();
       rs_.push_back(m);
       rs_present_.insert(m);
+    } else if (unpackaged_) {
+      m->ChangePackageId();
```
gonuke, line 282: I don't think we need to unpackage this if it's being absorbed, which also means this `else if` isn't needed.

src/toolkit/res_buf.h
```diff
@@ -305,8 +328,12 @@ class ResBuf {
 
     for (int i = 0; i < rss.size(); i++) {
       if (!is_bulk_ || rs_.size() == 0) {
+        rss[i]->ChangePackageId();
         rs_.push_back(rss[i]);
         rs_present_.insert(rss[i]);
+      } else if (unpackaged_) {
+        rss[i]->ChangePackageId();
```
gonuke, line 335: unpackaging isn't necessary here nor is this block needed.

src/toolkit/res_buf.h
```diff
@@ -305,8 +328,12 @@ class ResBuf {
 
     for (int i = 0; i < rss.size(); i++) {
       if (!is_bulk_ || rs_.size() == 0) {
+        rss[i]->ChangePackageId();
```
gonuke, line 331: Don't wee need to check `unpackaged_` here?

src/toolkit/res_buf.h
```diff
@@ -255,9 +272,15 @@ class ResBuf {
     } else if (rs_present_.count(m) == 1) {
       throw KeyError("duplicate resource push attempted");
     }
+
     if (!is_bulk_  || rs_.size() == 0) {
+      // strip package id and set as default
+      m->ChangePackageId();
```
gonuke, line 278: Don't wee need to check `unpackaged_` here?

gonuke (changes requested):
A few more changes...

src/package.cc
nuclearkatie: See header  49 to 62 now:
```
/// Returns optimal fill mass for a resource to be packaged. Can be used
/// to determine how to respond to requests for material, and to actually
/// package and send off trades.
/// Packaging strategy "first" simply fills the packages one by one to the
/// maximum fill. Therefore, it should always try to max fill.
/// Packaging strategy "equal" tries to fill all packages to the same mass.
/// This tries to find the optimal number and fill mass of packages given the
/// packaging limitations. It does this by calculating bounding fills, 
/// floor(quantity/fill_min) and ceiling(quantity/fill_max). 
/// There might be a scenario where there is no solution, i.e. an integer
/// number of packages cannot be filled with no remainder. In this case,
/// the most effective fill strategy is to fill to the max. Numeric example:
/// quantity = 5, fill_min = 3, fill_max = 4. num_min_fill = floor(5/3) = 1,
/// num_max_fill = ceil(5/4) = 2. num_min_fill < num_max_fill, so fill to max.
```

gonuke:
I think this look OK to me - just one lingering question about making `GetFillMass` and `Repackage` member functions of `Package`?

src/package.h
```diff
@@ -16,6 +16,45 @@ class Package {
...
+    /// quantity = 5, fill_min = 3, fill_max = 4. num_min_fill = floor(5/3) = 1,
+    /// num_max_fill = ceil(5/4) = 2. num_min_fill < num_max_fill, so fill to
+    /// the max.
+    double GetFillMass(Resource::Ptr r);
+
+    /// Repackages a single resource into a package. If some quantity of the 
+    /// resource cannot be packaged using the given packaging strategy and
+    /// restrictions, the remainder is left in the resource object. 
+    template <class T>
+    std::vector<typename T::Ptr> PackageResource(typename T::Ptr r) {
```
gonuke, line 40: Probably best to move this to the CC file instead of inline here.

gonuke: Also- do you use this anywhere yet?

gonuke (changes requested):
request to cleanup location of function now that it's moved to be a member function

src/package.h
nuclearkatie: No

gonuke (approved):

gonuke:
@abachma2 - can you review this again and/or accept the changes since your last review?

abachma2 (approved):
This all looks good to me. Thanks @nuclearkatie and @gonuke

PR #1684, opened 2024-03-07
Resolve Pytest warnings

bennibbelink:
There are many instances in `cycpp.py` where a standard string is used instead of a raw string literal leading to incorrect escaping.  This led to a failing test mentioned in #1683.  This PR resolves the regex warnings along with a handful of other pytest warnings.  Pytest should run cleanly now with no warnings.


gonuke (approved):
Thanks @bennibbelink - this looks clear/clean and passes tests.

PR #1689, opened 2024-03-10
More robust method for getting cyclus installation directory

### LINKED ISSUE #1688: GetEnv bug finding cyclus installation dir
in `src/env.cc.in` we use the env var `_` to find the path of the running executable (to infer the cyclus installation directory).  When I was experimenting with RHEL I discovered that when cyclus is called via python this variable is set to the location of python - not cyclus

bennibbelink:
Closes #1688.

Instead of relying on the environment variable `_` to get the path of the running application (which returns the python installation directory when cyclus is called via python) this calls `which cyclus` to find the location of the cyclus binary.  

This assumes that the machine has the `which` utility and that the cyclus installation directory is in the `PATH`.   Not 100% robust but I think these are assumptions we can safely make (or enforce on our users).

### DIFF
--- a/src/env.cc.in
+++ b/src/env.cc.in
@@ -36,7 +36,7 @@ const std::string Env::GetInstallPath() {
     return instdir_;
   std::array<char, PATH_MAX> buffer;
   std::string instdir;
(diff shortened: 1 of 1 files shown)


gonuke (approved):
Seems like there must be a better way 😁 but this looks ok to me.

PR #1691, opened 2024-03-12
CI Support for RHEL (new)

### LINKED ISSUE #1573: RHEL unable to install due to libxml++-2.6
RHEL does not have libxml++-2.6 available is there anyway to install cyclus with libxml++-2.4?

Must run RHEL on labs system for security requirements.

bennibbelink:
This PR adds a `Rocky.dockerfile` that builds cyclus on `rockylinux`.  Rocky Linux is an open source OS designed to be compatible with RHEL.  This PR also adds CI for Rocky Linux 8 and 9.  

Getting these builds to work is a little tricky since RHEL only supports older versions of most packages.  Notable examples are:
- `libxml++-2.42`.  Forces us to install `libxml++-4.0.3` from source
- `pandas`, `cython`, `jinja2`, `tables` don't exist for `python3.11`.  The Dockerfile `pip install`s these dependencies
- `meson`, `ninja`, `packaging` are needed for the `libxml++` installation, also installed via `pip`
-------
I opted to only include the `cyclus` build/test in CI for now.  I think it's fair to only perform downstream testing on our Ubuntu builds to avoid bloating CI.  Also, I'll leave it to your discretion @gonuke as to whether we **require** the `rocky_build_test.yml` workflow to pass to merge a PR.

Also... I'm open to naming convention suggestions for both the new Dockerfile and the container images pushed to GHCR.  Currently being pushed as `cyclus_rocky_${rocky_version_number}/cyclus`

Closes #1573


bennibbelink:
I added the rocky build/test as a separate job in `build_test.yml`.  It will "hard fail" as is.

cmake/UseCython.cmake
```diff
@@ -61,7 +61,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #=============================================================================
-
+# set(Python3_FIND_UNVERSIONED_NAMES NEVER)
```
gonuke, line 64: This is commented out - do we need it?

docker/Rocky.dockerfile
```diff
@@ -0,0 +1,89 @@
...
+        python3.11-pip \
+        python3.11-pytest \
+    && dnf clean all
+
+RUN mkdir -p $(python -m site --user-site) && python -m pip install pandas tables cython jinja2
+
+FROM dnf-deps as libxmlpp
+RUN dnf install -y m4 doxygen perl-open perl-XML-Parser diffutils pcre-cpp pcre-devel  && \
+    python -m pip install meson ninja packaging && \
+    wget https://github.com/libxmlplusplus/libxmlplusplus/releases/download/4.0.3/libxml++-4.0.3.tar.xz && \
```
gonuke, line 48: The RHEL world seems to be stubbornly opposed to updating this, huh?

gonuke (changes requested):
Only one minor request to remove a commented line

cmake/UseCython.cmake
bennibbelink: Nope, I removed it

docker/Rocky.dockerfile
bennibbelink: Yeah most of the packages on RHEL are very old.  This [website](https://packages.fedoraproject.org/pkgs/libxml++/libxml++/) is a good explorer if you're curious about any other package versions.

gonuke (approved):

PR #1698, opened 2024-03-13
Resolve more regex invalid escape sequences

### LINKED ISSUE #1695: Regex SyntaxWarnings from `cli/post-process-python.py` during build
I think this is due to invalid escape sequences similar to those in `cycpp.py` resolved by #1684.  Should be a simple fix of converting strings to raw string literals

bennibbelink:
Closes #1695 .

Converts a handful of strings uses by `re` to raw string literals to avoid invalid escape sequences.


gonuke (approved):
Thanks for tracking down more of these @bennibbelink

PR #1699, opened 2024-03-13
throw if invalid package id given

### LINKED ISSUE #1697: `Context::GetPackageById` compilation warning
This new method added by #1675 `Context::GetPackageById(int id)` is the source of a compilation warning: `/cyclus/src/context.cc:221:1: warning: control reaches end of non-void function [-Wreturn-type]`

I'm not familiar with the use-case of this method but it might be a good idea to explicitly return a null pointer or throw an exception if an invalid `id` is passed in.

nuclearkatie:
Built errors into `GetPackageByName` but forgot to do so for `GetPackageById`. Throws for both `id` < 0 and unable to find matching `id`

Closes #1697


bennibbelink (approved):
LGTM, thanks @nuclearkatie!

gonuke (approved):
Looks like this does the trick. Thanks for the quick fix @nuclearkatie

PR #1703, opened 2024-03-21
Updated cyclus dependencies listed in the README 

dean-krueger:
Changed README.rst to list the dependencies found in "ghcr.io/cyclus/cyclus_20.04_apt/cyclus:latest](http://ghcr.io/cyclus/cyclus_20.04_apt/cyclus:latest", which should represent the current minimum supported versions.


gonuke (approved):
Thanks @dean-krueger - this is helpful

gonuke:
Apparently there is a merge conflict here?  Not sure why...

gonuke (changes requested):
Actually, we should make sure we update this everywhere at once:
* DEPENDENCIES.rst
* INSTALL.rst

Anywhere else?

abachma2:
> Actually, we should make sure we update this everywhere at once:
> 
> * DEPENDENCIES.rst
> * INSTALL.rst
> 
> Anywhere else?

Probably need to update these files in the Cycamore repo as well.

DEPENDENCIES.rst
```diff
@@ -55,8 +57,8 @@ Package                Minimum Version
 ====================   ==================
 doxygen (for docs)     1.7.6.1
 tcmalloc (for speed)   any
-Cython                  >=0.25 and  <0.27
-Python (dev version)   2.7 or 3.3+
+Cython                 0.29+
```
gonuke, line 60: I thought we were updated to Cython 3 - (@bennibbelink) but maybe I missed that we are still compatible with cython 0.29?

gonuke (changes requested):
One comment about minimum Cython version....

Also - I wonder if there is a way to "include" one file in another so that we can maintain this dependency list in a single location rather than 3 locations?

DEPENDENCIES.rst
bennibbelink: We are still compatible with cython 0.29.  Installing `cython3` via `apt` installs version 0.29 in both Ubuntu 20.04 and 22.04

gonuke:
> Added the rst `include` directive. I will update this conversation when I build the site locally to make sure it works correctly.

Having suggested this... I'm guessing it doesn't render in github and only after processing with Sphinx... which maybe is not what we're after either?  Maybe we need to double down on just making folks click to get to the single definition of dependencies?

DEPENDENCIES.rst
```diff
@@ -25,16 +25,17 @@ dependencies is shown below:
...
+``libxml2``              2.9.10+
+``libxml++``             2.40.1
+``python``               3.8.10
+``sqlite3``              3.31.1
+``HDF5``                 1.10.4+
+``Coin-Cbc``             2.10.3
+``Coin-Clp``             1.17.5
+``zlib``                 1.2.11
 ====================   ==================
-
```
gonuke, line 37: I think this blank line may be necessary - at least for cleaner GH rendering

DEPENDENCIES.rst
```diff
@@ -25,16 +25,17 @@ dependencies is shown below:
...
 .. website_include_end
 
 On some platforms, such as Ubuntu 16.04, the following are also necessary:
```
gonuke, line 41: Do we want to keep this?  I'd say we leave 16.04 behind.  Are there other special platforms to call out?

DEPENDENCIES.rst
```diff
@@ -55,8 +56,8 @@ Package                Minimum Version
 ====================   ==================
 doxygen (for docs)     1.7.6.1
```
gonuke, line 57: (Not necessarily related to this PR) it might be useful to add a column to indicate what feature requires each optional dependency.

INSTALL.rst
```diff
@@ -12,20 +12,7 @@ provided with the source.
...
-``CMake``                2.8
-``boost``                1.46.1
-``libxml2``              2
-``libxml++``             2.36
-``python``               2.7 or 3.3+
-``sqlite3``              3.7.10
-``HDF5``                 1.8.4
-``Coin-Cbc``             2.5
-====================   ==================
+A full list of the Cyclus package dependencies is listed :doc:`here <DEPENDENCIES.rst>` :
```
gonuke, line 15:
```suggestion
A full list of the Cyclus package dependencies is listed `here <DEPENDENCIES.rst>` :
```

gonuke (changes requested):
A few more comments/requests.

DEPENDENCIES.rst
bennibbelink: That is a good idea.  I'll make an issue.

DEPENDENCIES.rst
bennibbelink: We could note that `libxml++` needs to be built from source in Rocky builds?

bennibbelink:
I think I've fixed all the links.  The project status table in the README still looks a little wonky, that's last big structural thing blocking(?) this PR.  About to hop on a flight though, I'll try to get to it tonight or tomorrow.

DEPENDENCIES.rst
abachma2: Does it though? On my machine (running Rocky 9.0) I have libxml++ and libxml++-4.0 from conda. Also, are enough people running on Rocky machines to warrant calling it out?

INSTALL.rst
```diff
@@ -12,20 +12,7 @@ provided with the source.
...
-``CMake``                2.8
-``boost``                1.46.1
-``libxml2``              2
-``libxml++``             2.36
-``python``               2.7 or 3.3+
-``sqlite3``              3.7.10
-``HDF5``                 1.8.4
-``Coin-Cbc``             2.5
-====================   ==================
+A full list of the Cyclus package dependencies is listed `here <DEPENDENCIES.rst>`_:
```
abachma2, line 15: Have y'all built the website when pointing to these changes? (see https://github.com/cyclus/cyclus.github.com/blob/source/rever.xsh) to make sure the website still builds with these changes?

gonuke: I'm not sure the role of  `rever.xsh` still?  There was a time when we used `rever` for releases, but I don't think that's true anymore?  I guess we should decide what we want to do instead, but I do still think it's best to keep only a single copy of these dependencies to ensure they don't diverge.

abachma2: As far as I can tell, for the website pages for dependencies and install Cyclus still point to this file, which points to the files in this repository and the ones in Cycamore. I think it's meant to keep everything in as few places as possible.

bennibbelink: Are you sure those files aren't pointed to in the `Makefile` under the `gh-preview` target?  Regardless I think we should be fine since we are pointing to `DEPENDENCIES.rst` which maintains the same name in the Sphinx build.  I will confirm with a local build.

DEPENDENCIES.rst
bennibbelink: True... libxml++ only needs to be built from source if you don't use conda on Rocky.  I'll leave it out for now.  If we receive feedback about difficulties building on Rocky we can re-evaluate.

INSTALL.rst
bennibbelink: I am fairly confident that we do not use `rever` anymore - should we remove it from our project repos? (can be a separate issue, don't want to block this PR any longer).  

That being said... you were right @abachma2 the new links I had were broken in the built website.  I wasn't able to find a trivial way to have the links render nicely in GitHub and by Sphinx so I'll leave them as is for now (with the Sphinx links working correctly).   The README is not pulled into the website build so those links are correctly formatted for use on Github.

bennibbelink:
> The project status table in the README still looks a little wonky, that's last big structural thing blocking(?) this PR. 

This is resolved, the table looks good now.

INSTALL.rst
```diff
@@ -12,29 +12,16 @@ provided with the source.
...
-``CMake``                2.8
-``boost``                1.46.1
-``libxml2``              2
-``libxml++``             2.36
-``python``               2.7 or 3.3+
-``sqlite3``              3.7.10
-``HDF5``                 1.8.4
-``Coin-Cbc``             2.5
-====================   ==================
+A full list of the Cyclus package dependencies is listed :doc:`here <DEPENDENCIES>`:
```
abachma2, line 15: Instead of pointing to the file, we should be able to have  like what's in line 20 (`.. website_start`) as a pointer around the table in `DEPENDENCIES.rst` that we can pull it. It'll probably be cleaner when building the website while also preventing from having to update multiple tables.

README.rst
```diff
@@ -72,24 +72,7 @@ Dependencies
...
-``libxml++``             2.36
-``python``               2.7 or 3.10
-``sqlite3``              3.7.10
-``HDF5``                 1.8.4
-``Coin-Cbc``             2.5
-``zlib``                 1.1.4
-====================   ==================
-
-For detailed instructions on installing dependencies, see `Installing Dependencies <DEPENDENCIES.rst>`_.
+Cyclus is built using ``CMake``. For detailed instructions on Cyclus dependencies, see `Installing Dependencies <DEPENDENCIES.rst>`_.
```
abachma2, line 75:
```suggestion
Cyclus is built using ``CMake``. For detailed instructions on Cyclus dependencies (including minimum version requirements), see `Installing Dependencies <DEPENDENCIES.rst>`_.
```

abachma2 (commented):
In general, I think all of these changes are great, and will make things easier for future developers. I have a few comments, with one being about the website build (and trying to streamline it) and the other is a small wording suggestion that won't block merging.

INSTALL.rst
gonuke: I think we're caught in a github rendering vs sphinx/website rendering bind.  I initially suggested the include like you suggested, but it doesn't render at all on github and looks like gibberish.  On the other hand, this sphinx based reference also doesn't render (although doesn't look like gibberish).  So maybe we go with the include?

abachma2: Do you have any screenshots of how the build looks and the code for it? If not, maybe we can hop on a call so I can see what you're getting.

gonuke: You can see how the `INSTALL` file renders on github when using the include approach [here](https://github.com/cyclus/cyclus/blob/5e2eed6781d2732dc0e8e70328055f4b05ffb3f1/INSTALL.rst) (this is from an old commit).

Notice how the section on "Dependencies" seems incomplete and then take a look at the raw code.

Screenshot 2024-03-28 at 11 13 19

abachma2: I see what you mean, and this is what I had in mind for code to use. Just pointing to it via what we have now should be good then.

gonuke (approved):

PR #1706, opened 2024-03-21
Remove smbchk for new release

abachma2:
The PR removes the `smbchk.py` and associated pytests. This functionality is now deprecated and the pytests were skipped.


src/relax_ng_validator.cc
```diff
@@ -81,9 +81,9 @@ bool RelaxNGValidator::Validate(const xmlpp::Document* doc) {
     throw ValidationError(e.message);
   }
 
-  if (res != 0) {
-    throw ValidationError("Document failed schema validation");
-  }
+  // if (res != 0) {
+  //  throw ValidationError("Document failed schema validation");
+  // }
```
bennibbelink, line 86: Are these commented lines meant to be included in this PR?

release/smbchk.py
gonuke, line 1: We intentionally left this file in the repository based on a conversation in #1396.  The purpose of this check was valuable, in principle, but became more complicated than we can support.  We decided to leave the file because, while it is theoretically possible to recover it from the git history, it could be hard to recall its name and/or know what to look for.

tests/test_smbchk.py
gonuke, line 1: Ditto above comment

gonuke (commented):
Some thoughts on leaving these files here even if we no longer refer to them.

src/relax_ng_validator.cc
abachma2: No, it was not. And it's probably why the tests are failing.

abachma2:
I get the reasoning for keeping the file around. Do we anticipate this ever being useful in the future? It looks like the `tests/test_abi.py` relies on this module and those tests are currently skipped. Should we remove that test?

gonuke:
> I get the reasoning for keeping the file around. Do we anticipate this ever being useful in the future? It looks like the `tests/test_abi.py` relies on this module and those tests are currently skipped. Should we remove that test?

I guess the utility of this in the future may depend a little on the size of the community.  With a small community, we can manage any incompatibility by discussion.  In an imagined world with a large community, it may be important to be able to make assurances about ABI compatibility, and thus invest the effort to do this.  Perhaps we we can put a meaningful enough entry in the Changelog to be a "breadcrumb" that allows us to find this again in the future using the git history?

If we do remove this, then I agree that we should remote the ABI tests as well.

abachma2:
We should get some more input on this one then @munkm @FlanFlanagan any thoughts?

abachma2:
Were there any other thoughts on this one? I'm okay with leaving this in the code base, so if I don't hear anything else in the next few days I'll close the PR.

CHANGELOG.rst
```diff
@@ -30,6 +30,7 @@ Since last release
 * Created Package class and optional declaration of packages in input files (#1673, #1699), package id is a member of resources (materials/products) (#1675)
 * CI support for Rocky Linux (#1691)
 * Added support for a ResBuf to behave as a single bulk storage with mixing & extraction of resources (#1687)
+* Removed deprecated `smbchk.py` functionality and tests (#1706)
```
gonuke, line 33: Maybe if we add a few more words to the changelog, it will leave a trail of where to find the code again, and we can continue with removing this code.
```suggestion
* Removed deprecated `smbchk.py` ABI consistency checking functionality and tests (#1706)
```

gonuke (approved):
I recommended some additional info in the changlog to support this removal @abachma2

abachma2:
That was a good suggestion @gonuke. I expanded upon that suggestion some to point to previous conversations about this functionality.

gonuke:
Thanks @abachma2

Issue #1707, opened 2024-03-21
Add timestamp to PR Comment


bennibbelink:
The PR comment for downstream build statuses is repeatedly updated on successive workflow runs.  It would be useful to have a timestamp/commit hash in the comment to signify which build it represents

PR #1708, opened 2024-03-23
Add timestamp/SHA to PR comment

bennibbelink:
Closes issue #1707.  Adds a timestamp and commit hash to the PR comment generated on downstream testing

### DIFF
--- a/.github/workflows/build_test.yml
+++ b/.github/workflows/build_test.yml
@@ -143,10 +143,13 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request'
     steps:
+      - name: Checkout cyclus
+        uses: actions/checkout@v4
+
       - name: Save PR number and Commit hash to file
         run: |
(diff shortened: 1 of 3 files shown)


bennibbelink:
Unfortunately we're not seeing what the comment will look like after this PR is merged.  I think it might be because of how `pr_comment.yml` is triggered (by the completion of `build_test.yml`).  This is what the header of the comment will look like:

## Downstream Build Status Report - 10ba101d0441f422383be2e32236827102c6f316

gonuke:
I this meant to add a timestamp as well? or just a commit hash?

bennibbelink:
I initially had it add a timestamp too but figured the commit hash was more relevant.  Forgot to change the PR title.  I can add the timestamp back in if you'd like me to.

gonuke:
I think both are useful since it can often be easier to check the timestamp without cross-referencing

gonuke (approved):
Thanks @bennibbelink - let's give this a shot

PR #1709, opened 2024-03-25
Parallelize Cyclus

bennibbelink:
This PR adds support for multi-threaded cyclus simulations.

### Assumptions 
`Tick()` and `Tock()` of each agent occur independently of other agents' `Tick()` and `Tock()`.  This is stated in our [documentation](https://fuelcycle.org/arche/timestep.html).

### Implementation
In `DoTick()` and `DoTock()` when the agents are iterated over we use `#pragma omp parallel for` to distribute the loop iterations amongst available threads.  This is **very** basic and I'm sure there is room for improvement, but that will likely require much more time to profile/benchmark different implementations (and note that performance will vary based on simulation properties).

### Testing
I implemented parameterized tests for those that call `RunSim()` (except `material_tests.cc`, the proportion of tests that call `RunSim()` is very small).  If OpenMP is enabled then these tests are just run with `1, 2, 3, 4` passed in as parameters, i.e. the number of threads to use.  If OpenMP is not enabled then the tests are run single-threaded.

### Usage
`--nthreads` or `-j` flag via the cyclus cli.  If this flag is not included then cyclus will run single threaded.  This also requires that cyclus is built with the `--parallel` flag to add flags for OpenMP and denote in `src/platform.h` that OpenMP is enabled (similar to how we denote Coin is enabled)

### Benchmarking
I benchmarked with and without the second `#pragma omp parallel for` in `DoTock()` to see how that affected performance with and without explicit_inventory.
I used `tests/input/lotka_volterra_determ.xml` to benchmark with the following changes implemented:
-  Modified the duration to be 30
- Added `<explicit_inventory>true/false</explicit_inventory>` to the `<control></control>` section

| nthreads | two #pragma in DoTock, explicit_inventory=true | two #pragma in DoTock explicit_inventory=false | one #pragma in DoTock explicit_inventory=true | one #pragma in DoTock explicit_inventory=false |
|---|---|---|---|---|
| 1 | 337.63 | 261.39 | 355.89 | 240.46 |
| 2 | 261.60 | 196.16 | 294.28 | 181.90 |
| 4 | 166.75 | 116.55 | 204.06 | 103.24 |
| 6 | 126.91 | 87.07 | 173.73 | 71.01 |
| 8 | 104.52 | 73.94 | 161.70 | 54.76 |
| 10 | 96.63 | 67.44 | 151.42 | 49.87 |
| 12 | 90.05 | 66.53 | 152.14 | 46.33 |
| 14 | 88.54 | 61.35 | 145.47 | 45.70 |
| 16 | 87.67 | 61.21 | 146.06 | 43.63 |
| 18 | 86.20 | 62.07 | 145.30 | 41.26 |
| 20 | 87.57 | 62.75 | 145.47 | 41.27 |

If we don't explicitly record inventory we don't take much of a performance hit from the second `#pragma` since we should never enter the `if` block.  However... if we **do** explicitly record inventory then the second `#pragma` improves performance significantly.  Additionally, the marginal performance gains by adding threads starts to flatten out around 10 threads or so... something to note in the user guide.

---------------

I don't intend to take focus away from the cyclus release/distribution.  I just started toying with this over the weekend and got a little carried away.


bennibbelink:
Hmm... Looks like the conda cycamore builds fail because it can't find `libgomp.so`.  If we choose to move forward with this I will make the appropriate changes to the cycamore CMake to link against OpenMP libraries.

bennibbelink:
Also - the code coverage report has unveiled that the cli flags are not being used in any tests.  I need to update the python test suite to call cyclus with different thread counts.

src/timer.cc
```diff
@@ -77,9 +81,10 @@ void Timer::DoBuild() {
 }
 
 void Timer::DoTick() {
-  for (std::map<int, TimeListener*>::iterator agent = tickers_.begin();
-       agent != tickers_.end();
-       agent++) {
+  #pragma omp parallel for
+  for (int i = 0; i < tickers_.size(); i++) {
```
gonuke, line 85: Is changing this to an integer loop necessary for the OMP prama?

gonuke (commented):
This looks pretty good.  One question about some code changes introduced here?

Do we know how the backends behave under this mode?  E.g. `RecordInventories()` is making a call to the backend, in this case SQL.  How does it handle getting multiple requests from these different threads?

src/timer.cc
bennibbelink: I think so, it will not compile using the previous method of iteration.  OpenMP only supports `random_access_iterator`s as a loop variable, but `std::map` only supports `bidirectional_iterator`s (per this [forum](https://stackoverflow.com/questions/8848870/use-openmp-in-iterating-over-a-map)).  If we wanted to keep the loop as is I think we would need to copy the contents of `tickers_` to a `std::vector` (or something else that supports `random_access_iterator`s) and then iterate over it.

bennibbelink:
> Do we know how the backends behave under this mode? E.g. RecordInventories() is making a call to the backend, in this case SQL. How does it handle getting multiple requests from these different threads?

SQLite should be able to handle the relatively small number of concurrent requests.  I can add tests for this, perhaps using a utility such as `sqldiff`.  If issues arise we can always leave `RecordInventories()` in serial or wrap the critical sections with  `#pragma omp critical` to ensure only one thread is writing to a db at a time.

gonuke:
This is looking good to me - let me know when it's ready for a review.  @abachma2 may want to test with the archetypes from her dissertation, as well.

abachma2:
I definitely want to try it all out, but it way take a while for me to get some updates on that archetype done, so please don't wait on me for that.

bennibbelink:
It's not ready yet.  I've been adding more tests and seeing some failures that I need to dig into.   I will update this conversation when it is ready for review.

bennibbelink:
It appears as if everything works great when using agents written in C++.  However... any tests that use agents written in Python   are segfaulting (non-deterministically) in Cyclus<Facility/Institition/Region>Shim.  I suspect it has something to do with the Python interpreter allocating/deallocating objects in multiple threads.  I would think that this would be handled by the GIL but my understanding must be incorrect.  My local attempts so far to implement locking have been unsuccessful (leads to deadlock).  I will do more debugging to get a better understanding of what is happening under the Cython hood.

gonuke:
The python limitation might impact @abachma2 's testing :)

abachma2:
Yes. My stuff is all written in python, and I don't think there is a way to convert it to C++. But even getting Cyclus parallelized for C++ archetypes may be helpful as a starting point.

bennibbelink:
I have found a working solution.  Since I don't fully understand what is happening when we try to Tick/Tock python agents in parallel we now run them in serial always.  `DoTick()` and `DoTock()` now partitions the `tickers_` map into C++ agents and Python agents so that it can execute C++ agents in parallel and python agents in serial.  

In order to make this work I added a function `IsShim()` to the `TimeListener` class that returns whether it is a Shim or not - this should never be accessed by the user and returns True in the Shim class definitions and False in the the standard Region/Institution/Facility definitions. 

Obviously we will take some performance hit from this but I think it is the safest thing to do for now, and we should still see significant performance gains - especially if a simulation has a small number of python agents and a large number of C++ agents (benchmarks pending).  Thorough benchmarking could provide us with information to further optimize simulations, e.g. under a certain # of agents it is likely optimal to run entirely in serial.  

Tests are failing with Ubuntu/conda due to stderr output `Experimental Warning: ResMap is experimental and its API may be subject to change`, and in Rocky 8 due to a compilation error.  I need to dive into these but my intuition tells me that they will **not** impede the parallelization strategy.

bennibbelink:
The conda builds were failing when I had two `pragma omp section`s for each vector (so that a single thread could execute py_agent tasks while the cpp_agent tasks were distributed amongst many other threads) so now the Python agents are executed outside of any parallel constructs and that seems to be working well.  (still interesting that we see different parallel behavior between conda and apt builds)

The only CI failure now is on Rocky 8 related to C++ syntax.  Out of curiosity, how long should we plan to [support](https://access.redhat.com/product-life-cycles?product=Red%20Hat%20Enterprise%20Linux,OpenShift%20Container%20Platform%204) RHEL 8?

Last thing to do is implement tests that `diff` generated sqlite files.

bennibbelink:
Some updates:

I put infrastructure in place in the pytests to diff the contents of the generated sqlite files only to realize that all the simulations have unique simulation ids which are present throughout the tables.  We do query generated tables in some of the pytests already for the record.

-----------------------------------------------------------------------

I finally did some proper profiling using `gperftools` and have some findings.  I profiled cyclus (once again on the lotka volterra simulation) and the multithreaded version was actually slower than cyclus on the `main` branch:

(samples refer to the number of times the profiler sampled the running code.  It is proportional to time)
`main` - 717 samples
`openmp` with 4 threads - 1289 samples
`openmp` with 10 threads - 1632 samples
`openmp` with 20 threads - 1596 samples

I was curious if we would ever see a performance benefit with parallelism so I multiplied the number of agents in the simulation by 5 to see if we could reap more benefits:
`main` - 15419 samples
`openmp` with 2 threads - 13957 samples
`openmp` with 10 threads - 13111 samples

We do see a marginal performance gain with 5x the number of agents.  I am also curious how results would differ if our agents' Tick and Tock methods were more complex.  This raises the question (for me at least) of what a "typical" simulation looks like for our users.  i.e. number/complexity of agents, duration, inventory recording, etc.  

Sadly, if `lotka_volterra_determ.xml` is a good representation of a real-life use case I think the `--parallel` build will be less performant 9 times out of 10.  However, this PR does put the infrastructure in place for using OpenMP that we could *possibly* use to speed up other sections of the code.  At your discretion @gonuke as to whether we move forward with this build option (but not a recommendation until we know the right place for it).

  On the bright side - I collected some valuable profiling data and call graphs (attached to this comment), so we know where in the code we can improve.
[profile_svgs.zip](https://github.com/cyclus/cyclus/files/14940750/profile_svgs.zip)
This is the zip of call graphs from various runs, you can open the svgs in a browser to view and explore.


----------------------------------------------------
My table of results above is **misleading** because I was testing cyclus with 1 thread when it was built with openmp support (and using the modified DoTick/DoTock implementations).  In the call graph I now see that using `std::advance` to advance the iterator was causing this implementation to be really slow

bennibbelink:
These were suggested as good resources to test with: 
- https://zenodo.org/records/4557613
- https://github.com/arfc/openmcyclus

docker/Dockerfile
```diff
@@ -130,7 +129,7 @@ WORKDIR /cyclus
 
 # You may add the option "--cmake-debug" to the following command
 # for further CMake debugging.
-RUN python3 install.py -j ${make_cores} --allow-milps
+RUN python3 install.py -j ${make_cores} --allow-milps --parallel
```
gonuke, line 132: Do we test that it still builds correctly without this flag?

CMakeLists.txt
```diff
@@ -337,7 +350,8 @@ IF(NOT CYCLUS_DOC_ONLY)
         "${SQLite3_INCLUDE_DIR}"
         "${HDF5_INCLUDE_DIRS}"
         "${Boost_INCLUDE_DIR}"
-        "${COIN_INCLUDE_DIRS}")
+        "${COIN_INCLUDE_DIRS}"
+        "${OpenMP_CXX_INCLUDE_DIRS}")
```
gonuke, line 354: Is this defined at all if we don't set the parallel flag? Is that a problem?

src/context.cc
```diff
@@ -118,29 +122,35 @@ void Context::DelAgent(Agent* m) {
...
-      ->AddVal("Prototype", proto_name)
-      ->AddVal("SchedTime", time())
-      ->AddVal("BuildTime", t)
-      ->Record();
 }
 
 void Context::SchedDecom(Agent* m, int t) {
-  if (t == -1) {
-    t = time();
+  #pragma omp critical
```
gonuke, line 142: Doesn't this happen once per thread, and therefore generate one database entry per thread?

It may also schedule the decommissioning too many times, and I'm not sure if that matters or not.

src/facility.h
```diff
@@ -161,6 +161,8 @@ class Facility : public TimeListener, public Agent, public Trader {
       const std::vector<std::pair<Trade<Product>,
       Product::Ptr> >& responses) {}
 
+  virtual bool IsShim() { return false; }
```
gonuke, line 164: I'm not 100% clear on the inheritance between the C++ and the python, but if this was implemented as a default implementation in `time_listener.h` wouldn't it prevent the need from implementing it in each of `facility.h`, `institution.h` and `region.h`?  Or does the python agent somehow inherit from that, too?

src/timer.cc
```diff
@@ -76,11 +79,32 @@ void Timer::DoBuild() {
   }
 }
 
+void Timer::PartitionTickers(std::vector<TimeListener*>& cpp_agents, std::vector<TimeListener*>& py_agents) {
+  for (std::pair<int, TimeListener*> pair : tickers_) {
```
gonuke, line 83: What if we generate this partition at the same time that we populate `tickers_`? (And at the places when we depopulate it for decommissioning?)

This avoids having to regenerate it at each Tick

tests/test_lotka_volterra.py
```diff
@@ -27,16 +26,16 @@ def test_predator_only():
     sim_input = os.path.join(DIR, "input", "predator.xml")
 
     holdsrtn = [1]  # needed because nose does not send() to test generator
-    outfile = which_outfile()
-
-    cmd = ["cyclus", "-o", outfile, "--input-file", sim_input]
+    outfile = which_outfile(thread_count)
+    print(outfile)
```
gonuke, line 30: Do we need this still?

gonuke (changes requested):
Thanks for getting this all pulled together @bennibbelink  .  I have a few implementation questions and queries..

CMakeLists.txt
bennibbelink: It won't be defined at all, but it shouldn't be a problem.  It will just append an empty string to the end of `inc_dirs`, which would happen when we build without COIN as well.  We could add the optional dependencies within respective `IF(<dep found>)...ENDIF()` blocks to handle this more gracefully

src/context.cc
bennibbelink: At each timestep each agent's Tick/Tock is executed by a single thread, so SchedDecom should only be executed once for each agent, right?.  I don't think that this behavior should differ from a serial implementation, apart from perhaps the order of these entries in the DB. There should the the same number of database entries.   Please let me know if I'm missing something about this though, my understanding of an agent's lifecycle might me lacking in some way  

The `#pragma omp critical` block is to ensure that DB writes are not attempted by two threads at the same time.

CMakeLists.txt
gonuke: It's fine this way, just wanted to check

src/context.cc
gonuke: Right... this is called by institutions... makes sense.

src/timer.cc
bennibbelink: I added std::vectors `cpp_tickers_` and `py_tickers_` that are registered/unregistered at the same time as `tickers_`. I opted to use a vector so we don't have to transform it from std::map to something that is parallelizable at each timestep

src/facility.h
bennibbelink: Good point, `IsShim()` is now only implemented in `time_listener.h`

docker/Dockerfile
bennibbelink: I added a build arg to the dockerfile and a matrix to the workflow.  We can also add a matrix for `--allow-milps` (or any other optional build flags we may want in the future), see `macos_build_test.yml`.  I didn't add it to `build_test.yml` yet because the cycamore Dockerfile doesn't support passing build flags through a `--build-arg` yet. We can add a coin_flags matrix to build_test.yml in a PR following this one (and a cycamore PR) assuming its desired

bennibbelink:
Experiencing non-deterministic test failures in SymbolicFunctionTests.  It looks like it might be related to overflow, but need to diagnose some more

src/timer.cc
```diff
@@ -91,21 +99,24 @@ void Timer::DoResEx(ExchangeManager<Material>* matmgr,
...
     std::set<Agent*> ags = ctx_->agent_list_;
-    std::set<Agent*>::iterator it;
-    for (it = ags.begin(); it != ags.end(); ++it) {
-      Agent* a = *it;
-      if (a->enter_time() == -1) {
-        continue; // skip agents that aren't alive
-      }
-      RecordInventories(a);
+    std::vector<Agent*> agent_vec(ags.begin(), ags.end());
+    #pragma omp parallel for
```
gonuke, line 114: Does this mean we can safely parallelize over python agents for this operation? Or that python agents don't support `RecordInventories`?

gonuke (changes requested):
Thanks for getting this all pulled together.  It will be a very interesting advance!

I know you're off the payroll, but wonder if you're willing to answer one lingering question here?  If not, that's fine, and I'll dig a little more myself.

src/timer.cc
bennibbelink: It might be worth looking into some more to confirm, but this is my understanding:

Python agents do support `RecordInventories()`, but their `RecordInventories()` methods can be executed in parallel because of the way they are implemented in cython.  If you look in `cyclus/agents.pyx` these methods are defined using `cdef` and only call other methods also defined via `cdef` (`lib.inventories_to_cpp()`).  Thus, the method is compiled as pure C++ and the python interpreter in not being invoked as it might be when calling a python agent's `Tick` or `Tock`

I have a suspicion that if you changed the definition of `RecordInventories` to use `cpdef` and call a python method within it you might start to see segmentation faults in this parallel for loop.

gonuke (approved):
Thanks to @bennibbelink for this big step forward

PR #1711, opened 2024-03-27
Unpackaged as default Package id

### LINKED ISSUE #1692: Tweak implementation of `default_package_id_`
The `default_package_id_` is used for any circumstance when packaging is not important and is effectively equivalent to being unpackaged.  I propose that we rename it `unpackaged_id_`.

I also propose that we make it a public const static member of the `Package` class, this requiring that it be referred to as `Package::unpackaged_id_` in most circumstances.  This change results in a slightly more verbose usage, but locates the variable in a more logical place for future maintenance and readability.

nuclearkatie:
Closes #1692


gonuke (approved):
Nice! This subtle change will have a big impact on readability and maintainability in the long run. Thanks @nuclearkatie

PR #1712, opened 2024-03-27
add tests for Package

gonuke:
Some simple tests for Package - hoping for a positive change to coverage!

### DIFF
--- a/src/product.cc
+++ b/src/product.cc
@@ -78,14 +78,14 @@ int Product::package_id() {
 }
 
 void Product::ChangePackageId(int new_package_id) {
-  if (ctx_ != NULL) {
+  if (ctx_ == NULL) {
     throw ValueError("Package Id cannot be changed with NULL context");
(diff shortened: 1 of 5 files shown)


src/product.cc
```diff
@@ -78,14 +78,14 @@ int Product::package_id() {
 }
 
 void Product::ChangePackageId(int new_package_id) {
-  if (ctx_ != NULL) {
+  if (ctx_ == NULL) {
     throw ValueError("Package Id cannot be changed with NULL context");
   }
```
gonuke, line 83: @nuclearkatie - in `material.cc` we just return rather than throwing an error in this case. Do we need to throw an error?

gonuke:
I was going to add tests for `GetFillMass` but I don't think it all works well enough so we'll leave that for a future test/PR.

nuclearkatie (approved):
LGTM

Issue #1713, opened 2024-03-28
Fix links in INSTALL.rst


abachma2:
PR #1703 got merged before I could point out that we are missing some underscores to link to other files in `INSTALL.rst`, so the links just show up as code instead of as hyperlinks.

gonuke:
oops! sorry

gonuke:
However that may be related to Sphinx formatting vs GH formatting?  There was some thrashing on this in the commit history of #1703

abachma2:
I don't think so. The `README.rst` has underscores after the URLs, (like ``text <URL>`_`) that allows things to render, and the website works with that too.

gonuke:
OK - we'll have to come up with a single approach and make sure it's applied throughout.

PR #1561, opened 2021-06-23
Mention necessity of creating `news` items for PRs.

dermesser:
Came up in #1560. 

EDIT: and in #1518

PR #1715, opened 2024-04-01
Small documentation fixes

abachma2:
This PR includes some small documentation fixes such as:
* a note in the `CONTRIBUTING` that tells developers to update the CHANGELOG, addressing what was started in #1561 
* fix the formatting for cross referencing in `INSTALL.rst` hat renders correctly on the website and when viewed on GitHub
* change conda package for dependencies to install only the dependencies for the cyclus package, allowing us to not have to maintain a core package and a dependencies package.
* Change the branch name for guiding new developers in the `README`

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -12,7 +12,7 @@ provided with the source.
 Dependencies
 ************
 
-A full list of the Cyclus package dependencies is listed :doc:`here <DEPENDENCIES>`_:
+A full list of the Cyclus package dependencies is listed `here <DEPENDENCIES>`_:
 
 ************
 Installation
@@ -21,7 +21,7 @@ Installation
 
 Before going further with the installation procedure be sure you have installed
 all the required dependencies. We have provided detailed
-instructions for :doc:`installing dependencies <DEPENDENCIES.rst>_` for the major supported
+instructions for :doc:`installing dependencies <DEPENDENCIES.rst>` for the major supported
 systems.
 
 
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -12,7 +12,7 @@ provided with the source.
 Dependencies
 ************
 
-A full list of the Cyclus package dependencies is listed `DEPENDENCIES`_.
+A full list of the Cyclus package dependencies is listed `here <DEPENDENCIES.rst>`_.
 
 ************
 Installation
@@ -21,7 +21,7 @@ Installation
 
 Before going further with the installation procedure be sure you have installed
 all the required dependencies. We have provided detailed
-instructions for `installing dependencies <DEPENDENCIES.rst>_` for the major supported
+instructions for `DEPENDENCIES_` for the major supported
 systems.
 
 
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -12,7 +12,7 @@ provided with the source.
 Dependencies
 ************
 
-A full list of the Cyclus package dependencies is listed `here <DEPENDENCIES.rst>`_.
+A full list of the Cyclus package dependencies is listed `here`_.
 
 ************
 Installation
@@ -21,7 +21,7 @@ Installation
 
 Before going further with the installation procedure be sure you have installed
 all the required dependencies. We have provided detailed
-instructions for `DEPENDENCIES`_ for the major supported
+instructions `for installing those dependencies for the major supported systems`_ for the major supported
 systems.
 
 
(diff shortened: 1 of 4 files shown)


INSTALL.rst
```diff
@@ -12,7 +12,7 @@ provided with the source.
 Dependencies
 ************
 
-A full list of the Cyclus package dependencies is listed :doc:`here <DEPENDENCIES>`:
+A full list of the Cyclus package dependencies is listed `here`_.
```
bennibbelink, line 15: Have you tried building the website with this change? This is what I get when I click on the link "here".

Screenshot 2024-04-01 at 6 34 38 PM

INSTALL.rst
```diff
@@ -21,7 +21,7 @@ Installation
 
 Before going further with the installation procedure be sure you have installed
 all the required dependencies. We have provided detailed
-instructions for :doc:`installing dependencies <DEPENDENCIES>` for the major supported
+instructions `for installing those dependencies for the major supported systems`_ for the major supported
```
bennibbelink, line 24: Same as above.

INSTALL.rst
abachma2: I did test this with a local build and it works for the instance below. I don't think this section shows up in the website, the `.. website_include_start` is after this section. Which page of the website does this link show up in?

bennibbelink: The Cyclus Installation Guide (/user/CYCLUS_INSTALL.html).  I am building it with these modifications to the Makefile so that it fetches your new files.

Screenshot 2024-04-02 at 9 13 01 AM

abachma2:
Ok. So this isn't working for me either, even though I could have sworn it worked yesterday. The main issue we are running into here is that I am trying to find a format for internal hyperlinks that works when rendering this on the website and when you view it on GitHub. I don't think this is possible, because we would need one to point to a `.rst` file and the other to point to a `.html` file. 

So a few ideas here:
* we can keep the install files as separate things in each repo instead of getting the ones from the Cyclus and Cycamore repos so that we can make sure everything renders as needed
* not worry about rendering on github (which is what we do in the website repo), and just link to the website.

I am also open to other ideas.

INSTALL.rst
abachma2: See my comment in the main conversation.

gonuke:
I wonder if we can use Markdown files and this [extension](https://www.sphinx-doc.org/en/master/usage/markdown.html) when we pull it into the website?  No idea if it will generate URLs correctly. 🤷‍♂️

bennibbelink:
There must ways to get what we want if we add a pre-processing step. That extension might work, I'll see what else I can find.

bennibbelink:
After some research I think I'm in support of the third bullet by @abachma2:
> We can have the website link to the file on Github, via the full web address, where everything renders correctly.

I haven't found a good way to make the link formatting work well in both Github and Sphinx contexts.  We have a good way to link to an external reference (Github in this case), so I my suggestion is that we change the references at the bottom of `INSTALL.rst` to point to https://github.com/cyclus/cyclus/blob/main/DEPENDENCIES.rst instead of `DEPENDENCIES.rst`.  It's not ideal but its only for these two instances, every other reference on the website to DEPENDENCIES should render fine.

abachma2:
I pushed that change. Is there anything else I needed to change for this to get merged?

abachma2:
And now that I've pushed this, we could also link to page on the website: https://fuelcycle.org/user/DEPENDENCIES.html. If y'all have a preference of one over the other.

bennibbelink:
Good point! I think referencing the page on https://fuelcycle.org would probably be better so we keep users in one place (although less robust).

abachma2:
Updated the link to the website instead of the github file. Is this ready to be merged now?

bennibbelink (approved):
Looks good to me, thanks @abachma2!

abachma2:
If this is approved @bennibbelink, can we merge it? Or do you think we need another reviewer on it?

bennibbelink:
I saw the review from @gonuke is pending.  Do you have a preference about linking to fuelcycle.org vs GitHub?

gonuke (changes requested):
Just one typo to fix - then I'm happy with it.

abachma2:
Thanks for catching that typo @gonuke. Is this ready for merging now?

nsryan2:
This PR closes #1713

gonuke (approved):

PR #1729, opened 2024-04-11
Package management by context and handling of repackaging process

### LINKED ISSUE #1714: Resolve templating for Pacakge functions
A couple of functions in the `Package` class are templated on the class of the argument (`Material` vs `Product`).  This seems to be a type of templating that is unresolvable by the compiler and results in missing symbols.

One approach is to promote these methods to use the parent class `Resource` as the argument and ensure that all children have a sufficient interface implemented.

### LINKED ISSUE #1730: Unpackaged id gets created as id 2, id 1 is not actually tied to a package
We set unpackaged id as equal to 1, but the unpackaged (default) package actually gets created in the xml file loader. Because the package `next_id` is 2 (in package.cc), this means when a simulation is spinning up, the default package type is created with id 2, and resources are automatically given the unpackaged id of 1. Because there is no actual `Package` with id 1, if you try to repackage a resource to unpackaged, an error is raised

### LINKED ISSUE #1731: New package types can't just be created, they must also be given to the context to manage
Need to use context AddPackage to generate new package types, which creates a package and then stores its info. Should not use Package::Create, which creates a package but then the context doesn't know about it and it can't be accessed by id

nuclearkatie:
- [x] Fix unpackaged (default) package type. Closes #1730 
- [x] Ensure that creating new packages is done through the context (context::AddPackage), which calls Package::Create. Closes #1731 
- [x] Package-filling process managed by resource, not by package. Closes #1714

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -85,10 +85,7 @@ Context::Context(Timer* ti, Recorder* rec)
       trans_id_(0),
       si_(0) {
         rng_ = new RandomNumberGenerator();
-        if (packages_.count("unpackaged") == 0) {
-          packages_["unpackaged"] = Package::CreateUnpackaged();
-        }
-}
+      }
 
 Context::~Context() {
   if (solver_ != NULL) {
@@ -207,6 +204,9 @@ Package::Ptr Context::AddPackage(std::string name, double fill_min, double fill_
 }
 
 Package::Ptr Context::GetPackageByName(std::string name) {
+  if (name == Package::unpackaged_name()) {
+    return Package::unpackaged();
+  }
   if (packages_.count(name) == 0) {
     throw KeyError("Invalid package name " + name);
   }
@@ -214,6 +214,9 @@ Package::Ptr Context::GetPackageByName(std::string name) {
 }
 
 Package::Ptr Context::GetPackageById(int id) {
+  if (id == Package::unpackaged_id()) {
+    return Package::unpackaged();
+  }
   if (id < 0) {
     throw ValueError("Invalid package id " + std::to_string(id));
   }
--- a/src/context.cc
+++ b/src/context.cc
@@ -191,7 +191,7 @@ Composition::Ptr Context::GetRecipe(std::string name) {
   return recipes_[name];
 }
 
-Package::Ptr Context::AddPackage(std::string name, double fill_min, double fill_max,
+void Context::AddPackage(std::string name, double fill_min, double fill_max,
                          std::string strategy) {
   packages_[name] = Package::Create(name, fill_min, fill_max, strategy);
   NewDatum("Packages")
@@ -200,7 +200,6 @@ Package::Ptr Context::AddPackage(std::string name, double fill_min, double fill_
     ->AddVal("FillMax", fill_max)
     ->AddVal("Strategy", strategy)
     ->Record();
-  return packages_[name];
 }
 
 Package::Ptr Context::GetPackageByName(std::string name) {
--- a/src/context.cc
+++ b/src/context.cc
@@ -84,9 +84,11 @@ Context::Context(Timer* ti, Recorder* rec)
       solver_(NULL),
       trans_id_(0),
       si_(0) {
-
         rng_ = new RandomNumberGenerator();
-      }
+        if (packages_.count("unpackaged") == 0) {
+          packages_["unpackaged"] = Package::CreateUnpackaged();
+        }
+}
 
 Context::~Context() {
   if (solver_ != NULL) {
@@ -192,7 +194,7 @@ Composition::Ptr Context::GetRecipe(std::string name) {
   return recipes_[name];
 }
 
-void Context::AddPackage(std::string name, double fill_min, double fill_max,
+Package::Ptr Context::AddPackage(std::string name, double fill_min, double fill_max,
                          std::string strategy) {
   packages_[name] = Package::Create(name, fill_min, fill_max, strategy);
   NewDatum("Packages")
@@ -201,6 +203,7 @@ void Context::AddPackage(std::string name, double fill_min, double fill_max,
     ->AddVal("FillMax", fill_max)
     ->AddVal("Strategy", strategy)
     ->Record();
+  return packages_[name];
 }
 
 Package::Ptr Context::GetPackageByName(std::string name) {
--- a/src/context.h
+++ b/src/context.h
@@ -251,7 +251,7 @@ class Context {
 
   /// Adds a package type to a simulation-wide accessible list.
   /// Agents should NOT add their own packages.
-  Package::Ptr AddPackage(std::string name, double fill_min = 0,
+  void AddPackage(std::string name, double fill_min = 0,
                           double fill_max = std::numeric_limits<double>::max(),
                           std::string strategy = "first");
 
--- a/src/context.h
+++ b/src/context.h
@@ -251,9 +251,9 @@ class Context {
 
   /// Adds a package type to a simulation-wide accessible list.
   /// Agents should NOT add their own packages.
-  void AddPackage(std::string name, double fill_min = 0,
-                  double fill_max = std::numeric_limits<double>::max(),
-                  std::string strategy = "first");
+  Package::Ptr AddPackage(std::string name, double fill_min = 0,
+                          double fill_max = std::numeric_limits<double>::max(),
+                          std::string strategy = "first");
 
   // Retrieve a registered package. 
   Package::Ptr GetPackageByName(std::string name);
--- a/src/context.h
+++ b/src/context.h
@@ -252,8 +252,8 @@ class Context {
   /// Adds a package type to a simulation-wide accessible list.
   /// Agents should NOT add their own packages.
   void AddPackage(std::string name, double fill_min = 0,
-                          double fill_max = std::numeric_limits<double>::max(),
-                          std::string strategy = "first");
+                  double fill_max = std::numeric_limits<double>::max(),
+                  std::string strategy = "first");
 
   // Retrieve a registered package. 
   Package::Ptr GetPackageByName(std::string name);
--- a/src/material.cc
+++ b/src/material.cc
@@ -227,13 +227,14 @@ double Material::DecayHeat() {
   return decay_heat;
 }
 
-std::vector<Material::Ptr> Material::Package(Package::Ptr pkg) {
+std::vector<Resource::Ptr> Material::Package(Package::Ptr pkg) {
   std::vector<Material::Ptr> ms_pkgd;
   Material::Ptr m_pkgd;
 
   double fill_mass = pkg->GetFillMass(quantity());
   if (fill_mass == 0) {
-    return ms_pkgd;
+    std::vector<Resource::Ptr> rs_pkgd;
+    return rs_pkgd;
   }
 
   while (quantity() > pkg->fill_min()) {
@@ -242,7 +243,11 @@ std::vector<Material::Ptr> Material::Package(Package::Ptr pkg) {
     m_pkgd->ChangePackageId(pkg->id());
     ms_pkgd.push_back(m_pkgd);
   }
-  return ms_pkgd;
+  std::vector<Resource::Ptr> rs_pkgd;
+  for (int i = 0; i < ms_pkgd.size(); ++i) {
+    rs_pkgd.push_back(boost::dynamic_pointer_cast<Resource>(ms_pkgd[i]));
+  }
+  return rs_pkgd;
 }
 
 Composition::Ptr Material::comp() const {
--- a/src/material.cc
+++ b/src/material.cc
@@ -227,6 +227,24 @@ double Material::DecayHeat() {
   return decay_heat;
 }
 
+std::vector<Material::Ptr> Material::Package(Package::Ptr pkg) {
+  std::vector<Material::Ptr> ms_pkgd;
+  Material::Ptr m_pkgd;
+
+  double fill_mass = pkg->GetFillMass(quantity());
+  if (fill_mass == 0) {
+    return ms_pkgd;
+  }
+
+  while (quantity() > pkg->fill_min()) {
+    double pkg_fill = std::min(quantity(), fill_mass);
+    m_pkgd = boost::dynamic_pointer_cast<Material>(ExtractRes(pkg_fill));
+    m_pkgd->ChangePackageId(pkg->id());
+    ms_pkgd.push_back(m_pkgd);
+  }
+  return ms_pkgd;
+}
+
 Composition::Ptr Material::comp() const {
   throw Error("comp() const is deprecated - use non-const comp() function."
               " Recompilation should fix the problem.");
--- a/src/material.cc
+++ b/src/material.cc
@@ -227,29 +227,6 @@ double Material::DecayHeat() {
   return decay_heat;
 }
 
-std::vector<Resource::Ptr> Material::Package(Package::Ptr pkg) {
-  std::vector<Material::Ptr> ms_pkgd;
-  Material::Ptr m_pkgd;
-
-  double fill_mass = pkg->GetFillMass(quantity());
-  if (fill_mass == 0) {
-    std::vector<Resource::Ptr> rs_pkgd;
-    return rs_pkgd;
-  }
-
-  while (quantity() > pkg->fill_min()) {
-    double pkg_fill = std::min(quantity(), fill_mass);
-    m_pkgd = boost::dynamic_pointer_cast<Material>(ExtractRes(pkg_fill));
-    m_pkgd->ChangePackageId(pkg->id());
-    ms_pkgd.push_back(m_pkgd);
-  }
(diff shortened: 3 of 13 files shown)


src/resource.cc
```diff
@@ -10,4 +10,47 @@ void Resource::BumpStateId() {
...
 }
 
+double Resource::GetFillMass(Package::Ptr pkg) {
+  if (quantity() < pkg->fill_min()) {
+    // less than one pkg of material available
+    return 0;
+  }
+
+  std::vector<Resource::Ptr> rs;
+  Resource::Ptr r_pkgd;
```
gonuke, line 20: Neither of these are used.

src/resource.cc
```diff
@@ -10,4 +10,47 @@ void Resource::BumpStateId() {
   nextstate_id_++;
 }
 
+double Resource::GetFillMass(Package::Ptr pkg) {
```
gonuke, line 13: I'm suddenly having second thoughts (third thoughts? fourth thoughts?) about moving this to `Resource`.  Here's why:
1. it only uses the `quantity()` from the `Resource` so we could pass that in a version that is `Package::GetFillMass(double qty)` and not need to worry about templates.
2. it depends on a lot of details of `Package` such as the fill strategies - separation of concerns suggests that it will be awkward to have to update a method in `Resource` if we change how `Package` internals work.

I was a little blinded by the templating challenge, and didn't see that these factors really make `GetFillMass` a property of a `Package`.  `PackageResource()` should stay in `Resource`.

Open for more discussion, sooner rather than later.

src/resource.cc
```diff
@@ -10,4 +10,47 @@ void Resource::BumpStateId() {
...
+  Resource::Ptr r_pkgd;
+  
+  double fill_mass = GetFillMass(pkg);
+  if (fill_mass == 0) {
+    return rs_pkgd;
+  }
+
+  while (quantity() > pkg->fill_min()) {
+    double pkg_fill = std::min(quantity(), fill_mass);
+    r_pkgd = boost::dynamic_pointer_cast<Resource>(ExtractRes(pkg_fill));
```
gonuke, line 49: It seems like `ExtractRes` already returns a `Resource::Ptr` - do we need to cast it here?

gonuke (changes requested):
With apologies to another flip-flop (let's discuss) here are some otherwise minor comments.

src/package.cc
```diff
@@ -4,71 +4,58 @@
...
+  }
+  else if (fill_min > fill_max) {
+    throw ValueError("fill_min must be less than or equal to fill_max");
+  }
+  Ptr p(new Package(name, fill_min, fill_max, strategy));
+  return p;
+}
+
+Package::Ptr Package::CreateUnpackaged() {
+  Ptr p(new Package(1, "unpackaged"));
```
gonuke, line 26: Safest to use `unpackaged_id_` here and maybe introduce `unpackaged_name_` as well?  That way there can always be a single place to set/change those things.

src/package.cc
```diff
@@ -4,71 +4,58 @@
...
-  }
-  Ptr p(new Package(name, fill_min, fill_max, strategy));
-  return p;
-}
-  
-Package::Package() : id_(next_id_++), fill_min_(0), fill_max_(std::numeric_limits<double>::max()) {}
+Package::Package(std::string name, double fill_min, double fill_max, std::string strategy) : name_(name), id_(next_package_id_++), fill_min_(fill_min), fill_max_(fill_max), strategy_(strategy) {}
 
-Package::Package(std::string name, double fill_min, double fill_max, std::string strategy) : name_(name), id_(next_id_++), fill_min_(fill_min), fill_max_(fill_max), strategy_(strategy) {}
+Package::Package(int id, std::string name) : id_(id), name_(name), fill_min_(0), fill_max_(std::numeric_limits<double>::max()) {}
```
gonuke, line 59: Do we ever want the user to set the `id`? I think it's safer to always rely on `next_package_id_`?

src/material.cc
```diff
@@ -227,6 +227,24 @@ double Material::DecayHeat() {
...
+  Material::Ptr m_pkgd;
+
+  double fill_mass = pkg->GetFillMass(quantity());
+  if (fill_mass == 0) {
+    return ms_pkgd;
+  }
+
+  while (quantity() > pkg->fill_min()) {
+    double pkg_fill = std::min(quantity(), fill_mass);
+    m_pkgd = boost::dynamic_pointer_cast<Material>(ExtractRes(pkg_fill));
```
gonuke, line 241: seems like this is the only line that really forces this method to be replicated 3 times... I'll have to dust off my C++ polymorphism to see if there's a better way...

src/package.h
```diff
@@ -11,10 +11,18 @@ namespace cyclus {
...
 class Package {
-
+  friend class Context;
   public:
     typedef boost::shared_ptr<Package> Ptr;
 
+    // create a new package type with default values. Should be called by 
+    // the context only (see Context::AddPackage), unless you want an 
+    // untracked package type (which you probably don't)
+    static Ptr Create();
```
gonuke, line 21: You don't appear to call this anywhere anymore - still needed?

src/package.cc
gonuke: Also - can't this just be a call to `Package::Create("unpackaged")`?

src/product.cc
```diff
@@ -87,8 +87,8 @@ void Product::ChangePackageId(int new_package_id) {
     package_id_ = new_package_id;
     return;
   }
-  
-  Package::Ptr p = ctx_->GetPackageById(package_id_);
+  Package::Ptr p = ctx_->GetPackageById(new_package_id);
+  Package::Ptr p_old = ctx_->GetPackageById(package_id_);
```
gonuke, line 91: You never use `p_old`?

gonuke: I think you need to fix something similar in `Material::ChangePackageId()`

gonuke (changes requested):
I think the initialization of the unpackaged package in the context has gotten more complicated than it needs to... unless I'm forgetting something about how it all works.

src/package.cc
nuclearkatie: I was getting weirdness with testing where a new context was called but the numbering wasn't reset so it was numbering unpackaged packages with ids >1

src/package.h
nuclearkatie: See above. Maybe since it's not currently used, it can be safely removed and only re-introduced if the need arises

src/package.cc
gonuke: hmmm... I'll have to think about that!  I guess our package_id will not be unique across multiple contexts but we do want `unpackaged` to be singular across those contexts.

src/package.cc
```diff
@@ -4,71 +4,58 @@
 namespace cyclus {
 
 // unpackaged id is 1, so start the user-declared packaging id at 2
-int Package::next_id_ = 2;
+int Package::next_package_id_ = 2;
 
-template <class T>
-double Package::GetFillMass(typename T::Ptr r) {
-  if (r->quantity() < fill_min_) {
+Package::Ptr Package::Create() {
```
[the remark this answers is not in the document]
gonuke: First, we could put default values for each argument and get the same thing....

Second, do we want to allow that or not?  That's a philosophical/design question?  Why would a user want a different package to be effectively the same as unpackaged?

src/package.cc
gonuke: I see that you only need this to force the unpackaged_id in each context.

src/context.cc
```diff
@@ -84,9 +84,11 @@ Context::Context(Timer* ti, Recorder* rec)
       solver_(NULL),
       trans_id_(0),
       si_(0) {
-
         rng_ = new RandomNumberGenerator();
-      }
+        if (packages_.count("unpackaged") == 0) {
+          packages_["unpackaged"] = Package::CreateUnpackaged();
+        }
+}
```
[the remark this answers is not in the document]
nuclearkatie: It's at the same indentation as the line before and one tab more indented than the line where the curly braces were opened, which way do think it should be moved?

src/package.cc
gonuke: New idea: what if the unpackaged object is a static member of the Package class, and we return a pointer to that package when it is requested through the context:

1. Class `Package` has member `static Package::Ptr unpackaged_;`
2. At startup, we initialize this with `id_ = Package::unpackaged_id_` and `name_ = Package::unpackaged_name_`
3. If we request a package from the context with `name == Package::unpackaged_name_` or with `id == Package::unpackaged_id_` we return `Package::unpackaged_`

That is, this object is not actually stored in the context, but those methods check first to see if this is what has been  requested.

src/package.cc
nuclearkatie: That's also why I made `CreateUnpackaged` protected and gave no other pathway to use `Package(id, name)`, so there's not another easy way to use (misuse?) that constructor

src/package.cc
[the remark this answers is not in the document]
nuclearkatie: I guess the easy thing is to remove it and let someone else add it back in if it becomes important/useful

src/context.cc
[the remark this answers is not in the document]
gonuke: Was mostly noticing where it used to be on old line 86, before this change, which is the same level of indentation as the `si_(0)` initializer.

src/package.cc
gonuke: Yeah - I noticed that.  Just wondering if it's a big hammer for what we're after.

src/package.cc
nuclearkatie: A static member of the class that's also an instance of the class? I either don't fully understand your suggestion or don't understand how to successfully implement that

gonuke: See the PR I made into this branch

src/package.h
```diff
@@ -11,10 +11,17 @@ namespace cyclus {
...
+  friend class Context;
   public:
     typedef boost::shared_ptr<Package> Ptr;
 
+    // create a new package type. Should be called by the context only
+    // (see Context::AddPackage), unless you want an untracked package
+    //  type (which you probably don't)
+    static Ptr Create(std::string name, double fill_min = 0,
+                      double fill_max = std::numeric_limits<double>::max(),
+                      std::string strategy);
```
gonuke, line 23: All arguments with defaults have to come after those without.  I suggest a default for `strategy = "first"`.

gonuke (commented):
One error here

src/package.h
nuclearkatie: oops missed that one i guess

src/package.h
```diff
@@ -11,10 +11,17 @@ namespace cyclus {
 
 /// Packager is a class that packages materials into discrete items in ways that mimic realistic nuclear material handling. Packages will eventually be a required parameter of resources.
 class Package {
-
+  friend class Context;
```
gonuke, line 14: Don't think we need this anymore

src/context.h
```diff
@@ -251,9 +251,9 @@ class Context {
 
   /// Adds a package type to a simulation-wide accessible list.
   /// Agents should NOT add their own packages.
-  void AddPackage(std::string name, double fill_min = 0,
-                  double fill_max = std::numeric_limits<double>::max(),
-                  std::string strategy = "first");
+  Package::Ptr AddPackage(std::string name, double fill_min = 0,
```
gonuke, line 254: Is there a reason that this returns a Ptr to the package now? I don't see it used anywhere. Not a reason to block, but curious.

src/resource.cc
```diff
@@ -10,4 +10,22 @@ void Resource::BumpStateId() {
...
+  Resource::Ptr r_pkgd;
+
+  double fill_mass = pkg->GetFillMass(quantity());
+  if (fill_mass == 0) {
+    return rs_pkgd;
+  }
+
+  while (quantity() > pkg->fill_min()) {
+    double pkg_fill = std::min(quantity(), fill_mass);
+    r_pkgd = ExtractRes(pkg_fill);
```
gonuke, line 24: This method is explicitly undefined for `Resource` objects.  I don't think we can/should have a `Package` method either.

See comment in the header.

src/resource.h
```diff
@@ -98,6 +98,11 @@ class Resource {
   /// Changes the product's package id
   virtual void ChangePackageId(int new_package_id = Package::unpackaged_id()) {};
 
+  /// Repackages a single resource into a package. If some quantity of the 
+  /// resource cannot be packaged using the given packaging strategy and
+  /// restrictions, the remainder is left in the resource object. 
+  std::vector<Resource::Ptr> Package(Package::Ptr pkg);
```
gonuke, line 104: This should be pure virtual:
```suggestion
  virtual std::vector<Resource::Ptr> Package(Package::Ptr pkg) = 0;
```

src/product.h
```diff
@@ -77,6 +77,9 @@ class Product : public Resource {
   /// Changes the product's package id
   virtual void ChangePackageId(int new_package_id = Package::unpackaged_id());
 
+  // Packages the product
+  std::vector<Product::Ptr> Package(Package::Ptr pkg);
```
gonuke, line 81: This should be virtual
```suggestion
  virtual std::vector<Product::Ptr> Package(Package::Ptr pkg); 
```

src/material.h
```diff
@@ -161,6 +161,9 @@ class Material: public Resource {
   /// type minimum and maximum mass criteria.
   virtual void ChangePackageId(int new_package_id = Package::unpackaged_id());
 
+  /// Packages the material
+  std::vector<Material::Ptr> Package(Package::Ptr pkg);
```
gonuke, line 165: This should be virtual
```suggestion
  virtual std::vector<Material::Ptr> Package(Package::Ptr pkg);
```

gonuke (changes requested):
Now that we've resolved the main design questions, a few final cleanup changes.

src/context.h
nuclearkatie: Nothing uses this anymore except in testing (which I'll change)

src/product.h
nuclearkatie: I guess this raises the problem that packaging returns different types in `Material`/`Product` than it does for `Resource`

src/material.h
```diff
@@ -161,6 +161,9 @@ class Material: public Resource {
   /// type minimum and maximum mass criteria.
   virtual void ChangePackageId(int new_package_id = Package::unpackaged_id());
 
+  /// Packages the material
+  virtual std::vector<Material::Ptr> Package(Package::Ptr pkg);
```
gonuke, line 165: Looks like this needs to be:
```suggestion
  virtual std::vector<Resource::Ptr> Package(Package::Ptr pkg);
```

src/product.h
```diff
@@ -77,6 +77,9 @@ class Product : public Resource {
   /// Changes the product's package id
   virtual void ChangePackageId(int new_package_id = Package::unpackaged_id());
 
+  // Packages the product
+  virtual std::vector<Product::Ptr> Package(Package::Ptr pkg);
```
gonuke, line 81: This one also needs to be
```suggestion
  virtual std::vector<Resource::Ptr> Package(Package::Ptr pkg); 
```

gonuke (changes requested):
These comments have the fix for the current issue, but not sure what it will mean for nuances when we call these methods.

tests/resource_tests.cc
```diff
@@ -99,3 +101,45 @@ TEST_F(ResourceTest, ProductExtractGraphid) {
...
+  std::vector<Product::Ptr> p3_pkgd = p3->Package<Product>(pkg);
+
+  // everything stays in old product, with same (default) package id
+  EXPECT_EQ(p3->package_id(), Package::unpackaged_id());
+  EXPECT_EQ(p3->quantity(), 0.5);
+
+  // all packaged
+  std::vector<Product::Ptr> p1_pkgd = p1->Package<Product>(pkg);
+  EXPECT_EQ(p1->quantity(), 0);
+  EXPECT_EQ(p1_pkgd[0]->package_id(), pkg_id);
```
gonuke, line 144: should we also check the number of packages?  If I read this right, `p1_pkgd` should be length one: all of its 3 units of quantity in a single package.

We could also package `p2` and its 7 units would end up in two packages of size 5 and 2

src/context.h
```diff
@@ -252,8 +252,8 @@ class Context {
   /// Adds a package type to a simulation-wide accessible list.
   /// Agents should NOT add their own packages.
   void AddPackage(std::string name, double fill_min = 0,
-                  double fill_max = std::numeric_limits<double>::max(),
-                  std::string strategy = "first");
+                          double fill_max = std::numeric_limits<double>::max(),
+                          std::string strategy = "first");
```
gonuke, line 256: this got extra indented as far as I can see.

gonuke (changes requested):
Suggest a couple of additional tests.

tests/resource_tests.cc
nuclearkatie: you'll never guess what was in my stashes.... though I didn't I have the two packages case

gonuke (approved):

PR #1734, opened 2024-04-12
Move apt-update to <pkg_mgr>-deps build stage

### LINKED ISSUE #1733: "Unable to fetch some archives" in CI
In the cyclus Dockerfile we run `apt update` in the first build stage `common-deps`.  If you modify the dependency list in the following `apt-deps` build stage you may run into the error `Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?` - [example here](https://github.com/bennibbelink/cyclus/actions/runs/8656566165/job/23761006815).

I think this is due to the `common-base` build stage being pulled in from our CI layer cache, then the `apt install -y ...` command in the `apt-deps` stage is working with outdated package indexes.

bennibbelink:
Closes #1733.  Whenever the dependency list changes the `RUN` step will have to re-run and `apt update` the package indexes.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,7 +38,7 @@ Since last release
 **Changed:**
 
 * Moved to unified CHANGELOG Entry and check them with GithubAction (#1571)
-* Major update and modernization of build (#1587, #1632)
+* Major update and modernization of build (#1587, #1632, #1734)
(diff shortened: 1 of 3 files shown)


gonuke (approved):
Seems reasonable - thanks @bennibbelink

PR #1735, opened 2024-04-12
Update Dependency tables in DEPENDENCIES.rst

### LINKED ISSUE #1710: New Dependency Table Column
In the table of dependencies it might be helpful to add a new column to identify which options require which dependencies.

bennibbelink:
Closes #1710. This PR updates the dependency tables with correct lists of **required** and **optional** dependencies, along with notes about the optional dependencies.  

It also contains some minor build changes that allow users to build without Cython present on their machine.

### DIFF
--- a/DEPENDENCIES.rst
+++ b/DEPENDENCIES.rst
@@ -22,20 +22,19 @@ In order to facilitate future compatibility with multiple platforms,
 Cyclus is built using ``CMake``. A full list of the Cyclus package
 dependencies is shown below:
 
-====================   ==================
-Package                Minimum Version
-====================   ==================
-``CMake``                3.16.3
-``boost``                1.71.0
-``libxml2``              2.9.10+
-``libxml++``             2.40.1
-``python``               3.8.10
-``sqlite3``              3.31.1
-``HDF5``                 1.10.4+
-``Coin-Cbc``             2.10.3
-``Coin-Clp``             1.17.5
-``zlib``                 1.2.11
-====================   ==================
+====================      ==================
+Package                   Minimum Version
+====================      ==================
+``CMake``                   3.16.3
+``pkg-config``              0.29.1
+``boost``                   1.71.0
+``libxml2``                 2.9.10
+``libxml++``                2.40.1
+``python`` (dev version)    3.8.10
+``sqlite3``                 3.31.1
+``HDF5``                    1.10.4
+``LAPACK``                  3.9.0
+====================      ==================
 
 .. website_include_end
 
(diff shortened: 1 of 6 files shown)


bennibbelink:
It also updates the `conda install <dependencies>` instruction in the README, but once the conda-forge feedstock is up to date I think we can change the instructions to `conda install -y cyclus`

tests/toolkit/infile_converters_tests.cc
```diff
@@ -2,6 +2,8 @@
 
 #include "cyclus.h"
 
+#ifdef CYCLUS_WITH_PYTHON
```
gonuke, line 5: Do all these tests require python? Including the XML<->JSON?

gonuke (commented):
Just a confirmation/clarification question about all infile converter tests

tests/toolkit/infile_converters_tests.cc
bennibbelink: No, the XML<->JSON does not require python.  I moved the `#ifdef CYCLUS_WITH_PYTHON` to encompass only the tests requiring python.  Thanks for catching that @gonuke!

docker/Rocky.dockerfile
```diff
@@ -29,10 +29,8 @@ RUN dnf update -y && dnf install -y \
         hdf5-devel \
         libxml2-devel \
         boost-devel \
-        blas-devel \
```
abachma2, line 32: Were there only 2 packages to remove from the list for Rocky?

DEPENDENCIES.rst
```diff
@@ -22,45 +22,43 @@ In order to facilitate future compatibility with multiple platforms,
...
-Pandas                 any
-====================   ==================
+====================   ==================  =============================================  ==================        
+Package                Minimum Version     Purpose                                        Notes
+====================   ==================  =============================================  ==================  
+``Coin-Cbc``             2.10.3             Enables use of Branch-and-Cut solver          Cyclus must be built with ``--allow-milps`` flag
+``Coin-Clp``             1.17.5             Enables use of Linear Programming solver      Cyclus must be built with ``--allow-milps`` flag
+``doxygen``              1.7.6.1            Building documentation
+``TCMmalloc``            any                Improves performance                          Only used if Cython is not present
+``Cython``               0.29               Enables use of Python agents and input files
```
abachma2, line 52: I don't think these get separated out properly when this renders. I would go back and check on that.

abachma2 (commented):
Hi @bennibbelink Thanks for taking care of all of this. I have a few small comments on some sections.

docker/Rocky.dockerfile
bennibbelink: Correct, only 2 to remove.  However, I did move the installation of `wget` to the `libxmlpp` stage because that is where it is needed - thanks for making me take a closer look.

DEPENDENCIES.rst
bennibbelink: I believe I fixed the table spacing so they both show up in the website now (tested locally but please double check).  I also removed the note about installing the Python development version as the Python C API is included with standard python installations now

abachma2: The issue I found with these lines is when GitHub renders the table, not the website. The Coin-clp and doxygen end up in the same line, and the TCMmalloc and Cython end up in the same line, as shown in the figure. My guess is because the last column is blank for the doxygen and cython lines.
Screenshot (3)

bennibbelink: Ah I see.  Just added `N/A` in the blank cells and it is rendering correctly, is that an acceptable solution?

abachma2 (approved):
Thanks for making those changes @bennibbelink. I'm approving and merging this PR.

gonuke (approved):
Thanks for the careful/thorough review of these

PR #1737, opened 2024-04-17
Python install modifications for compatibility with conda-forge

bennibbelink:
I think our python installation script is more complex than is necessary and it is causing builds in `conda-forge/cyclus` to fail at pytests due to not being able to find the `cyclus` package (stemming from incorrect installation).

My understanding of the issue is as follows:
- Conda has environment variables to define the installation location (e.g. PREFIX, SP_DIR) to support the portability of our package. These variables are kept as placeholders throughout the build process (you can see this in the install location in the build output - Installing <file> to PREFIX/bin).
- In the cyclus build we currently generate `SetupPyInstall.cmake` based on some vars in the build. I think because we configure this during the build (substitute the @@ variables) the placeholder used by conda was not being kept as a placeholder, rather it was expanded out to the absolute path instead of SP_DIR/cyclus (as I saw in build output)
This PR eliminates the need for configuring a separate CMake script by replacing the `SCRIPT` argument with [`CODE`](https://cmake.org/cmake/help/latest/command/install.html#code).  

We can then pass in the conda placeholder via `-DPYTHON_SITE_PACKAGES`


bennibbelink:
Ok I had to make one more change (which is really a change to cycamore, but I replicated it here for consistency).

`FindCyclus.cmake` finds `libcyclus.so` for cycamore to link against.  In the conda cycamore builds the cython generated libraries (`eventhooks.so`, `pyinfile.so`, and `pymodule.so`) were not being linked agains because they were not being reported by `FindCyclus.cmake`.  I will make a PR in cycamore to implement this - I have it working locally with the two respective branches, but in order to see CI pass we will need to merge these changes in cycamore first.

This was not a problem before because we always export `LD_LIBRARY_PATH` to `/root/.local/lib` which is the location of these three cython generated libraries.  Most literature I can find considers the use of `LD_LIBRARY_PATH` to be relatively unsafe.  With these build changes we no longer need to set `LD_LIBRARY_PATH` for the linker to find all the libraries it needs.

bennibbelink:
Hmm it looks like making the changes to `FindCyclus.cmake` in this PR fixed the issue.  I now see that Cycamore's `CMakeLists.txt` we set the `CMAKE_MODULE_PATH` to look in `/root/.local/share/cyclus/cmake` which has the Cyclus version of `FindCyclus.cmake`.  Thus modifying this repo's version of the file fixes our issue, but we should make these same changes in cycamore ([#596](https://github.com/cyclus/cycamore/pull/596)).

cmake/FindCyclus.cmake
gonuke, line 1: Should we be installing this somewhere standard so that cycamore can use it without needing its own copy that we have to keep synchronized with this?

gonuke (commented):
Thanks @bennibbelink - this all looks reasonable.  Just one question about installing `FindCyclus.cmake` "properly"?

cmake/FindCyclus.cmake
bennibbelink: By "somewhere standard" you mean somewhere like `/root/.local/share/cyclus/cmake`?  Because we already do that (and then reference these modules in the Cycamore build at `CMakeLists.txt` line 86 in Cycamore.  We could likely delete `FindCyclus.cmake` from the cycamore repo - I will test that out to confirm

bennibbelink: My suspicions were correct - see Cycamore [#597](https://github.com/cyclus/cycamore/pull/597)

gonuke (approved):
Thanks for the update @bennibbelink

PR #1738, opened 2024-04-20
Update GTest

bennibbelink:
According to [CEP3](http://fuelcycle.org/cep/cep3.html#maintainence-tasks) we should update GTest prior to release.  The linked instructions for fusing GTest just points to the googletest GitHub repo, and I wasn't able to find instructions for fusing GTest.  However, I followed [instructions from GoogleTest](https://google.github.io/googletest/quickstart-cmake.html) (and review different installation options in the [googletest README](https://github.com/google/googletest/tree/main/googletest)) and it appears that the method with the least limitation (and easiest maintenance) is to fetch the GTest source code within the CMake build.

This PR makes the following changes:
- Modifies the CMake build to fetch the GTest source automatically (and build the gtest library as we did before)
- Delete the `tests/GoogleTest` directory as it is no longer necessary to maintain our own GTest source
- `GTEST_HAS_PARAM_TEST` is no longer defined because gtest [always supports parameterized test fixtures](https://github.com/google/googletest/issues/360).  A handful of tests had checks for this, those checks have been removed.


bennibbelink:
Currently this PR fetches googletest `v1.14.0` (the latest release)

bennibbelink:
The coverage check was failing due a decrease in coverage.  I took a closer look at the coveralls report and there were odd coverage changes.  According to [this stackoverflow post](https://stackoverflow.com/questions/42003783/lcov-gcov-branch-coverage-with-c-producing-branches-all-over-the-place) running `lcov` with branch coverage can lead to many branches where scope exits are possible (like when an exception could be thrown).  I think I removed to proper `lcov` flags for branch coverage - we may want to consider moving forward without branch coverage.

bennibbelink:
This is ready for review, but note that CI is not passing currently.  See https://github.com/cyclus/cycamore/pull/598#issuecomment-2067778328 for the sequence of events that needs to happen.

tests/CMakeLists.txt
```diff
@@ -7,35 +7,17 @@ SET(GENERATED_TEST_LIST "" CACHE INTERNAL "")
...
-IF(CMAKE_USE_PTHREADS_INIT)  # The pthreads library is available.
-    SET(cxx_base_flags "${cxx_base_flags} -DGTEST_HAS_PTHREAD=1")
-ENDIF()
-
-# Build libgtest
-SET(GTest
-    "${CMAKE_CURRENT_SOURCE_DIR}/GoogleTest/gtest/gtest-all.cc"
-    )
+include(FetchContent)
+FetchContent_Declare(
```
gonuke, line 11: Does CMake do any caching/checking for existing installs/fetches? Or does this mean that every time CMake is invoked, the ZIP file is fetched?  Seems OK for an occasional install, but annoying/a lot for developers, and even more for CI. 

Obviously the existence of `FetchContent` means this is some kind of best (common?) practice.  Should we be worried about it?

gonuke (commented):
Thanks @bennibbelink - definitely work a discussion on the idea of fetching Gtest with every build.

tests/CMakeLists.txt
bennibbelink: It will cache the source files and build files in separate subdirectories of our build directory (`_deps/googletest-src` and `_deps/googletest-build`, respectively).  If you re-run CMake it accesses this cache.  My understanding is that ZIP will only be fetched if you run a clean build (deleting the fetched cached source/build).  It automatically hides a lot of the logging so the build output isn't swamped, and in my experience building locally I have not noticed any slowdown from having to fetch the ZIP.

I don't think we should be worried about it.  This is the method GoogleTest suggests in their [CMake Quickstart](https://google.github.io/googletest/quickstart-cmake.html), and CMake even uses GoogleTest as the example in their [FetchContent docs](https://cmake.org/cmake/help/latest/module/FetchContent.html) and [Using Dependencies Guide](https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#downloading-and-building-from-source-with-fetchcontent) so I am pretty confident that it is (and will continue to be) well supported.

bennibbelink: > I have not noticed any slowdown from having to fetch the ZIP

Expanding on this a little.  The FetchContent docs say:.

 > "Whereas [ExternalProject_Add()](https://cmake.org/cmake/help/latest/module/ExternalProject.html#command:externalproject_add) downloads at build time, the FetchContent module makes content available immediately, allowing the configure step to use the content in commands like [add_subdirectory()](https://cmake.org/cmake/help/latest/command/add_subdirectory.html#command:add_subdirectory), [include()](https://cmake.org/cmake/help/latest/command/include.html#command:include) or [file()](https://cmake.org/cmake/help/latest/command/file.html#command:file) operations."

> "The [FetchContent_MakeAvailable()](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable) command ensures the named dependencies have been populated, either by an earlier call or by populating them itself"


I interpret that to mean it fetches content in the background while the CMake build continues to run (until you call `FetchContent_MakeAvailable()`).  I am not sure how optimized this is or whether we take full advantage of this functionality in the build, but I don't see the build hanging anywhere while waiting to fetch the ZIP.

bennibbelink: Just built locally with `FetchContent` in verbose mode and can confirm that the population happens very fast - the ZIP is only 1MB.  If we were fetching a larger project I might suggest re-ordering our CMake so that we could take advantage of background fetching but I don't think that is necessary here.

gonuke: Thanks so much for the thorough information @bennibbelink

gonuke (approved):
Thanks for another step forward @bennibbelink

PR #1741, opened 2024-04-30
Default to Release build

bennibbelink:
Updates `install.py` to set `-DCMAKE_BUILD_TYPE=Release` if no `--build-type` is specified.  Removes this flag from the dockerfiles where we install cyclus.


install.py
```diff
@@ -79,6 +79,8 @@ def install_cyclus(args):
                           ]
         if args.build_type:
             cmake_cmd += ['-DCMAKE_BUILD_TYPE=' + args.build_type]
+        else:
+            cmake_cmd += ['-DCMAKE_BUILD_TYPE=Release']
```
gonuke, line 83: Or should we set a default in the argparse declaration of this CLI parameter?

gonuke (changes requested):
If we set the default in argparse it will be visible to users

install.py
bennibbelink: That is a better solution, thanks for the suggestion @gonuke

gonuke:
Do we need the same change for Cycamore?

bennibbelink:
Yes, please see https://github.com/cyclus/cycamore/pull/600

gonuke (approved):

PR #1743, opened 2024-05-03
Support MacOS

bennibbelink:
This PR includes:
- [x]  a handful of CMake and code modifications to support building cyclus on MacOS (x86 **and** ARM)
- [x] some "qualify of life" modifications - mostly to do with how we reference python/python3 (eliminates the need to `update-alternatives` and have `python` defined in our environment as well as `python3`)
- [x] CI to test the build on MacOS
- [x] Downstream testing and PR comment
- [x] Updates documentation (in this repo, documentation updates still required in cyclus/cyclus.github.com)

Changes to be made before merge:
- [ ] do not rely on DYLD_LIBRARY_PATH
- [x] separate workflow into different files by platform
- [ ] currently references `macos` branch in cycamore and cymetric, merge PRs in those repos first then change the ref in this PR back to `main`


bennibbelink:
This is relying on setting DYLD_LIBRARY_PATH, now that I have tests passing I will try to eliminate that step

docker/Dockerfile
```diff
@@ -36,9 +36,8 @@ RUN apt update -y && apt install -y \
 
 RUN apt install -y python3-pprintpp; exit 0
 RUN apt clean -y all
-
 RUN update-alternatives --install /usr/bin/python python /usr/bin/python3 10
```
gonuke, line 39: We've always relied on users doing this to ensure that a "bare" `python` is on the path.  Looking at all your updates for MacOS, I wonder if it's best practice to always specify `python3`.  It seems clunky to me, but maybe too many systems keep python 2.x around as `python`?

.github/workflows/build_test.yml
```diff
@@ -189,3 +167,140 @@ jobs:
           cache-from: type=registry,ref=ghcr.io/cyclus/cyclus_rocky_${{ matrix.rocky_versions }}/cyclus:ci-layer-cache
           build-args: |
             rocky_version=${{ matrix.rocky_versions }}
+
+
+  build-test-macos:
```
gonuke, line 172: Thoughts on having these different platform builds in different workflows?  We've done that in other projects, but I'm not sure what a good practice is?  One benefit is that they can run in parallel.... but not sure of other benefits or downsides.

.github/workflows/build_test.yml
```diff
@@ -189,3 +167,140 @@ jobs:
...
+          cd tests && python3 -m pytest
+
+      - name: Checkout Cymetric
+        if: ${{ github.event_name == 'pull_request' && steps.build-cycamore.outcome == 'success' }}
+        uses: actions/checkout@v4
+        with:
+          repository: cyclus/cymetric
+          path: ${{ github.workspace }}/cymetric
+
+      - name: Build and Test Cycamore
```
gonuke, line 244:
```suggestion
      - name: Build and Test Cymetric
```

tests/toolkit/matl_buy_policy_tests.cc
```diff
@@ -486,7 +486,7 @@ TEST_F(MatlBuyPolicyTests, NormalActiveDormant) {
 TEST_F(MatlBuyPolicyTests, MixedActiveDormant) {
   using cyclus::QueryResult;
   
-  boost::shared_ptr<NormalIntDist> a_dist = boost::shared_ptr<NormalIntDist>(new NormalIntDist(5, 1, 0, 1e299));
+  boost::shared_ptr<NormalIntDist> a_dist = boost::shared_ptr<NormalIntDist>(new NormalIntDist(5, 1, 0, std::numeric_limits<int>::max()));
```
gonuke, line 489: Was this necessary somehow, or is this just some cleanup?

docker/Dockerfile
bennibbelink: I agree, I think it would be better to not rely on a "bare" `python`.  I initially removed that line in the Dockerfile, but it will require appropriate adjustments in the downstream projects.  Given that there are other changes required for the downstream projects I will remove it again, and make PRs in cycamore and cymetric and we can merge them in sync.

.github/workflows/build_test.yml
bennibbelink: - As is (multiple jobs in the same workflow) the jobs do run in parallel.  They only run serially if you specify dependencies using the `needs` key
- I like the idea of separating concerns and having different platforms in different workflows.  That feels like better practice to me.  The only significant change I can think of is that [this comment](https://github.com/cyclus/cyclus/pull/1743#issuecomment-2092064703) will need to be split apart - one for each platform

bennibbelink: Should I pull Rocky out to its own workflow or leave it in `build_test.yml`?  (can be a separate PR)

install.py
maxschalz, line 1: Minor comment: It might be helpful to add the 'required on Apple Silicon' comment to the `--slow` flag, similarly to how it's done in `INSTALL.rst`

This would change [lines 210-213](https://github.com/cyclus/cyclus/pull/1743/files#diff-a5fa297e66b71ac1a8685caa5c37577774be2c36024ecda7abeb70bb50214d0fL210-L213) to:
```python3
    parser.add_argument('--slow', dest='fast',
                        action='store_false', help="Will NOT try to compile "
                        "from assembly, if possible. This is slower as it "
                        "must compile from source. **This is required to "
                        "build on Apple Silicon**")
```

INSTALL.rst
maxschalz, line 1: Re. 'Run the install script' ([lines 30 to 35](https://github.com/cyclus/cyclus/pull/1743/files#diff-ce342037df7e6de102c3157b708a389b128142aae879421094c71d9f97f5084cR30-R35)).

You might want to add specific instructions for MacOSX right away, e.g.:
```
python3 install.py --slow -DCMAKE_CXX_FLAGS="-ld_classic"
```
The second flag is relevant if one uses Apple's XCode v15.x (which is the current version). Without that flag, 'duplicate library'-errors can arise:       
```
[  1%] Linking CXX shared library ../lib/eventhooks.dylib
ld: warning: ignoring duplicate libraries: '-ldl', '-lm'
ld: multiple errors: duplicate LC_RPATH '@loader_path' in '/Users/maxschalz/mambaforge/envs/cyclus-env/lib/libgfortran.5.dylib'; duplicate LC_RPATH '@loader_path' in '/Users/maxschalz/mambaforge/envs/cyclus-env/lib/libgfortran.5.dylib'; duplicate LC_RPATH '@loader_path' in '/Users/maxschalz/mambaforge/envs/cyclus-env/lib/libgfortran.5.dylib'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

maxschalz, line 1: Re. 'Run the install script' ([lines 30 to 35](https://github.com/cyclus/cyclus/pull/1743/files#diff-ce342037df7e6de102c3157b708a389b128142aae879421094c71d9f97f5084cR30-R35)).

You might want to add specific instructions for MacOSX right away, e.g.:
> If you are using Mac OSX, add the following flags when running the install script:
> ```
> python3 install.py --slow -DCMAKE_CXX_FLAGS="-ld_classic"
> ```
The second flag is relevant if one uses the current version (v15.x) of XCode (XCode is Apple's developer toolset, including for example C/C++ compilers). Without that flag, 'duplicate library'-errors can arise (see [this SO-thread](https://stackoverflow.com/a/77330350)):       
```
[  1%] Linking CXX shared library ../lib/eventhooks.dylib
ld: warning: ignoring duplicate libraries: '-ldl', '-lm'
ld: multiple errors: duplicate LC_RPATH '@loader_path' in '/Users/maxschalz/mambaforge/envs/cyclus-env/lib/libgfortran.5.dylib'; duplicate LC_RPATH '@loader_path' in '/Users/maxschalz/mambaforge/envs/cyclus-env/lib/libgfortran.5.dylib'; duplicate LC_RPATH '@loader_path' in '/Users/maxschalz/mambaforge/envs/cyclus-env/lib/libgfortran.5.dylib'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

install.py
maxschalz, line 1: Minor comment: It might be helpful to add the 'required on Apple Silicon' comment to the `--slow` flag, similarly to how it's done in `INSTALL.rst`

This would change [lines 210-213](https://github.com/cyclus/cyclus/pull/1743/files#diff-a5fa297e66b71ac1a8685caa5c37577774be2c36024ecda7abeb70bb50214d0fL210-L213) to:
```python3
    parser.add_argument('--slow', dest='fast',
                        action='store_false', help="Will NOT try to compile "
                        "from assembly, if possible. This is slower as it "
                        "must compile from source. **This is required to "
                        "build on Apple Silicon**")
```

bennibbelink: Good catch, thanks @maxschalz

tests/toolkit/matl_buy_policy_tests.cc
bennibbelink: Trying to get cycamore working on macos and running into issues with the random number generator again.  We use 1e299 many places across the codebases.  Is it just intended to represent a really large number or is there another justification to this that I'm not aware of?  If the former is true should we make a separate issue/PR to follow best practices?

gonuke: Yes. I agree. We could also declare a `cyclus_large` somewhere and set it to the max int, and use that to be both descriptive and allow a single point of changing it

bennibbelink:
Downstream testing is working with a brew-based install - depending on solutions for #1757 and https://github.com/cyclus/cycamore/issues/607.


Still working on a conda-based install.  Like we've discussed before there is value to having conda-based CI in place for compatibility with conda-forge feedstocks.

bennibbelink:
I have gotten a conda based install (mostly) working locally on https://github.com/bennibbelink/cyclus/tree/macos-cyclus-large, but am having trouble getting it to pass tests in CI.  It seems as if locally cmake is using the system's `libc++` but on the GitHub runner it is using `libc++` in the conda environment (which apparently might not support pthread).

Solved a big blocker of how to link with python correctly in a conda environment on macOS, so definite progress has been made.

bennibbelink:
Closing this in favor of #1814.

PR #1744, opened 2024-05-05
Update README conda installation instructions

### LINKED ISSUE #1736: README installation instructions
After updating the conda-forge feedstock we should update the install instructions in the README.  I propose we have two sets of instructions: 
- For users `conda install -y cyclus`
- For developers `conda install --only-deps -y cyclus`, then build from source

bennibbelink:
Closes #1736.  The current instructions in our README say the quickest way to install cyclus is by installing all conda dependencies then build from source.  This changes those instructions to simply `conda install -y cyclus`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -78,38 +78,28 @@ Cyclus is built using ``CMake``. For detailed instructions on Cyclus dependencie
 Quick Cyclus Installation
 *************************
 The quickest way to install Cyclus and its dependencies relies on using the `conda-forge` channel and the `conda` package manager within the Anaconda python environment.  The following instructions guide you through that approach.
-To install Cyclus and its dependencies onto a clean Ubuntu machine (tested on 18.04 LTS):
+To install Cyclus and its dependencies onto a clean Ubuntu machine (tested on 22.04 LTS):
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -78,38 +78,33 @@ Cyclus is built using ``CMake``. For detailed instructions on Cyclus dependencie
...
-
-- Use ``sudo apt install`` to install and configure git
-
-- Clone the Cyclus repository by running ``git clone https://github.com/cyclus/cyclus.git``
-
-- Navigate to the folder containing Cyclus
-
-- Run the command ``python install.py``
+  conda config --add channels conda-forge
+  conda create -n cyclus python=3.11
```
bennibbelink, line 105: For some reason our conda-forge package was not published for Python 3.12, so this needs to be pinned to 3.8, 3.9, 3.10, or 3.11.  I will look into why that is happening

bennibbelink: The conda-forge bot figured this out before I even had a chance to check it out.  Should work with python3.12 once the new conda-forge package is published but I will confirm tomorrow

bennibbelink: Can confirm these instructions work without pinning python.

PR #1747, opened 2024-05-08
Use `python3` instead of `python` in environment

bennibbelink:
No longer necessitate the existence of a version agnostic `python` in the environment.  This should be considered with https://github.com/cyclus/cycamore/pull/602 and https://github.com/cyclus/cymetric/pull/196


bennibbelink:
Currently references the PR branches in my forks to show that it will be compatible with downstream changes in those PRs.  When we are happy with these changes we should merge the downstream PRs first and I will change this workflow back to reference `main`

bennibbelink:
That was fast thanks @gonuke! I'll revert the workflow changes and we should see CI pass 🤞

gonuke:
Yeah - I should have looked more carefully and merged them in a better order ...

gonuke (approved):
Thanks @bennibbelink

PR #1748, opened 2024-05-11
CMake changes to clean up conda-forge dependencies

bennibbelink:
See https://github.com/conda-forge/cyclus-feedstock/pull/80 and https://github.com/conda-forge/cyclus-feedstock/pull/81 for more dialogue about these changes.  Notably:
- remove `pandas` as a build-time dependency
- update how we link against `libxml++` libs
- update how we find `blas` libs
- use `-` consistently in `install.py` flags instead of `_` (might necessitate some documentation updates in the website)


CMakeLists.txt
```diff
@@ -331,7 +330,6 @@ IF(NOT CYCLUS_DOC_ONLY)
         include(UseCython)
         find_package (Python3 REQUIRED COMPONENTS NumPy)
         find_package(Jinja2 REQUIRED)
```
gonuke, line 332: Is jinja also run-time only?

bennibbelink: No, `jinja2` is used by `gentypesystem.py` during the build process

CMakeLists.txt
```diff
@@ -207,7 +208,6 @@ IF(NOT CYCLUS_DOC_ONLY)
       link_directories(${HDF5_LIBRARY_DIRS})
     endif()
     set(LIBS ${LIBS} ${HDF5_C_LIBRARIES} ${HDF5_C_HL_LIBRARIES})
-    MESSAGE("--    HDF5 Root: ${HDF5_ROOT}")
```
gonuke, line 210: Do we want to delete this message? (and the BOOST one?)

gonuke (changes requested):
One question about messages

CMakeLists.txt
bennibbelink: I added them back in.

gonuke (approved):
Thanks @bennibbelink - LGTM

PR #1749, opened 2024-05-13
Add packaging to matl sell policy

nuclearkatie:
Material sell policy can be initialized with a package type. When responding to requests, checks against packaging restrictions. When returning trades, packages material popped from buffer.


src/toolkit/matl_sell_policy.cc
```diff
@@ -100,6 +112,13 @@ void MatlSellPolicy::Start() {
     ss << "No manager set on Sell Policy " << name_;
     throw ValueError(ss.str());
   }
+  if ((quantize_ < package_->fill_min()) || 
+  (quantize_ > package_->fill_max()))  {
```
gonuke, line 116: Could you do this in `Init()`?

src/toolkit/matl_sell_policy.cc
```diff
@@ -153,8 +173,15 @@ std::set<BidPortfolio<Material>::Ptr> MatlSellPolicy::GetMatlBids(
     for (rit = requests.begin(); rit != requests.end(); ++rit) {
       req = *rit;
       qty = std::min(req->target()->quantity(), limit);
-      nbids = excl ? static_cast<int>(std::floor(qty / quantize_)) : 1;
-      qty = excl ? quantize_ : qty;
+      package_fill = package_->GetFillMass(qty);
```
gonuke, line 176: Could you do something like:
```
double bid_qty = excl ? quantize_ : package_->GetFillMass(qty);
```
and the use `bid_qty` throughout?

src/toolkit/matl_sell_policy.cc
```diff
@@ -153,8 +173,15 @@ std::set<BidPortfolio<Material>::Ptr> MatlSellPolicy::GetMatlBids(
...
+        nbids = 0;
+      } else {
+        nbids = excl ? static_cast<int>(std::floor(qty / quantize_)) : static_cast<int>(std::floor(qty / package_fill));
+      }
+      double remaining_qty = fmod(qty, package_fill);
+
+      qty = excl ? quantize_ : package_fill;
       for (int i = 0; i < nbids; i++) {
         m = buf_->Pop();
         buf_->Push(m);
```
gonuke, line 187: This isn't your code, but shouldn't we do this before the loop?

...and can't we `Peek()` instead of `Pop()` and `Push()`?

Also, what if multiple items need to be popped to fulfill an order?  Let's discuss...

src/toolkit/matl_sell_policy.cc
```diff
@@ -153,16 +174,30 @@ std::set<BidPortfolio<Material>::Ptr> MatlSellPolicy::GetMatlBids(
     for (rit = requests.begin(); rit != requests.end(); ++rit) {
       req = *rit;
       qty = std::min(req->target()->quantity(), limit);
-      nbids = excl ? static_cast<int>(std::floor(qty / quantize_)) : 1;
-      qty = excl ? quantize_ : qty;
+      bid_qty = excl ? quantize_ : package_->GetFillMass(qty);
+      if (bid_qty == 0) {
+        nbids = 0;
+      } else {
+        nbids = excl ? static_cast<int>(std::floor(qty / quantize_)) : static_cast<int>(std::floor(qty / bid_qty));
```
gonuke, line 181: Now we can use `bid_qty` without the conditional check here

gonuke (changes requested):

src/toolkit/matl_sell_policy.cc
```diff
@@ -178,9 +215,17 @@ void MatlSellPolicy::GetMatlTrades(
...
+    // push any extra material that couldn't be packaged back onto buffer
+    buf_->Push(mat);
+    if (mat_pkgd.size() == 0) {
+      // material couldn't be packaged, e.g. if the accepted bid was < fill min
+      return;
+    }
+    if (ignore_comp_) {
+      mat_pkgd[0]->Transmute(it->request->target()->comp());
+      }
+    responses.push_back(std::make_pair(*it, mat_pkgd[0]));
```
nuclearkatie, line 228:
```suggestion
    if (mat_pkgd.size() > 0) {
      if (ignore_comp_) {
        mat_pkgd[0]->Transmute(it->request->target()->comp());
        }
     responses.push_back(std::make_pair(*it, mat_pkgd[0]));
}
```

nuclearkatie:
Ok unfortunately pulling this one back into drafts because as I'm developing it with [cycamore#603](https://github.com/cyclus/cycamore/pull/603) something odd is happening. It seems like everything is working well, but then when I try to run a toy test input, like the one [here](https://github.com/nuclearkatie/dissertation-toy-problems/blob/main/packaging/toy-packages.xml), the simulation aborts... because it can't find any packages. 

However, this appears to happen _after_ the package is already found once successfully. I've so far identified that `context::GetPackagesByName()` is getting called once successfully, and then a second time where it claims to have no packages in its list at all. I'm not yet sure why `context::GetPackagesByName()` is even being called more than once in a simulation with a single agent using the sell policy and packages (one Storage)??

I'm a bit at a loss here... More digging is definitely needed, it might be the case that I'm missing something in the setup of packages because I'm just not that familiar with the process of spinning up a full simulation

nuclearkatie:
Ok, I think I understand:
1. The XML file loader loads info from the input file, and makes packages (and recipes and other stuff) through the context, which get recorded in the database
2. Then a simulation gets initialized, and info including packages are loaded in, this time from the database. However, the way that I previously set up packages, this resulted in packages getting re-created from scratch, which results in new package ids
3. Now the packages have been read multiple times and confusion ensues

I'm going to fix this by allowing, but not requiring, packages to be created with a `package_id`. I know we chose not to do this before, but now there _is_ a reason to create a package with a known package id. This will actually bring packages more in line with recipes, which in SimInit actually create a dummy composition with a known QualId (which was read in from the database, where it was created during xml file loading). 

But instead of creating dummy packages, I think I'm just going to create the package again? If this is controversial, happy to consider another option, but I don't think we need dummy packages in the same way that dummy compositions are useful.

New workflow
1. `XMLFileLoader` loads info from file. Adds unpackaged to database, then creates and adds packages, created with ids
2. `SimInit` loads info from database. Creates packages from database info, using the same package id's that were previously created
3. No extra packages are created, packages can be found in the context as expected for use by matl sell policy (or whatever else)

gonuke:
Actually... how hard would it be to remove the concept of `package_id` and just use package names?

nuclearkatie:
Probably not that hard

gonuke (approved):
This looks good to me.  Now that we've changed to package name, I noticed that it's still a little different from how we use recipes: we store the `Composition::Ptr` for each recipe, rather than the name.  I'll make an issue to review this for packages too.

PR #1755, opened 2024-05-16
Deprecate `ResourceBuff`

### LINKED ISSUE #1754: `ResourceBuff` is deprecated
I noticed a warning in https://fuelcycle.org/arche/toolkit.html#resourcebuff that says `ResourceBuff` will soon be deprecated in favor of `ResBuf`.  The git blame shows this line was added ~10 years ago... shall we remove the `ResourceBuff` implementation from the code and replace instances of it with `ResBuf`, or should it be left in for backwards compatibility?

bennibbelink:
Closes #1754. 

- [x] Change workflow cycamore reference back to `cyclus/cycamore:main`


bennibbelink:
Currently references the branch in this cycamore PR https://github.com/cyclus/cycamore/pull/604 to show that this will not break downstream.  Need to correct the ref in the workflow before merging

bennibbelink:
Pretty sizeable drops in lines covered in some of the cython generated files, but i guess given the size of those files the % coverage drop isn't very big

cyclus/pyagents.py
```diff
@@ -42,7 +42,7 @@ class Sink(Facility):
         tooltip="sink capacity",
         default=100.0,
         )
-    inventory = ts.ResourceBuffInv(capacity='max_inv_size')
+    inventory = ts.ResBufProduct()
```
gonuke, line 45: Maybe this should be
```suggestion
    inventory = ts.ResBufMaterial()
```
assuming such a thing exists

tests/sim_init_tests.cc
```diff
@@ -62,23 +62,23 @@ class Inver : public cy::Facility {
...
+    buf2.Push(invs["buf2"]);
     return invs;
   }
   virtual void Tick() { context()->Snapshot(); }
   virtual void Tock() {};
 
-  cy::toolkit::ResourceBuff buf1;
-  cy::toolkit::ResourceBuff buf2;
+  cy::toolkit::ResBuf<cy::Resource> buf1;
+  cy::toolkit::ResBuf<cy::Resource> buf2;
```
gonuke, line 81: Looking at the changes below, it think these should be `Material` buffers
```suggestion
  cy::toolkit::ResBuf<cy::Material> buf1;
  cy::toolkit::ResBuf<cy::Material> buf2;
```

tests/sim_init_tests.cc
```diff
@@ -395,19 +395,19 @@ TEST_F(SimInitTest, InitAgentInventories) {
     EXPECT_EQ(2, init_agent->buf2.count());
 
     // check agents' buf1 inventory
-    cy::Material::Ptr mat1 = agent->buf1.Pop<cy::Material>();
-    cy::Material::Ptr init_mat1 = init_agent->buf1.Pop<cy::Material>();
+    cy::Resource::Ptr mat1 = agent->buf1.Pop();
+    cy::Resource::Ptr init_mat1 = init_agent->buf1.Pop();
```
gonuke, line 399: May be unnecessary if these are defined as `Material` buffers above

tests/sim_init_tests.cc
```diff
@@ -395,19 +395,19 @@ TEST_F(SimInitTest, InitAgentInventories) {
...
 
     // check agents' buf2 inventories
-    mat1 = agent->buf2.Pop<cy::Material>();
-    init_mat1 = init_agent->buf2.Pop<cy::Material>();
-    cy::Material::Ptr mat2 = agent->buf2.Pop<cy::Material>();
-    cy::Material::Ptr init_mat2 = init_agent->buf2.Pop<cy::Material>();
+    mat1 = agent->buf2.Pop();
+    init_mat1 = init_agent->buf2.Pop();
+    cy::Resource::Ptr mat2 = agent->buf2.Pop();
+    cy::Resource::Ptr init_mat2 = init_agent->buf2.Pop();
```
gonuke, line 410: Ibid

gonuke (changes requested):
This looks pretty good - I have a few questions about how some of the replacements are templated

cyclus/pyagents.py
bennibbelink: Is a `Sink` only intended to be used with `Material`s?

tests/sim_init_tests.cc
bennibbelink: Made the switch back to `ResBuf<Material>`.  Required that we use `PopNRes()` at lines 70/71 since `cyclus::Inventories` expects keys of type `cyclus::toolkit::ResVec`

gonuke (commented):
LGTM - thanks @bennibbelink

gonuke (approved):

PR #1761, opened 2024-06-06
Prevent extra unecessary records in resources database due to packaging

### LINKED ISSUE #1760: Resources of zero quantity get created in packaging process
In #1729, re-packaging a resource was developed with the following strategy:
- Get the fill mass for packages
- If fill mass is not zero/below packaging minimum, 
  - extract single package fill quantity from resource
  - change package on new resource
  - Add new packaged resource to vector of packaged resources
  - Repeat until not enough material is left for another package

This will not result in newly created resources of zero quantity, however it can result in the old resource being left with zero quantity left, which is recorded by the res tracker on the Resources table. This effect only showed up when #1749 was merged, as this was the first time that re-packaging was actually implemented, in the material sell policy in the toolkit..

Not sure if this is technically a bug, but can and does result in resources of size zero being recorded in the output database. In tests, which often check a specific line in the database for an expected quantity, there are now new lines of size zero "clogging" up the database, which has caused [cycamore#611](https://github.com/cyclus/cycamore/issues/611#issuecomment-2146384080).

nuclearkatie:
Closes #1760 
Solves root cause issue of [cycamore#611](https://github.com/cyclus/cycamore/issues/611)

Two changes arose out of #1749
1. When materials were fully packaged, the now-zero-quantity old material was recorded as a new resource in the table, even though it was fully absorbed into the new packaged material. This created unexpected zero quantity lines in the Resources database
2. When materials were effectively not packaged (i.e. unpackaged, or going through packaging with the new packaging type equal to the old type), they were still being "packaged" in the sense that they became new resources, with new resource ids and records in the Resources database. The material sell policy would package all trades, meaning that even when packaging wasn't being used, resources were getting bumped with new state ids every time they exited an agent using the policy

This PR solves both of those issues

### DIFF
--- a/src/res_tracker.cc
+++ b/src/res_tracker.cc
@@ -45,15 +45,17 @@ void ResTracker::Extract(ResTracker* removed) {
     return;
   }
 
-  parent1_ = res_->state_id();
-  parent2_ = 0;
+  if (res_->quantity() > eps_rsrc()) {
+    parent1_ = res_->state_id();
+    parent2_ = 0;
+    
+    Record();
+  }
+  
   removed->parent1_ = res_->state_id();
   removed->parent2_ = 0;
   removed->tracked_ = tracked_;
 
-  if (res_->quantity() > eps_rsrc()) {
-    Record();
-  }
   removed->Record();
 }
 
--- a/src/res_tracker.cc
+++ b/src/res_tracker.cc
@@ -1,6 +1,7 @@
 #include "res_tracker.h"
 
 #include "recorder.h"
+#include "cyc_limits.h"
 
 namespace cyclus {
 
@@ -50,7 +51,9 @@ void ResTracker::Extract(ResTracker* removed) {
   removed->parent2_ = 0;
   removed->tracked_ = tracked_;
 
-  Record();
+  if (res_->quantity() > eps_rsrc()) {
+    Record();
+  }
   removed->Record();
 }
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,7 +20,7 @@ Since last release
 * Removed the ResourceBuff class and replaced its instances with ResBuf (#1755)
 
 **Fixed:**
-
+* Removed unnecessary records being added to the Resource database by packaging process (#1761)
 
 v1.6.0
 ====================
--- a/src/toolkit/matl_sell_policy.cc
+++ b/src/toolkit/matl_sell_policy.cc
@@ -212,15 +212,26 @@ void MatlSellPolicy::GetMatlTrades(
     double qty = it->amt;
     LGH(INFO3) << " sending " << qty << " kg of " << it->request->commodity();
     Material::Ptr mat = buf_->Pop(qty, cyclus::eps_rsrc());
-    std::vector<Material::Ptr> mat_pkgd = mat->Package<Material>(package_);
-    // push any extra material that couldn't be packaged back onto buffer
-    buf_->Push(mat);
-    if (mat_pkgd.size() > 0) {
-      if (ignore_comp_) {
-        mat_pkgd[0]->Transmute(it->request->target()->comp());
-        }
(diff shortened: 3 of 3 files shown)


nuclearkatie:
Not sure why the condas are failing, I re-ran them a few times

gonuke:
Oops... I was overzealous here in merging this 🫠

src/res_tracker.cc
```diff
@@ -44,13 +45,17 @@ void ResTracker::Extract(ResTracker* removed) {
...
   }
 
-  parent1_ = res_->state_id();
-  parent2_ = 0;
+  if (res_->quantity() > eps_rsrc()) {
+    parent1_ = res_->state_id();
+    parent2_ = 0;
+    
+    Record();
+  }
```
gonuke, line 53: @nuclearkatie Now that we're digging into when we Record things, and my understanding is expanding... I'm a little concerned about this check for size > `eps_rsrc()`.  Don't we want to record that this resource is depleted?  It should certainly have it's state ID bumped and be in a different state now than it was before.  Is that happening reliably somwhere else?

nuclearkatie: No, this is not happening elsewhere, it is not typically recorded when resources cease to have any quantity. 
From what I understand, outside of packaging, Absorb is the main way that happens. 

Absorb does not bump the absorbed resource, as a second new, zero resource. It only creates one new resource with parent1 resource and parent2 absorbed resource. I personally do not think it would make sense for the logic to flow:
```Resource + Asborbed => NewBigResource with two parents + ZeroResource with one parent Absorbed```
In part because it hasn't worked like that before. 

Extract acts similarly in my mind. If we have fully extracted an old resource into one new resource, then the table should reflect one new Resource, i.e.
```Resource => NewResource with parent Resource and the same mass.```

As far as I can tell, Cyclus does not already record zero resources anywhere else, so if we did here we would be changing the overall logic of the Resource table. which I'm trying to avoid doing because I don't think doing so is on my critical path

gonuke: Wasn't this happening before #1760 when all of the quantity was extracted from an initial material object?

nuclearkatie: I guess yes technically but I don't think any archetype was actually extracting a full resource (there's really no reason to, if a resource is the right size you would just pop it and use as is)

gonuke: I guess a separate and related question is whether depleted resources formally "expire" or just linger...  that may depend on the archetypes they experience, I guess.

gonuke: > I guess yes technically but I don't think any archetype was actually extracting a full resource (there's really no reason to, if a resource is the right size you would just pop it and use as is)

Not all resources are in buffers.  I could imagine an algorithm that calls `ExtractRes()` on one resources and happens to extract all of the first one. i.e. it doesn't bother checking to see whether they're the same size.

gonuke: > As far as I can tell, Cyclus does not already record zero resources anywhere else, so if we did here we would be changing the overall logic of the Resource table. which I'm trying to avoid doing because I don't think doing so is on my critical path

I agree that this is not in scope for you.  I think this all actually depends on many aspects of how a resource is managed in and out of ResBufs.  In many cases, they may ultimately just be merged with other resources through arbitrary operations that are properly tracked by the tracker.

I'm worried about an edge case where some algorithm chooses to neglect a resource whose quantity is too small, and we effectively lose track of the resource.

gonuke: Let's say that Material M is at the front of a ResBuf and some packaging operation results in it having `0 < M.quantity < eps_rsrc()`.  It will still be at the front of the ResBuf.  When the next pacaking request comes, we'll pop M and things behind it, and merge them all into a single Material M'.  That new material will have M as one of it's parents, but in the state it was in before the first round of packaging happened.

PR #1763, opened 2024-06-07
Use miniforge instead of miniconda

### LINKED ISSUE #1762: Cannot find libexpat.so.1
Conda CI fails with: 

```
ImportError: libexpat.so.1: cannot open shared object file: No such file or directory
#20 36.26 CMake Error at cmake/UseCyclus.cmake:235 (message):
#20 36.26   /cyclus/cli/cycpp.py failed on '/cyclus/agents/null_inst.h' with exit code
#20 36.26   '1'
```

bennibbelink:
Closes #1762.

Looks like a forced reinstall of libexpat resolves the issue.  This PR also changes the build to use conda-forge::conda which should be more up to date and uses libmamba as the solver by default (can confirm via `conda info`)

One other change - we were using the compilers intended for CentOS.  Now we reference the generic compilers instead.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Since last release
 * Remove ``pandas`` as build dependency (#1748)
 * Consistently use hyphens in ``install.py`` flags (#1748)
 * Material sell policy can package materials (#1749)
+* Use miniforge for conda CI builds instead of miniconda (#1763)
 
 **Removed:**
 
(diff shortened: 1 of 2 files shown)


bennibbelink:
Was experiencing a compatibility issue with glibc.  In an attempt to avoid incompatibilities this now installs miniforge instead of miniconda - I figured since we already rely heavily on conda-forge this shouldn't be an issue (and will help keep things more up to date).  This no longer requires that we `--force-reinstall` libexpat and libsqlite  

The failing downstream tests (`StorageTest.PackageSplitFirst`) should be unrelated to this PR

bennibbelink:
Hmm... not sure why the downstream test failures are not happening deterministically.

gonuke (approved):
Thanks @bennibbelink - this all looks fine to me

PR #1769, opened 2024-07-08
Add CI for Ubuntu 24.04

bennibbelink:
Closes #1767 and #1764.

PR #1770, opened 2024-07-08
Add Ubuntu 24.04 to CI (Replacement)

### LINKED ISSUE #1767: XInclude on Ubuntu 24.04
`test_include_recipe.py` fails on Ubuntu 24.04.  I tracked the bug down to `XMLParser::Document()`, but haven't nailed it completely down yet.  I think we might need to change our solution for removing the `xml:base` element, but have not found a fix yet.

bennibbelink:
Replaces #1769.  Closes #1769, closes #1767.


bennibbelink:
I had modified the `code_coverage.yml` workflow to use 24.04 as well, but that upgrade turned out to be nontrivial.  I vote we keep it at 22.04 for now, and I can make an issue to upgrade to 24.04 as a separate PR.

docker/Dockerfile
```diff
@@ -52,7 +52,7 @@ RUN echo 'export PATH=/opt/conda/bin:$PATH' > /etc/profile.d/conda.sh && \
     /bin/bash ~/miniforge.sh -b -p /opt/conda && \
     rm ~/miniforge.sh
 
-ENV PATH /opt/conda/bin:$PATH
+ENV PATH=/opt/conda/bin:$PATH
```
gonuke, line 55: Wow - did they change the spec or are we just lucky that this has always worked before?

src/xml_parser.cc
```diff
@@ -63,14 +63,21 @@ xmlpp::Document* XMLParser::Document() {
   // This adds the capability to have nice include semantics
   bool generate_xinclude_nodes = true;
   bool fixup_base_uris = false;
-  #if LIBXMLXX_MAJOR_VERSION == 2 
-  doc->process_xinclude(generate_xinclude_nodes);
+  #if LIBXMLXX_MAJOR_VERSION < 2
```
gonuke, line 66: When does this happen?  Or is it just for completeness?

gonuke (approved):
I think this looks OK, but do have a couple of curiosities....

Thanks @bennibbelink

docker/Dockerfile
bennibbelink: They changed the spec.  I discovered this from "legacy" warnings being thrown by `gcov` when I was messing around in 24.04

src/xml_parser.cc
bennibbelink: On 24.04 `test_include_recipe.py` would fail.  You might remember that when libxml++ processes XIncludes it generates a `base` attribute (I saw you made some commits to this last summer/fall) which renders the xml invalid according to our RelaxNG schema.  There is logic in place to parse through and remove this attribute, but after v2.42.0 there is an argument to `process_xinclude()` to avoid generating the `base` attribute altogether.  See the [changelog with that release](https://github.com/libxmlplusplus/libxmlplusplus/releases/tag/2.42.0).

On ubuntu 22.04 the version of libxml++ is 2.40, and on our conda builds we use libxml++ 4.0

gonuke: Sorry - my question wasn't clear - when do we ever get libxml++ version < 2?

bennibbelink: Oh sorry, tracking now. It'll never be <2. The version states we need to check for are <2.42 and >2.42, if I'm not mistaken. I will clean up that logic

bennibbelink: Should be good now.

PR #1771, opened 2024-07-10
warning and limit for splitting resources into too many packages

### LINKED ISSUE #1766: Attempting to package a large quantity into small packages can throw vector length error
I set up a toy simulation to play with packaging, where I forgot to set a throughput. So, the source traded 1e299 kg to the storage immediately. This allowed a storage agent with packaging to try to package 1e299 kg of material into packages with `fill_min` of 1 and `fill_max` of 2. Obviously, this broke the simulation because storage tried to create a bid vector with 5e298 bids  for 2 units each.

Could be solved by setting a flag in the material sell policy to check that the number of packages that could be created is less than the relevant numeric limit and throwing a more descriptive error if not.

However, should we be even more restrictive here? We *probably* never want to create billions of resources by way of packaging. But, where's the line?

Error
`terminate called after throwing an instance of 'std::length_error'`
`  what():  cannot create std::vector larger than max_size()`

nuclearkatie:
Adds a warning value (1e5) and a limit value (unsigned int / 10) for re-packaging. Implemented in Material Sell Policy, when calculating how many bids to respond with, and in the repackaging process itself (in Resource). If trying (or in the case of sell pol, intending) to repackage a single resource into more than warn/limit values, raise warning/limit

Closes #1766

### DIFF
--- a/src/resource.h
+++ b/src/resource.h
@@ -5,8 +5,6 @@
 #include <vector>
 #include <boost/shared_ptr.hpp>
 
-#include "error.h"
-#include "logger.h"
 #include "package.h"
 
 class SimInitTest;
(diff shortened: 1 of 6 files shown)


nuclearkatie:
aha, this has raised an additional issue that trying to convert a float to int (for the number of bids to respond with) can have overflow issues even before we deal with the limits on vector length.

I think this gives us an additional point at which we need to throw an error, I don't think we should be proceeding at all if overflow errors are raised

src/resource.h
```diff
@@ -138,6 +139,18 @@ std::vector<typename T::Ptr> Resource::Package(Package::Ptr pkg) {
     return ts_pkgd;
   }
 
+  int approx_num_pkgs = quantity() / fill_mass;
+  // Check if the number of packages is within the limits, including if
+  // int overflow is reached
+  if (approx_num_pkgs > Package::SplitLimit() || approx_num_pkgs == std::numeric_limits<int>::min()) {
```
gonuke, line 145: I wonder about making a function `Package::ExceedsSplitLimit(approx_num_pkgs)` that can handle these two variations of ways to exceed the limit - and any other we find?

gonuke (changes requested):
This looks good to me, but it seems like the conditions could become more complicated so I wonder if we make it a function?

gonuke (approved):
This looks good to me - thanks @nuclearkatie

PR #1774, opened 2024-07-16
rounding can cause fmod errors

### LINKED ISSUE #1773: Use of fmod in material sell policy packaging process causes can result in extra full bid
Rounding in the size of bid quantities can result in errors when determining the remainder, i.e. extra material that wouldn't be packaged into the fixed bid quantity. Determining this value is important because sometimes this remainder, while lower than the regular bid quantity, may still be above the minimum and thus can be added as an additional, smaller bid.

Specifically, rounding in the `bid_qty` value can result in an fmod value (`fmod(qty / bid_qty`)) that is equal to the bid quantity, meaning that an extra bid of the same size as all other bids would be added to the response to request for bids. In this case, more material than is available might accidentally be offered

nuclearkatie:
Closes #1773

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ Since last release
 * Rely on ``python3`` in environment instead of ``python`` (#1747)
 * Remove ``pandas`` as build dependency (#1748)
 * Consistently use hyphens in ``install.py`` flags (#1748)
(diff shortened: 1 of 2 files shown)


gonuke (approved):
Seems reasonable.... thanks @nuclearkatie

PR #1778, opened 2024-07-24
use keep_packaging in resbuf

nuclearkatie:
Replace the `unpackaged` parameter in ResBuf with `keep_packaging`, switching the default from true to false to keep the logic the same.

This reflects the changes that are being made in Reactor and Sink to keep the packaging by default ([cycamore#618](https://github.com/cyclus/cycamore/pull/618))

### DIFF
--- a/src/toolkit/res_buf.h
+++ b/src/toolkit/res_buf.h
@@ -61,7 +61,11 @@ typedef std::vector<Product::Ptr> ProdVec;
 template <class T>
 class ResBuf {
  public:
-  ResBuf(bool is_bulk=false, bool keep_packaging=false) : cap_(INFINITY), qty_(0), is_bulk_(is_bulk), keep_packaging_(keep_packaging) { }
+  ResBuf(bool is_bulk=false, bool keep_packaging=false) : cap_(INFINITY), qty_(0), is_bulk_(is_bulk), keep_packaging_(keep_packaging) {
+    if (is_bulk_ && keep_packaging_) {
+      throw ValueError("bulk storage resbufs cannot keep packaging. Only one of the two options can be true.");
+    }
+   }
 
   virtual ~ResBuf() {}
 
--- a/src/toolkit/res_buf.h
+++ b/src/toolkit/res_buf.h
@@ -61,7 +61,7 @@ typedef std::vector<Product::Ptr> ProdVec;
 template <class T>
 class ResBuf {
  public:
-  ResBuf(bool is_bulk=false, bool unpackaged=true) : cap_(INFINITY), qty_(0), is_bulk_(is_bulk), unpackaged_(unpackaged) { }
+  ResBuf(bool is_bulk=false, bool keep_packaging=false) : cap_(INFINITY), qty_(0), is_bulk_(is_bulk), keep_packaging_(keep_packaging) { }
 
   virtual ~ResBuf() {}
 
@@ -84,6 +84,13 @@ class ResBuf {
     }
     cap_ = cap;
   }
+
+  /// Sets whether the buffer should keep packaged resources
+  void keep_packaging(bool keep_packaging) {
+    keep_packaging_ = keep_packaging;
+  }
+
+  bool keep_packaging() const { return keep_packaging_; }
 
   /// Returns the total number of constituent resource objects
   /// in the buffer. Never throws.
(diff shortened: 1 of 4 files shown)


gonuke:
Should `keep_packaging` override `is_bulk`? I suppose this is not a new question as it was relevant before as well, but it kind of jumps out at me now that you can't have both a `keep_packaging` and an `is_bulk` ResBuf

nuclearkatie:
Hm, good question. I think no? My feeling is that bulk should override packaging...

nuclearkatie:
Maybe there should be an error if you try to make both true?

gonuke:
> Hm, good question. I think no? My feeling is that bulk should override packaging...

How do we want to handle this, then?  Don't allow `keep_packaging_` if `is_bulk_` is set? and throw exceptions? issue warning?

src/toolkit/res_buf.h
```diff
@@ -85,6 +89,13 @@ class ResBuf {
     cap_ = cap;
   }
 
+  /// Sets whether the buffer should keep packaged resources
+  void keep_packaging(bool keep_packaging) {
+    keep_packaging_ = keep_packaging;
```
gonuke, line 94: I think we need to test here, too.  We could put the test only here, and then use this method to set this variable in the constructor.

gonuke (changes requested):
One more place to keep an eye on consistency with is_bulk

src/toolkit/res_buf.h
nuclearkatie: we don't use the `capacity` method to set the capacity variable though

nuclearkatie: I did add a test though

gonuke: Maybe we should??

gonuke: Is it every possible at initialization to have an inconsistency in capacity?

nuclearkatie: Currently no, though it probably shouldn't allow a zero or negative capacity

gonuke: I realize that when I said we should "test", here that may not have been clear.  Whether we change the constructor or not, we need to make sure that if we change `keep_packaging_` to a value that is incompatible with `is_bulk_` in this function that we catch it.

nuclearkatie: Ah, yes yes

src/toolkit/res_buf.h
```diff
@@ -61,7 +61,13 @@ typedef std::vector<Product::Ptr> ProdVec;
 template <class T>
 class ResBuf {
  public:
-  ResBuf(bool is_bulk=false, bool unpackaged=true) : cap_(INFINITY), qty_(0), is_bulk_(is_bulk), unpackaged_(unpackaged) { }
+  ResBuf(bool is_bulk=false, bool keep_pkg=false) : qty_(0), is_bulk_(is_bulk) {
+    capacity(INFINITY);
+    keep_packaging(keep_pkg);
+    if (is_bulk_ && keep_packaging_) {
+      throw ValueError("bulk storage resbufs cannot keep packaging. Only one of the two options can be true.");
+    }
```
gonuke, line 69: I think we can/should move this to `keep_packaging()` now

src/toolkit/res_buf.h
```diff
@@ -85,6 +95,13 @@ class ResBuf {
     cap_ = cap;
   }
 
+  /// Sets whether the buffer should keep packaged resources
+  void keep_packaging(bool keep_packaging) {
+    keep_packaging_ = keep_packaging;
+  }
```
gonuke, line 101: After the discussion of making sure this is consistent with `is_bulk_`, I realized that we don't let anyone change `is_bulk_` after initialization, which probably makes a lot of sense.

Should we also not allow `keep_packing_` to change after initialization?  (Thus undoing everything we discussed yesterday)

gonuke (changes requested):
Let me know if you want to chat about this...

src/toolkit/res_buf.h
nuclearkatie: No, I feel like it's fine to do so

gonuke (approved):

PR #1781, opened 2024-08-04
Multi-phase initialization for cython modules

### LINKED ISSUE #1779: Cython API is not implemented correctly
We still rely on importing our three cython modules in `__init__.py` to fix the segfaults at the end of `cyclus.main()`.  Now in macos builds I am experiencing the same segfaults, even with those imports.  A proper/correct fix to the python interpreter management should be found.

bennibbelink:
Closes #1779.

Best practice for calling cython methods from within C involves appending the modules to an initialization table prior to the Python interpreter being initialized.  The issue that we run into with our cython modules is that we run cyclus with two origins: 1 - directly from the C++ driver, and 2 - we invoke the C++ driver via Python wrappers. 

### Calling cyclus via C++ driver
We have had structure in place to correctly append to the initialization table, initialize the interpreter, and import the modules.  This is why we haven't had any issues with cyclus when used this way or with `cyclus_unit_tests`

### Calling cyclus via Python wrappers
Since the python interpreter is already started at the time we want to import/call our cython methods, we cannot follow the [usual inittab flow](https://cython.readthedocs.io/en/stable/src/userguide/external_C_code.html#inittab-guide).  [Others](https://groups.google.com/g/cython-users/c/CtcxYXHsTbs) have had similar circumstances, and there seem to be two potential workarounds.  One is disabling multiphase initialization entirely (this is not preferred according to the cython docs).  The other is [manually run the multi-phase initialization](https://cython-guidelines.readthedocs.io/en/latest/troubleshooting/embedding.html#working-with-multi-phase) ourselves.  I have mimicked the example from the cython guidelines and created some generic helper methods to make it easy on us to call cython methods should our modules change in the future.  I hope that it is intuitive enough to not cause maintainability issues down the road.  Our hacky fix to this before was importing the three modules in `cyclus/__init__.py`, those imports have been removed.

--------------------------------------------------

One thing that I am not completely satisfied with - we have two working solutions for two different circumstances, but they are not isolated at all.  When running cyclus via C++, the inittab mechanism works great, but it still calls the cython methods via CallCython<Non>Void().  Similarly, when running cyclus via Python the CallCython<Non>Void() works well but we still attempt to perform the inittab mechanism which seems to fail silently.  Still some work to be done here.

Another oddity about this PR... CI passes on all workflows except the 20.04_conda image **unless** I comment out the docker cache-from line in the workflow.  CI was failing and I was not able to replicate it locally so I tried commenting out the cache in CI and what do you know :man_shrugging: I'm not sure what to make of this, need to diagnose.


bennibbelink:
Closing in favor of #1809

PR #1782, opened 2024-08-10
Add doxygen to dependency list in dockerfile

bennibbelink:
In trying to build the api docs for cyclus.github.com I was having to rebuild cyclus and cycamore after installing doxygen.  I think it's easier if we just include it in our published images from the start so that the `cyclusdoc` and `cycamoredoc` make targets exist for when we need them downstream.


gonuke (approved):
Thanks @bennibbelink

gonuke:
Looks like this needs a rebase on the `CHANGELOG.rst`

gonuke (approved):
Thanks @bennibbelink

PR #1784, opened 2024-08-14
transportunits in siminit

### LINKED ISSUE #1783: Transport units are loaded but then lost before agents are created
Transport units get loaded from the input file, but then somehow the list of transport units gets cleared before agents get created. This results in a KeyError when trying to create agents whose prototypes use transport units. Need to figure out where/how the list of all transport units gets cleared and prevent that from happening

```
terminate called after throwing an instance of 'cyclus::KeyError'  
what():  No user-created transport units exist
```

nuclearkatie:
Closes #1783. Oops I repeated the same mistake in #1750 as I did in #1749 (not putting packaging in SimInit), which I fix here.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ Since last release
 
 **Added:**
 
-* Added TransportUnits (#1750, #1772)
+* Added TransportUnits (#1750, #1772, #1784)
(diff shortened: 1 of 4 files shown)


gonuke (approved):
Looks good

PR #1787, opened 2024-08-18
Make Doxygen docs prettier

bennibbelink:
Uses https://jothepro.github.io/doxygen-awesome-css/ to style the api docs.  Also adds a short blurb on the mainpage.  These docs will be built and published by the action in cyclus.github.com.  For reference of what the new styled docs look like I've attached a zip, or feel free to checkout this PR locally, build cyclus (with doxygen installed) and `make cyclusdoc` from the build folder (docs will be generated at `build/doc/html`)

[html.tar.gz](https://github.com/user-attachments/files/16649590/html.tar.gz)


gonuke:
The failure here is also showing up in the conda-forge feedstock... ???

bennibbelink:
It also appears to be failing because of `dirty_quantity_` not being mutable.  I had fixed that in #1782 but I'll resolve these build issues in a new separate branch to be merged first

bennibbelink:
It appears that Boost has updated their Sha1 class in v1.86.0.  Now it uses a `char[20]` as the digest type instead of `unsigned int[5]`

[v1.85.0](https://www.boost.org/doc/libs/1_85_0/boost/uuid/detail/sha1.hpp) vs [1.86.0](https://www.boost.org/doc/libs/1_86_0/boost/uuid/detail/sha1.hpp)

One option is to just modify our Digest class to be compatible with new versions of boost, but based on this comment from scopatz we might be able to change how we handle this entirely.  We already enforce C++17 in our build.  I will try to rework this to just use a std::array

```
/// The digest type for SHA1s.
///
/// This class is a hack around a language deficiency in C++. You cannot pass
/// around an array (unsinged int[5]) between function calls. You can only
/// pass pointers, which would involve lost of new/free and heap shenanigans
/// that are not needed for a dumb container. Therefore Sha1::Digest() cannot
/// return what would be most natural. The second most natural thing would be
/// a std::array<unsigned int, 5>. However, std::array is a C++11 feature and
/// we are not yet ready to go down that road.
///
/// To pass an array into and out of a function it has to be inside of struct
/// or a class. I chose a class here since there are many member functions.
///
/// The reason why this is public is that it needs to be directly writable
/// from buffers coming from HDF5. In the future, this really should just be
/// a std::array.
```

doc/CMakeLists.txt
```diff
@@ -17,6 +17,24 @@ IF(DOXYGEN_FOUND)
...
 
+    # -----------------------------------------------------------------------------
+    # Fetch doxygen stylesheet from https://github.com/jothepro/doxygen-awesome-css
+    # -----------------------------------------------------------------------------
+    SET(AWESOME_CSS_URL "https://raw.githubusercontent.com/jothepro/doxygen-awesome-css/main/doxygen-awesome.css")
+    SET(AWESOME_CSS_DESTINATION "${DOC_INPUT_DIR}/doxygen-awesome.css")
+    FILE(DOWNLOAD ${AWESOME_CSS_URL} ${AWESOME_CSS_DESTINATION}
+        STATUS DOWNLOAD_STATUS
+        LOG DOWNLOAD_LOG
+    )
```
gonuke, line 28: I wonder if we should just copy a static version of this into Cyclus?  On one hand, it could become stale, but on the other hand, it wouldn't update with no notice and suddenly change our docs.

I could go either way, but just looking for a brief discussion....

doc/mainpage.dox
gonuke, line 1: I'm guessing this `mainpage` is replaced by the entry in `cyclus.h` below... but this one pointed to the developer guide (possibly with an out of date URL) and that one does not.  Should we do that somewhere?

gonuke (changes requested):
A couple of little queries - thanks for your work on this @bennibbelink

doc/CMakeLists.txt
bennibbelink: I think thats a good strategy.  Fetching it adds complexity that probably isn't necessary for something like css.  I will update this

doc/mainpage.dox
bennibbelink: Added links to the kernel and archetype developer guides

doc/logo.png
gonuke, line 1: thoughts on using the animated GIF logo from cyclus.github.com - is that supported?

gonuke (changes requested):
Thanks @bennibbelink - one question on logo file choice... :)

doc/logo.png
bennibbelink: GREAT idea.  Tried it out and it works

gonuke (approved):
Thanks for indulging me on the GIF @bennibbelink

PR #1789, opened 2024-08-28
Update stubs for newer build infrastructure

### LINKED ISSUE #1788: Update cycstub to match new build infrastructure
It appears that the `CMakeList.txt` file and perhaps other parts of cycstub have not been updated to match the newest build infrastructure.

gonuke:
Minor changes to CMake config and install script to match newest cyclus build.

Fixes #1788 

Should probably devise a test for this?


stubs/CMakeLists.txt
```diff
@@ -14,20 +14,20 @@ ENDIF()
 
 PROJECT(@lib_name@)
 
-# check for and enable c++11 support
+# check for and enable c++17 support
 INCLUDE(CheckCXXCompilerFlag)
-CHECK_CXX_COMPILER_FLAG("-std=c++11" COMPILER_SUPPORTS_CXX11)
+CHECK_CXX_COMPILER_FLAG("-std=c++17" COMPILER_SUPPORTS_CXX11)
```
bennibbelink, line 19:
```suggestion
CHECK_CXX_COMPILER_FLAG("-std=c++17" COMPILER_SUPPORTS_CXX17)
```

stubs/CMakeLists.txt
```diff
@@ -14,20 +14,20 @@ ENDIF()
...
 IF(COMPILER_SUPPORTS_CXX11)
```
bennibbelink, line 20:
```suggestion
IF(COMPILER_SUPPORTS_CXX17)
```

bennibbelink (changes requested):
What kind of test? Sorry I'm not sure I follow

gonuke:
> What kind of test? Sorry I'm not sure I follow

I'll confess that I'm not sure what the test would be either.  It would start by running cycstub to generate a stub archetype and then some combination of:
* confirming that it's `CMakeLists.txt` file is compatible with the main cyclus one (somehow???(
* building the stub and ensuring it loads with cyclus

bennibbelink:
There is a lot of redundancy between `stubs/CMakeLists.txt` and root `CMakeLists.txt`.  What if we write a module in `cmake/` to hold most of the boilerplate and call it `SetupCyclusBuild.cmake` or `CyclusBuildPreamble.cmake` (or something along those lines) so that the same CMake code can be used in the `CmakeLists.txt` and `stubs/CmakeLists.txt`.  The modules in `cmake/` are installed as part of the build already, so we should be able to do the same thing in cycamore too and reduce redundancy across repos.  I can try to implement this over the long weekend.

As for the tests, it looks like `tests/test_stubs.py` does what you're describing (?).  Do we need to expand on this?

bennibbelink:
> What if we write a module in `cmake/`

https://github.com/bennibbelink/cyclus/tree/cmake-boilerplate this is sort of a minimum viable product of what I mean.  We could choose to wrap more of it in macros if we want (e.g. finding and linking against packages) but I think there is probably a balance to find - verbosity is helpful to convey what is actually being done but abstracting the code out to macros should reduce redundancy across builds and help maintainability.  Curious to hear your thoughts @gonuke.

gonuke:
Is this partly what the `UseCyclus` macro is meant to do? I definitely like this idea!

bennibbelink:
Yes, but the macros in `UseCyclus.cmake` are targeted towards building agent libraries.  We can add build setup macros to that file or a different one, doesn't really matter.  I'd probably lean towards having another file to separate concerns but if you think we should keep them all in one place thats no problem

gonuke:
Another file is fine - thanks

bennibbelink:
Closing this PR in favor of #1793.

Issue #1786, opened 2024-08-16
MatlSellPolicy + packaging cause errors when more than one agents wants commod


nuclearkatie:
When more than one request is made for the same outcommod, the material sell policy makes multiple bids on the same material. For example 3 kg are ready and can be packaged into 1 kg packages. If two agents each request 2 kg, the material sell policy will make the following bids:
- 1 kg and 1 kg (for the first agent), 
- 1 kg and 1 kg (for the second). 

For a total of 4 kg of bids. The DRE sorts this out so that only 3 kg of trades are actually completed. This generally means that the first agent will get 2 kg (1+1) and the second agent will get 1 kg

Now, imagine that 2.5 kg are ready and must still be packaged into 1 kg each. The material sell policy will do the same thing, make the same bids. The DRE will solve and ensure that only 2.5 kg of trades are created. Now the first agent gets 2 (1+1) and the second agent gets 0.5. Uh oh, that can't be packaged. This is a different version of what I just raised in #1785 

Now add the transport units. Lets go back to 3 kg, packages of 1. Now imagine the transport units must be 2-3 packages. When making the bids, the material sell policy notes that each of the two requesting agents wants 2 kg, which is 2 packages of 1 kg each. Great! That complies with the transport unit restrictions. Same bids are made as above.

Now again, the DRE solves for 2 kg (1+1) for the first agent, and 1 kg for the second. Now when executing the trades, the material sell policy checks the total number of trades. There are three total. Great, three packages fits in a transport unit! Except there aren't three trades to one agent, there are two trades to one agent and one trade to another. This shouldn't work, the second agent should have a "failed" trade.

How to address:
- Need to solve the "failed" trade issue (#1785) first before any progress can be made here
- Transport unit restrictions should be checked per requesting agent
- Do we need to revist the idea of whether packaging calculations should be in the DRE?

PR #1790, opened 2024-08-28
Packaging updates - failed trades send zero material and more robust GetFillMass 

### LINKED ISSUE #1785: Package-related "failed" trades cause seg faults
In #1749, we dealt with the reality that sometimes an agent with material sell policy would make a bid of material that could be packaged, and the requesting agent would accept a smaller amount of material (for example, if they were able to get their request partially filled with a higher preference material from another agent). Then, the DRE would execute the trade and suddenly the agent would be faced with a trade it couldn't make. E.g. request for 2 kg. Bid for 2 kg with packaging requirements of 2 kg. Requester ultimately accepts 1 kg. When trying to execute the trade, can't package 1 kg, so sends an empty trade back

I thought this "failed" trade could just be sent with no material.  Turns out, this actually seg faults. 

Possible solutions:
- Change how trades work so that zero trades ("failed" trades) can go through and be accepted
- Make all packaged trades (anything not using `unpackaged`) exclusive

nuclearkatie:
Addresses @gonuke's comments on [cycamore#621](https://github.com/cyclus/cycamore/pull/621)

Closes #1785 and is part of process for addressing #1786

### DIFF
--- a/src/package.cc
+++ b/src/package.cc
@@ -29,13 +29,15 @@ Package::Ptr& Package::unpackaged() {
   return unpackaged_;
 }
 
-double Package::GetFillMass(double qty) {
+std::pair<double, int> Package::GetFillMass(double qty) {
   if (qty < fill_min_) {
     // less than one pkg of material available
-    return 0;
+    return std::pair<double, int> (0, 0);
   }
 
   double fill_mass;
+  int num_at_fill_mass;
+  
   if (strategy_ == "first") {
     fill_mass = fill_max_;
   } else if (strategy_ == "equal") {
@@ -49,7 +51,13 @@ double Package::GetFillMass(double qty) {
       fill_mass = fill_max_;
     }
   }
-  return std::min(qty, fill_mass);
+  fill_mass = std::min(qty, fill_mass);
+  if (fill_mass >= fill_min_) {
+    num_at_fill_mass = static_cast<int>(std::floor(qty / fill_mass));
+    return std::pair<double, int>(fill_mass, num_at_fill_mass);
+  } else {
+    return std::pair<double, int>(0, 0);
+  }
 }
   
 Package::Package(std::string name, double fill_min, double fill_max,
--- a/src/package.cc
+++ b/src/package.cc
@@ -52,12 +52,8 @@ std::pair<double, int> Package::GetFillMass(double qty) {
     }
   }
   fill_mass = std::min(qty, fill_mass);
-  if (fill_mass >= fill_min_) {
-    num_at_fill_mass = static_cast<int>(std::floor(qty / fill_mass));
-    return std::pair<double, int>(fill_mass, num_at_fill_mass);
-  } else {
-    return std::pair<double, int>(0, 0);
-  }
+  num_at_fill_mass = static_cast<int>(std::floor(qty / fill_mass));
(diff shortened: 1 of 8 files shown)


nuclearkatie:
After several re-runs, still getting failures on two of the three conda runs (but not 20.04)... don't really know why

src/package.cc
```diff
@@ -49,7 +51,13 @@ double Package::GetFillMass(double qty) {
       fill_mass = fill_max_;
     }
   }
-  return std::min(qty, fill_mass);
+  fill_mass = std::min(qty, fill_mass);
+  if (fill_mass >= fill_min_) {
```
gonuke, line 55: I think by returning this pair, you may not even need this test anymore, at all, because you've removed the pathological case.  It used to be possible to return `GetFillMass() = 0` because of the early return near the top of the function when `qty < fill_min_`.  You now return (0,0) for that case.  The only way we can get to this point in `GetFillMass()` is if `qty >= fill_min_`.  Therefore, we can guarantee that `fill_mass >= fill_min_` at this point and that `num_at_fill_mass` is well-behaved.

src/toolkit/matl_sell_policy.cc
```diff
@@ -197,9 +199,12 @@ std::set<BidPortfolio<Material>::Ptr> MatlSellPolicy::GetMatlBids(
     for (rit = requests.begin(); rit != requests.end(); ++rit) {
       req = *rit;
       qty = std::min(req->target()->quantity(), limit);
-      bid_qty = excl ? quantize_ : package_->GetFillMass(qty);
+      std::pair<double, int> fill = package_->GetFillMass(qty);
+      bid_qty = excl ? quantize_ : fill.first;
       if (bid_qty != 0) {
-        n_full_bids = static_cast<int>(std::floor(qty / bid_qty));
+        std::cerr << "bid_qty: " << bid_qty << std::endl;
+        std::cerr << "full bids: " << fill.second << std::endl;
```
gonuke, line 206: I don't think we need these anymore

gonuke (changes requested):
I thought through the test we moved more carefully and I think this changes removes the need for the test at all.

src/toolkit/matl_sell_policy.cc
nuclearkatie: Oops

src/toolkit/matl_sell_policy.cc
```diff
@@ -280,6 +283,9 @@ void MatlSellPolicy::GetMatlTrades(
           // packaging successful
           trade_mat = mat_pkgd[0];
           shippable_pkgs-=1;
+        } else {
+          // packaging failed. Will need to ship empty trade
+          trade_mat = Material::CreateUntracked(0, mat->comp());
```
gonuke, line 288: is there a test for this case?

gonuke (commented):
Wondering about another tests for the failed trade?

gonuke:
Rebase on #1796 to pass tests

nuclearkatie:
Cycamore build fails because this PR changes `GetFillMass` from returning a double to returning a pair (double and int). [cycamore#621](https://github.com/cyclus/cycamore/pull/621) fixes that, after this is merged

gonuke (approved):

PR #1792, opened 2024-09-02
Support Boost 1.86.0

### LINKED ISSUE #1791: HDF5 backend not compatible with Boost 1.86.0
Boost Sha1 class underwent a major upgrade in version 1.86.0, including the digest type converting from `unsigned int[5]` to `unsigned char[20]`.  Our hdf5 backend infrastructure should be updated to support this change.  See https://github.com/cyclus/cyclus/pull/1787 for more info

bennibbelink:
This is still a work in progress (closes #1791).  Attempting to conditionally store Sha1 keys in HDF5 as `char[20]` or `unsigned int[5]` based on the boost version.  I think I'm on the right track, still not passing tests though, new VL_STRING columns are being created as integer types (I think, see `col0` below).

```
{
  "name": "test0",
  "path": "/test0",
  "attributes": [
    {
      "name": "CLASS",
      "shape": [],
      "type": {
        "class": "String",
        "charSet": "ASCII",
        "strPad": "null-terminated",
        "length": 6
      }
    },
    {
      "name": "FIELD_0_NAME",
      "shape": [],
      "type": {
        "class": "String",
        "charSet": "ASCII",
        "strPad": "null-terminated",
        "length": 5
      }
    },
    {
      "name": "TITLE",
      "shape": [],
      "type": {
        "class": "String",
        "charSet": "ASCII",
        "strPad": "null-terminated",
        "length": 6
      }
    },
    {
      "name": "VERSION",
      "shape": [],
      "type": {
        "class": "String",
        "charSet": "ASCII",
        "strPad": "null-terminated",
        "length": 4
      }
    },
    {
      "name": "cyclus_dbtypes",
      "shape": [
        1
      ],
      "type": {
        "class": "Integer",
        "endianness": "little-endian",
        "size": 32
      }
    },
    {
      "name": "shape0",
      "shape": [
        1
      ],
      "type": {
        "class": "Integer",
        "endianness": "little-endian",
        "size": 32
      }
    }
  ],
  "kind": "dataset",
  "shape": [
    0
  ],
  "type": {
    "class": "Compound",
    "fields": {
      "col0": {
        "class": "Array",
        "base": {
          "class": "Integer (unsigned)",
          "endianness": "little-endian",
          "size": 8
        },
        "dims": [
          20
        ]
      }
    }
  },
  "chunks": [
    1024
  ],
  "filters": [
    {
      "id": 1,
      "name": "deflate",
      "cd_values": [
        6
      ]
    }
  ],
  "rawType": {
    "signed": false,
    "type": 6,
    "compound_type": {
      "nmembers": 1,
      "members": [
        {
          "signed": false,
          "type": 10,
          "array_type": {
            "signed": false,
            "type": 0,
            "vlen": false,
            "littleEndian": true,
            "size": 1,
            "shape": [
              20
            ],
            "total_size": 20
          },
          "vlen": false,
          "littleEndian": true,
          "size": 20,
          "name": "col0",
          "offset": 0,
          "shape": []
        }
      ]
    },
    "vlen": false,
    "littleEndian": true,
    "size": 20,
    "total_size": 0
  }
}
```

compared to when built with apt (still using older version of boost):
```
{
    "name": "test0",
    "path": "/test0",
    "attributes": [
      {
        "name": "CLASS",
        "shape": [],
        "type": {
          "class": "String",
          "charSet": "ASCII",
          "strPad": "null-terminated",
          "length": 6
        }
      },
      {
        "name": "FIELD_0_NAME",
        "shape": [],
        "type": {
          "class": "String",
          "charSet": "ASCII",
          "strPad": "null-terminated",
          "length": 5
        }
      },
      {
        "name": "TITLE",
        "shape": [],
        "type": {
          "class": "String",
          "charSet": "ASCII",
          "strPad": "null-terminated",
          "length": 6
        }
      },
      {
        "name": "VERSION",
        "shape": [],
        "type": {
          "class": "String",
          "charSet": "ASCII",
          "strPad": "null-terminated",
          "length": 4
        }
      },
      {
        "name": "cyclus_dbtypes",
        "shape": [
          1
        ],
        "type": {
          "class": "Integer",
          "endianness": "little-endian",
          "size": 32
        }
      },
      {
        "name": "shape0",
        "shape": [
          1
        ],
        "type": {
          "class": "Integer",
          "endianness": "little-endian",
          "size": 32
        }
      }
    ],
    "kind": "dataset",
    "shape": [
      3
    ],
    "type": {
      "class": "Compound",
      "fields": {
        "col0": {
          "class": "Array",
          "base": {
            "class": "Integer (unsigned)",
            "endianness": "little-endian",
            "size": 32
          },
          "dims": [
            5
          ]
        }
      }
    },
    "chunks": [
      1024
    ],
    "filters": [
      {
        "id": 1,
        "name": "deflate",
        "cd_values": [
          6
        ]
      }
    ],
    "rawType": {
      "signed": false,
      "type": 6,
      "compound_type": {
        "nmembers": 1,
        "members": [
          {
            "signed": false,
            "type": 10,
            "array_type": {
              "signed": false,
              "type": 0,
              "vlen": false,
              "littleEndian": true,
              "size": 4,
              "shape": [
                5
              ],
              "total_size": 5
            },
            "vlen": false,
            "littleEndian": true,
            "size": 20,
            "name": "col0",
            "offset": 0,
            "shape": []
          }
        ]
      },
      "vlen": false,
      "littleEndian": true,
      "size": 20,
      "total_size": 3
    }
  }
```


bennibbelink:
A thread to explore - it appears as though there might be an overflow error somewhere (total_size is negative)

```
{
  "name": "StringVals",
  "path": "/StringVals",
  "attributes": [],
  "kind": "dataset",
  "shape": [
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255,
    255
  ],
  "type": {
    "class": "String",
    "charSet": "ASCII",
    "strPad": "null-terminated"
  },
  "chunks": [
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1,
    1
  ],
  "filters": [],
  "rawType": {
    "signed": false,
    "type": 3,
    "cset": 0,
    "strpad": 0,
    "vlen": true,
    "littleEndian": true,
    "size": 4,
    "total_size": -1933710335
  }
}
```

gonuke:
Could we pin an older version of Boost while we sort this out, so tests pass?

bennibbelink:
A summary of my progress on this...

I am pretty sure that overflow was occurring because I was trying to create a hyperspace for the hashmap corresponding to the new digest type (`char[20]`).  Thus, an array of length 20 where each element is 1 byte (256 bits) resulted in a hyperspace of size 255^20 - a **really** big number.

After realizing that storing VL values this way is not feasible, I realized that we can simply conform to Boost's digest type when we hash values, but still treat the digest as `uint[5]` when we interact with HDF5.  Fortunately (probably by design?) a `uchar[20]` and `uint[5]` are both size 20 bytes, so its possible for us to cast those 20 `chars` to 5 `uints` and use the existing hdf5 backend infrastructure.

This passes tests, but I definitely want a second opinion on this approach.  I don't have much experience working with HDF5,  and I'm not sure I have the greatest handle on the bidirectional hashmap for storing VL values.

src/query_backend.h
```diff
@@ -7,19 +7,22 @@
...
 #define CYCLUS_SHA1_SIZE 20
-#define CYCLUS_SHA1_NINT 5
+#define CYCLUS_UUID_SIZE 16
+#if BOOST_VERSION_MINOR < 86
+  #define BOOST_DIGEST_NINT 5
+  using BOOST_DIGEST_TYPE = unsigned int;
+#else
+  #define BOOST_DIGEST_NINT 20
+  using BOOST_DIGEST_TYPE = unsigned char;
+#endif
```
gonuke, line 25: Given how I see this new variable used, can't we just always use the 20 x char version, and always unpack it the way you have implemented it?

gonuke: Could also do some version of
`BOOST_DIGEST_NINT = CYCLUS_SHA1_NINT * sizeof(int)`?

src/query_backend.h
```diff
@@ -1086,7 +1089,19 @@ class Sha1 {
...
+    hash_.get_digest(tmp);
+
+    for (int i = 0; i < CYCLUS_SHA1_NINT; i++) {
+        #if BOOST_VERSION_MINOR < 86
+            d.val[i] = tmp[i];
+        #else
+            d.val[i] = (static_cast<uint>(tmp[i*4]) << 24) |
+                    (static_cast<uint>(tmp[i*4 + 1]) << 16) |
+                    (static_cast<uint>(tmp[i*4 + 2]) << 8) |
+                    (static_cast<uint>(tmp[i*4 + 3]));
```
gonuke, line 1102: Should we use `sizeof(int)` and multiples instead of these magic numbers?

src/query_backend.h
gonuke: and maybe localize it to where it is used?

gonuke (changes requested):
A few thoughts to make this simpler?

src/query_backend.h
bennibbelink: > can't we just always use the 20 x char version, and always unpack it the way you have implemented it?

Boost < 1.86 still expects to store the digest as a 5 x uint, so this would fail to compile on older versions boost

> and maybe localize it to where it is used?

I localized the `#if / #else` to `Digest digest()`, and dropped the type aliases altogether.  I also changed `Digest` to a std::array, per the comment above `class Digest`.

bennibbelink:
Downstream cycamore fails to build, I'm hoping all that needs to be done is add the `BOOST_VERSION_MINOR` definition to the build

src/query_backend.h
```diff
@@ -1082,11 +1004,26 @@ class Sha1 {
...
-  /// \}
 
   Digest digest() {
     Digest d;
-    hash_.get_digest(d.val);
+    #if BOOST_VERSION_MINOR < 86
+        unsigned int tmp[5];
+    #else
+        unsigned char tmp[20];
+    #endif
```
gonuke, line 1014: I'm not sure I understand all of what's happening behind the scenes here, but it seems that it might always work to use the `unsigned char tmp[20]` option and unpack it into the `Digest` as you do below??

gonuke (changes requested):
A question here about using the `char` version for all versions of BOOST?

If I'm inferring properly (and I might not be) - the `char` version should always work since we're just treating memory as a byte array.  The `int` version happened to work for some versions of Boost.  What am I missing?

src/query_backend.h
bennibbelink: On Boost<1.86 `boost::uuids::detail::sha1::get_digest(tmp)` is declared with tmp as `unsigned int[5]`, but on Boost==1.86 `get_digest(tmp)` is declared with tmp as `unsigned char[20]`, so if we try to always declare `tmp` as a `unsigned char[20]` this will fail to compile on Boost<1.86.  

Prior to Boost 1.86, we passed the digest around as is, so our hdf5 backend infrastructure is built to handle `uint[5]` (its actually pretty generic, so long as the dimensions arent [too high](https://github.com/cyclus/cyclus/pull/1792#issuecomment-2458665109), but lets leave it as `uint[5]` since that seems to work well).  On Boost==1.86 we can no longer pass it around as is without modifying our hdf5 infrastructure, so this converts the `unsigned char[20]` array  to `unsigned int[5]` to make it compatible.  In theory we could use an arbitrary type/dimension `atype[dim]` for our hdf5 backend infrastructur, so long as we convert it to `atype[dim]` *from* the format that `get_digest()` requires (either `uint[5]` or `uchar[20]`).


Does this make sense or am I overcomplicating things?

gonuke: That's useful.  I had was stuck thinking we could just pass in a pointer to some allocated memory, and didn't think of the Boost signature being so strongly typed.

gonuke, line 1014: I'm not married to the variable name `block_size`, but how about this?
```suggestion
    unsigned int block_size = sizeof(unsigned int);
    #if BOOST_VERSION_MINOR < 86
        unsigned int tmp[CYCLUS_SHA1_NINT];
    #else
        unsigned char tmp[CYCLUS_SHA1_NINT * block_size];
    #endif
```

src/query_backend.h
```diff
@@ -1082,11 +1004,26 @@ class Sha1 {
...
+    hash_.get_digest(tmp);
+
+    for (int i = 0; i < CYCLUS_SHA1_NINT; i++) {
+        #if BOOST_VERSION_MINOR < 86
+            d[i] = tmp[i];
+        #else
+            d[i] = (static_cast<uint>(tmp[i*sizeof(unsigned int)]) << 24) |
+                    (static_cast<uint>(tmp[i*sizeof(unsigned int) + 1]) << 16) |
+                    (static_cast<uint>(tmp[i*sizeof(unsigned int) + 2]) << 8) |
+                    (static_cast<uint>(tmp[i*sizeof(unsigned int) + 3]));
```
gonuke, line 1024:
```suggestion
            d[i] = (static_cast<uint>(tmp[i*block_size]) << block_size * 6) |
                   (static_cast<uint>(tmp[i*block_size + 1]) << block_size * 4) |
                   (static_cast<uint>(tmp[i*block_size + 2]) << block_size * 2) |
                   (static_cast<uint>(tmp[i*block_size + 3]));
```

gonuke (changes requested):
Some final(?) suggestions to get rid of magic numbers.

src/query_backend.h
gonuke: Does this work for both big-/little-endian systems?

gonuke: Can we do this as 
```
d[i] = static_cast<uint>(tmp + i * sizeof(unsigned int))
```
And rely on the casting to consume 4 bytes of data?

bennibbelink: Good idea, this is much cleaner. `static_cast` can do this, and as far as I can tell it should work for both big/little endian systems

bennibbelink:
As for the downstream cycamore failure... I tried building cycamore locally on top of this PR, and all it took was a single line of CMake to resolve the build failure: `ADD_DEFINITIONS(-DBOOST_VERSION_MINOR=${Boost_VERSION_MINOR})`.  Cycamore PR incoming...

gonuke (changes requested):
Sorry - one more request to help me understand/remember all this in 18 or 24 months...

bennibbelink:
> `static_cast` can do this, and as far as I can tell it should work for both big/little endian systems

After some more research I think I may have been wrong on both of these points.  I think the only way to make this conversion safely while accounting for endianness is to do the bit shift ourselves

gonuke (approved):
This looks good - thanks for sleuthing this out @bennibbelink

PR #1793, opened 2024-09-02
Wrap cmake boilerplate in macros

### LINKED ISSUE #1788: Update cycstub to match new build infrastructure

bennibbelink:
Closes #1788.  See #1789 for discussion.


stubs/CMakeLists.txt
```diff
@@ -35,57 +25,18 @@ ENDIF()
...
-ENDIF("${isSystemDir}" STREQUAL "-1")
-
-# Tell CMake where the modules are
-SET(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_INSTALL_PREFIX}/share/cyclus/cmake")
-SET(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${STUB_SOURCE_DIR}/cmake)
-MESSAGE("--CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH}")
+cyclus_set_rpath()
 
 # Find cyclus
 FIND_PACKAGE(Cyclus REQUIRED)
```
gonuke, line 42: If we move this to the top... could we use it to set the `CMAKE_MODULE_PATH` more robustly?

gonuke (commented):
This looks good, but I'm worried about trying to guess the `CMAKE_MODULE_PATH`

bennibbelink:
Just noticed that test_stubs.py is skipped when pytest is run, it can still be executed by running `python3 test_stubs.py`. I will look into why this is being skipped

stubs/CMakeLists.txt
bennibbelink: > could we use it to set the `CMAKE_MODULE_PATH` more robustly?

I tried moving line 42 to the top but I'm pretty sure we need to set the CMAKE_MODULE_PATH in order for cmake to find our `FindCyclus.cmake` file in `$HOME/.local/share/cyclus/cmake`

This is the log from `test_stubs.py`:
```bash
/usr/bin/cmake
CMake Error at CMakeLists.txt:4 (FIND_PACKAGE):
  No "FindCyclus.cmake" found in CMAKE_MODULE_PATH.


CMake Warning (dev) at CMakeLists.txt:4 (FIND_PACKAGE):
  FindCyclus.cmake must either be part of this project itself, in this case
  adjust CMAKE_MODULE_PATH so that it points to the correct location inside
  its source tree.

  Or it must be installed by a package which has already been found via
  find_package().  In this case make sure that package has indeed been found
  and adjust CMAKE_MODULE_PATH to contain the location where that package has
  installed FindCyclus.cmake.  This must be a location provided by that
  package.  This error in general means that the buildsystem of this project
  is relying on a Find-module without ensuring that it is actually available.

This warning is for project developers.  Use -Wno-dev to suppress it.

-- Configuring incomplete, errors occurred!
  File "/tmp/tmpyubn5vng/foo/install.py", line 137, in <module>
    main()
  File "/tmp/tmpyubn5vng/foo/install.py", line 134, in main
    install(args)
  File "/tmp/tmpyubn5vng/foo/install.py", line 61, in install
    rtn = subprocess.check_call(cmake_cmd, cwd=args.build_dir,
  File "/usr/lib/python3.10/subprocess.py", line 369, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['cmake', '/tmp/tmpyubn5vng/foo', '-DCMAKE_INSTALL_PREFIX=/tmp/tmpyubn5vng/baz', '-DCYCLUS_ROOT_DIR=/home/ben/.local']' returned non-zero exit status 1.

```

bennibbelink: Hmm... you're right though.  I wonder if we should copy a tiny bit of logic from FindCyclus.cmake so that we have more confidence in where the cyclus install is (call `which cyclus`)

bennibbelink:
Removed `if __name__ == "__main__:"` at the bottom of `test_stubs.py`, now the test is executed by `pytest` and is passing.  Ready for another review.

gonuke (approved):
Thanks @bennibbelink - I just tested this locally and it all seems to work.

PR #1795, opened 2024-09-03
Recording for 1:1 extracted resources

### LINKED ISSUE #1794: Issues with recording data tables when a resources is fully extracted into another resource
#1775 raised recording issues by not appropriately recording information when a resources is fully extracted into a new resource. This was done in an attempt to limit unnecessary lines in a resource table, specifically  in the packaging process, which often results in resources being extracted 1:1 to a new, identical resource.

Fix by re-considering exactly what information needs to be recorded by the res_tracker in this situation to both (1) not mess with the recording system and also (2) not result in additional, identical lines in resource table

### LINKED ISSUE #1797: Hard checks against packaging fill_min instead of eps_rsrc can result in failed trades
In packaging, everything should be checked as a difference against `eps_rsrc()`. For example, `quantity - fill_min > -eps_rsrc()`, rather than as a direct comparison, e.g. `quantity - fill_min`.  Otherwise, rounding does result in occasional unnecessary failed trades, especially when many resources are being popped and packaged

nuclearkatie:
Closes #1794. Closes #1797 

Fixes the broken cycamore python tests

~~Still has the failures on conda 22.04 and 24.04 that I don't understand~~

~~#1790 should be merged first~~ (merged)

### DIFF
--- a/src/res_tracker.cc
+++ b/src/res_tracker.cc
@@ -23,8 +23,8 @@ void ResTracker::Create(Agent* creator) {
 
   parent1_ = 0;
   parent2_ = 0;
-  bool no_bump = true;
-  Record(no_bump);
+  bool bumpId = false;
+  Record(bumpId);
   ctx_->NewDatum("ResCreators")
       ->AddVal("ResourceId", res_->state_id())
       ->AddVal("AgentId", creator->id())
@@ -83,8 +83,8 @@ void ResTracker::Package(ResTracker* parent) {
     
     // Resource was just created, with packaging info, and assigned a state id. 
     // Do not need to bump again
-    bool no_bump = true;
-    Record(no_bump);
+    bool bumpId = false;
+    Record(bumpId);
   } else {
     // Resource was not just created. It is being re-packaged. It needs to be
     // bumped to get a new state id.
@@ -96,8 +96,8 @@ void ResTracker::Package(ResTracker* parent) {
   
 }
 
-void ResTracker::Record(bool no_bump) {
-  if (!no_bump) {
+void ResTracker::Record(bool bumpId) {
+  if (bumpId) {
     res_->BumpStateId();
   }
   ctx_->NewDatum("Resources")
--- a/src/res_tracker.cc
+++ b/src/res_tracker.cc
@@ -53,9 +53,9 @@ void ResTracker::Extract(ResTracker* removed) {
     Record();
   }
 
-    removed->parent1_ = res_->state_id();
-    removed->parent2_ = 0;
-    removed->tracked_ = tracked_;
+  removed->parent1_ = res_->state_id();
+  removed->parent2_ = 0;
+  removed->tracked_ = tracked_;
 
   removed->Record();
 }
--- a/src/res_tracker.cc
+++ b/src/res_tracker.cc
@@ -23,7 +23,8 @@ void ResTracker::Create(Agent* creator) {
 
   parent1_ = 0;
   parent2_ = 0;
-  Record();
+  bool no_bump = true;
+  Record(no_bump);
   ctx_->NewDatum("ResCreators")
       ->AddVal("ResourceId", res_->state_id())
       ->AddVal("AgentId", creator->id())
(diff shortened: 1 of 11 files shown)


nuclearkatie:
This might be totally the wrong strategy. I'll bring it up at FC&Sec meeting to discuss and then pick this back up afterwards

bennibbelink:
> Still has the failures on conda 22.04 and 24.04 that I don't understand

#1796 should resolve the conda failures

gonuke:
Rebase now that #1796 is merged should fix the conda issues

nuclearkatie:
Rebased again since #1790 was merged

src/resource.h
```diff
@@ -94,6 +95,8 @@ class Resource {
   /// Returns the package id.
   virtual std::string package_name() { return Package::unpackaged_name(); };
 
+  virtual Ptr PackageExtract(double qty, std::string new_package_name = Package::unpackaged_name()) = 0;
```
gonuke, line 98: I'm not sure you need to define a default here since this is a pure virtual function
```suggestion
  virtual Ptr PackageExtract(double qty, std::string new_package_name) = 0;
```

src/res_tracker.cc
```diff
@@ -23,7 +23,8 @@ void ResTracker::Create(Agent* creator) {
 
   parent1_ = 0;
   parent2_ = 0;
-  Record();
+  bool no_bump = true;
+  Record(no_bump);
```
gonuke, line 27: Just for my education.... this change isn't strictly related to the purpose of this PR.  It seems like we previously bumped the ID unnecessarily here, but probably decided that was fine because it only happened once per resource and it was deemed too cumbersome to avoid it. Now that we have a good reason to avoid it in some cases and have added that capability, you've gone ahead and introduced it here, too.

Do you think that's a good characterization? (recognizing that we'd both need to speculate on our past intent...)

src/res_tracker.h
```diff
@@ -42,11 +42,14 @@ class ResTracker {
...
-  void Package();
+  /// Should be called when a resource's package gets modified. If the resource
+  /// was just created from a parent resource, the parent should be passed in.
+  /// If the resource is just being repackaged (e.g. to unpackaged), the parent
+  /// should be NULL.
+  void Package(ResTracker* parent = NULL);
 
  private:
-  void Record();
+  void Record(bool no_bump = false);
```
gonuke, line 52: What are your thoughts on flipping the logic on this?
```suggestion
  void Record(bool bumpId = true);
```

My sense it that there is a little extra cognitive load to thinking through what it means to set a variable with a negative name to true....

src/res_tracker.cc
nuclearkatie: Yes, but also we create new resources a lot in packaging. So the old workflow of create new resource (comes with new state id) then record new resource (bump state ID, second line on resource table for the exact same resource we just created) would just be much more obvious because it would happen much, much more

src/res_tracker.cc
```diff
@@ -51,10 +52,10 @@ void ResTracker::Extract(ResTracker* removed) {
...
     Record();
   }
-  
-  removed->parent1_ = res_->state_id();
-  removed->parent2_ = 0;
-  removed->tracked_ = tracked_;
+
+    removed->parent1_ = res_->state_id();
+    removed->parent2_ = 0;
+    removed->tracked_ = tracked_;
```
nuclearkatie, line 58:
```suggestion
  
  removed->parent1_ = res_->state_id();
  removed->parent2_ = 0;
  removed->tracked_ = tracked_;
```

src/res_tracker.cc
nuclearkatie: Oh! You raised something important that I missed though. This is important when a new resource is created via `Material::Create`, which calls `ResTracker::Create`, but not when a resource is created via `new Material`. Right now I use `new Material`, because that's what `ExtractComp` uses. But maybe that's not great? Am I wrong in thinking that using `Material::Create` is better practice?

The only issue is that a resource, once created, does not know its creator any more. It doesn't save that information. So if I'm trying to create a new resource from within a resource, I can't actually call `Material::Create`, I don't have enough info. So in order to use `Create` from within `PackageExtract`, I think I'd have to pass the creator in to `Resource::Package`, and then down to `PackageExtract` ? Does this make sense? or is using `new` perfectly fine anyway?

nuclearkatie: I guess as long as you agree that it's not bad that `ExtractComp` currently uses `new Material` instead of `Create`, we should probably keep packaging out of `Create` anyway. Because I don't think new materials created to be packaged should end up in the `ResCreators` table since they're not being created from thin air (as with Sources) they're being created using quantity from existing resources

gonuke (approved):

PR #1796, opened 2024-09-04
Pin Boost<1.86.0

bennibbelink:
Pin boost for now until #1791 is resolved.

This was more complex than anticipated, but now the shell is modified so that we always execute RUN commands inside of a conda env called `cyclus`.  Was running into issues where the compiler was having trouble finding GLIBCXX and this seemed to resolve the issue.  I tried to make it so that it is initializing the conda environment a more traditional way, instead of just setting our path to /opt/conda/bin:$PATH

### DIFF
--- a/src/toolkit/res_map.h
+++ b/src/toolkit/res_map.h
@@ -233,7 +233,7 @@ class ResMap {
   }
 
   /// Whether quantity_ should be recomputed or not.
-  bool dirty_quantity_;
+  mutable bool dirty_quantity_;
 
   /// Current total quantity of all resources in the mapping.
   double quantity_;
--- a/src/toolkit/res_map.h
+++ b/src/toolkit/res_map.h
@@ -233,7 +233,7 @@ class ResMap {
   }
 
   /// Whether quantity_ should be recomputed or not.
-  mutable bool dirty_quantity_;
+  bool dirty_quantity_;
 
   /// Current total quantity of all resources in the mapping.
   double quantity_;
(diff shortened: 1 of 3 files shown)


gonuke:
Is this generally a better way to handle conda anyway? I.e lesson we should transfer to other projects?

bennibbelink:
> Is this generally a better way to handle conda anyway? I.e lesson we should transfer to other projects?

Maybe.  [This is a really good walkthrough](https://pythonspeed.com/articles/activate-conda-dockerfile/) of different ways you could do it, including failed solutions.  One quality that I like about this method is that it more accurately simulates a user environment, since they likely won't be working in a base environment.

src/toolkit/res_map.h
```diff
@@ -233,7 +233,7 @@ class ResMap {
   }
 
   /// Whether quantity_ should be recomputed or not.
-  bool dirty_quantity_;
+  mutable bool dirty_quantity_;
```
gonuke, line 236: This seems unrelated... any context?

gonuke (changes requested):
One question about the addition of `mutable` - what drove this to happen now/here?

src/toolkit/res_map.h
bennibbelink: On newer compilers an error will be thrown by that line.  I think I originally installed the compilers explicitly and it was installing newer versions of gcc, but by simply installing `compilers` via conda it installs a slightly older version of gcc that doesn't throw the error.  I do think installing `compilers` instead of explicitly specifying `gxx_linux-64 gcc_linux-64` is the better strategy, since conda will find the appropriate compiler packages for you based on your system.  As time goes on we will likely eventually need to make this change (especially if I can get the macOS build to work), but for now we can leave it out since it's not necessary.

It first presented itself to me when working on the macOS and using clang.

gonuke (approved):
Thanks @bennibbelink

PR #1806, opened 2024-09-20
ResTracker Extract gave the wrong parent_id to one of the child resources

### LINKED ISSUE #1805: When material is split for transacting, the parent of the second child resource is wrong
When a material is split, one of the child resources ends up with the wrong state id as its parent. It gets the other child as parent instead of the parent as parent.

Example: 
- Parent with resource id 0 and size X+Y
- Y amount is designated for trade

What should happen
- Child1 with resource id 1 is created. Size is X, parent is 0
- Child2 with resource id 2 is created. Size is Y, parent is also 0

What does happen
- Child1 with resource id 1 is created. Size is X, parent is 0
- Child2 with resource id 2 is created. Size is Y, **parent is 1**

What I think is happening
- Size Y quantity is popped of Parent. Parent has been split, so undergoes a state id bump and gets resource ID 1. THEN, child of size Y is created, gets state id 2 from creation, and is assigned parent based on the resource that has already been bumped to a new state id. So it gets the post-modification state id as parent, instead of the pre-modification state id as it should have

nuclearkatie:
Setting parent1 for the removed resource must come before the original resource is recorded. Otherwise, the original (parent) resource becomes its own child, with a new state id. Then that first child becomes the parent1 for the second child.

Closes #1805

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,6 +38,7 @@ Since last release
 * Removed GTest source code from code coverage reports (#1759)
 * Updated action versions to avoid node.js deprecation (#1802)
 * Removed the module imports in `cyclus/__init__.py` and removed the need for multi-phase initialization of Cython modules (#1809)
+* ResTracker Extract gave the wrong parent_id to one of the child resources (#1806)
 
 
 v1.6.0
--- a/src/res_tracker.cc
+++ b/src/res_tracker.cc
@@ -46,18 +46,20 @@ void ResTracker::Extract(ResTracker* removed) {
     return;
   }
 
+  // removed parent must be set before the resource is recorded, otherwise
+  // removed ends up with parent1 of the other child (which is this resource
+  // after being bumped)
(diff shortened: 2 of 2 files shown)


gonuke:
I think this was the behavior historically: https://github.com/cyclus/cyclus/blob/fdde8b45b52a925430df9de610a3d8a5b9ca064a/src/res_tracker.cc#L42
showing that both of the `Record()` actions happen last before ID's are bumped.  This may be the behavior we want to preserve.

nuclearkatie:
> I think this was the behavior historically:
> 
> https://github.com/cyclus/cyclus/blob/fdde8b45b52a925430df9de610a3d8a5b9ca064a/src/res_tracker.cc#L42
> 
> 
> showing that both of the `Record()` actions happen last before ID's are bumped. This may be the behavior we want to preserve.

Ok, I did that

gonuke (approved):
Nice catch on this @nuclearkatie

PR #1809, opened 2024-09-21
Modify Cython api to use `api` keyword 

### LINKED ISSUE #1779: Cython API is not implemented correctly

bennibbelink:
I stumbled upon [this section ](https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#c-api-declarations)of the cython docs (I can't believe I missed it before!!) that discusses how to make cython declarations available to C code.  Since we only access cython methods in PyHooks we can define those methods with the `api` keyword and avoid dealing with the messy multi-phase module initialization (see #1781).  It also does not require that `libcyclus` is linked with our cython modules since "the Python import machinery is used to make the connection dynamically" (if you look at the generated `_api.h` headers it looks like they are doing a lot of the messy Python C-API calls that I was making in #1781, but in a robust/correct way :smile:).

This PR replaces #1781 and closes #1779.

### DIFF
--- a/cyclus/__init__.py
+++ b/cyclus/__init__.py
@@ -1,2 +1,2 @@
 __version__ = 'x.x.x' # will be populated during build
-import pymodule, eventhooks, pyinfile
+# import pymodule, eventhooks, pyinfile
--- a/cyclus/__init__.py
+++ b/cyclus/__init__.py
@@ -1,2 +1,1 @@
 __version__ = 'x.x.x' # will be populated during build
-# import pymodule, eventhooks, pyinfile
--- a/.gitignore
+++ b/.gitignore
@@ -65,13 +65,17 @@ Debug/
 # generated cython ignores
 *.cc.gen
 *.cc.h
+*.cc_api.h
 *.h.gen
 src/eventhooks.cc
 src/eventhooks.h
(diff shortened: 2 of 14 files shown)


cyclus/__init__.py
```diff
@@ -1,2 +1,2 @@
 __version__ = 'x.x.x' # will be populated during build
-import pymodule, eventhooks, pyinfile
\ No newline at end of file
+# import pymodule, eventhooks, pyinfile
```
gonuke, line 2: Should we just delete this line?

gonuke (changes requested):
Thanks @bennibbelink - I'm not sure I understand all of this, but probably didn't understand it well enough before, either.  Just one little comment here.

cyclus/__init__.py
bennibbelink: Yes, thanks for catching that

gonuke (approved):
Thanks @bennibbelink

PR #1813, opened 2024-09-25
Add normal and uniform packaging strategies

### LINKED ISSUE #1777: Package R-NG schema requires all elements
Need to fix R-NG schema for packaging (and presumably transport units). `fill_min`, `fill_max`, and `strategy` should be optional. `strategy` especially should be optional in the case that `fill_min` and `fill_max` are equal

### LINKED ISSUE #1812: Random fill package? Re-do GetFillMass, again?
As I'm using packaging more, I find myself desiring a random fill strategy, even a relatively simple implementation like a uniform distribution between `fill_min` and `fill_max`.

However, this idea doesn't play nicely with `Package::GetFillMass()`, which assumes a single fill mass and returns that number plus the number of packages that can be created at that fill mass.

For some reason, I had avoided making `GetFillMass` return a vector of fill masses, even in the recent change #1790. Maybe because that felt wasteful when all the strategies would have a single fill mass (with the exception of the very last package, which could be smaller)? Is there any other reason not to switch `GetFillMass` to just returning a vector?

nuclearkatie:
Closes #1812. Also closes #1777 

Note that the new strategies do not come with any flexibility. The uniform distribution is created with min of `fill_min` and max of `fill_max`. The (truncated) normal distribution is created with mean in the middle of `fill_min` and `fill_max` and its three standard deviations to the boundaries of `fill_min` and `fill_max`. It may be of interest to someone in the future to add user inputs to change these values, but I don't think that's of critical importance now. Or, maybe it's not important enough to be changed later either! 

Because the distributions are dependent on the package values of `fill_min` and `fill_max`, they do not sample when less than `fill_max` of material is available. Otherwise, we would be sampling from only part of the distribution which would throw off the randomness. Instead, when less than `fill_max` (but more than `fill_min`) is available, that amount is used directly to fill up a partial package

### DIFF
--- a/src/package.cc
+++ b/src/package.cc
@@ -81,14 +81,14 @@ std::vector<double> Package::GetFillMass(double qty) {
   if (strategy_ == "uniform" || strategy_ == "normal") {
     // only use random if a full package amount is available. if less than one
     // full amount is available, below will fill a partial package (no random).
-    while ((qty > 0) && (qty >= fill_max_)) {
+    while ( qty > std::max(0,fill_max_) ) {
       fill_mass = dist_->sample();
       packages.push_back(fill_mass);
       qty -= fill_mass;
     }
   }
 
-  if (qty > eps_rsrc() && qty >= fill_min_) {
+  if (qty > std::max(eps_rsrc(),fill_min_) ) {
     // leftover material is enough to fill one more partial package. 
     packages.push_back(qty);
   }
--- a/src/package.cc
+++ b/src/package.cc
@@ -30,11 +30,27 @@ Package::Ptr& Package::unpackaged() {
   return unpackaged_;
 }
 
-std::pair<double, int> Package::GetFillMass(double qty) {
+void Package::SetDistribution() {
+  if (strategy_ == "uniform") {
+    dist_ = UniformDoubleDist::Ptr (new UniformDoubleDist(fill_min_, fill_max_));
+  } else if (strategy_ == "normal") {
+    dist_ = NormalDoubleDist::Ptr (new NormalDoubleDist(
+      (fill_min_ + fill_max_) / 2,
+      (fill_max_ - fill_min_) / 6,
+      fill_min_,
+      fill_max_));
+  }
+}
+
+std::vector<double> Package::GetFillMass(double qty) {
+  std::vector<double> packages;
   if ((qty - fill_min_) < -eps_rsrc()) {
     // less than one pkg of material available
-    return std::pair<double, int> (0, 0);
+    return packages;
   }
+
+  // simple check for whether vector limits *might* be exceeded
+  ExceedsSplitLimits(qty / fill_max_);
 
   double fill_mass;
   int num_at_fill_mass;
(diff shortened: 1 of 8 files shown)


nuclearkatie:
Cycamore build will fail because this updates the call to `GetFillMass`, which Cycamore:Source calls. [cycamore#630](https://github.com/cyclus/cycamore/pull/630), which needs to be merged after this, will fix this issue on the cycamore side

src/package.cc
```diff
@@ -52,9 +68,34 @@ std::pair<double, int> Package::GetFillMass(double qty) {
...
+    ExceedsSplitLimits(num_at_fill_mass);
+    packages.assign(num_at_fill_mass, fill_mass);
+
+    qty -= num_at_fill_mass * fill_mass;
+  }
+
+  if (strategy_ == "uniform" || strategy_ == "normal") {
+    // only use random if a full package amount is available. if less than one
+    // full amount is available, below will fill a partial package (no random).
+    while ((qty > 0) && (qty >= fill_max_)) {
```
gonuke, line 84: aren't these two conditions redundant?  Or maybe cleaner to do:
```suggestion
    while ( qty > std::max(0,fill_max_) ) {
```

src/package.cc
```diff
@@ -52,9 +68,34 @@ std::pair<double, int> Package::GetFillMass(double qty) {
...
+      fill_mass = dist_->sample();
+      packages.push_back(fill_mass);
+      qty -= fill_mass;
+    }
+  }
+
+  if (qty > eps_rsrc() && qty >= fill_min_) {
```
gonuke, line 91: I think I know why these are not redundant because it's not inconceivable for `fill_min_ < eps_rscr()`, but we could again do:
```suggestion
  if (qty > std::max(eps_rsrc(),fill_min_) ) {
```

src/package.h
```diff
@@ -9,6 +9,7 @@
 
 #include "error.h"
 #include "logger.h"
+#include "random_number_generator.h"
```
gonuke, line 12: I wonder if we can just forward declare the classes here rather than importing the full header?
```suggestion
class DoubleDistribution;
class DoubleDistribution::Ptr;
```
If this works, we'll need to include the header in the `package.cc` file instead.

If it works, it's preferred because it reduced the dependency graph. If it stays this way, every file that includes `package.h` will include the `random_number_generator.h` and all of the things it includes.  Any change in any file down that dependency tree causes the whole stack to be rebuilt.

If any part of the interface is uses in the package header, then it would need to be included.  But if we only need to declare that the classes exist, then this can work instead.  I'm just not sure about the `::Ptr` part..

gonuke (changes requested):
A few small suggestions.

src/package.h
nuclearkatie: I'm still getting `invalid use of incomplete type` errors on the `::Ptr`s even on my machine after including the RNG in `package.cc`... I think this means I need to include it in the header

gonuke (approved):
Thanks @nuclearkatie - this update definitely makes the output of `GetFillMass` simpler...

PR #1814, opened 2024-10-02
CI for MacOS

bennibbelink:
There have been a handful of build updates since https://github.com/cyclus/cyclus/pull/1743 that have made this simpler.  This build is entirely conda based so it hopefully should be relatively easy to update the feedstock.  

The workflow currently just builds cyclus, but we could add downstream testing in this workflow if desired.

Some issues that will follow this PR being merged: 

- Add arm64 versions in pyne/data so you don't need `--slow` on arm
- Cycamore `macos_build_test.yml`
- Cymetric `macos_build_test.yml`
- conda-forge feedstock updates


src/toolkit/res_map.h
```diff
@@ -233,7 +233,7 @@ class ResMap {
   }
 
   /// Whether quantity_ should be recomputed or not.
-  bool dirty_quantity_;
+  mutable bool dirty_quantity_;
```
gonuke, line 236: We may have discussed this before, but why do we suddenly need this? And only here?

gonuke (commented):
Just one little question

src/toolkit/res_map.h
bennibbelink: At line 100 we declare a const function which guarantees members of the class will not be modified.  In that function `dirty_quantity_` is modified, and clang throws a fatal error:
 ```clang
error: cannot assign to non-static data member within const member function 'operator[]'
  101 |     dirty_quantity_ = true;

 ```
 
 Marking `dirty_quantity_` as `mutable` allows us to modify it as we please.

gonuke: I guess it's not really a `const` function then... but I see why it needs to exist, so maybe that's why we have `mutable` to allow us to lie about the `const`-ness of a function 🤣

gonuke (approved):
Thanks for this big leap forward @bennibbelink

src/CMakeLists.txt
```diff
@@ -166,9 +166,15 @@ if(Cython_FOUND)
           TARGETS ${name}
           LIBRARY DESTINATION lib
           )
-        INSTALL(
-          TARGETS ${name}
-          LIBRARY DESTINATION ${PYTHON_SITE_PACKAGES}
+        # Add symlinks to PYTHON_SITE_PACKAGES so python can find our cython modules
+        # Python will only import libraries with .so extensions
+        # so we hardcode the symlink with .so even on APPLE
+        INSTALL(CODE "
```
ahnaf-tahmid-chowdhury, line 172: Instead of using system links, I may suggest to use rpath, this may help to build python wheels more easier, since python sdist do not copy system links while making tar file.

bennibbelink: That's a good point.  The build is configured to use rpath for linking (and can confirm via `otool`), but I'm not sure that alone addresses the purpose of this symlink.  We install the cython modules (`pymodule`, `pyinfile`, and `eventhooks`) to `PYTHON_SITE_PACKAGES` because they are imported by the python interpreter and this is more robust than adding `INSTALL_PATH/lib` to the PYTHONPATH.  

I'm thinking it might be smart to just install the libraries twice - once as `.dylib` files to `/lib` and once as `.so` files to `PYTHON_SITE_PACKAGES` (because python will not import a shared library that is a `.dylib`). This is what was done prior to this PR, however it installed it as a `.dylib` both places.  It will likely need to be hardcoded as `.so` so the python interpreter can import on macOS.

I recognize that installing the same library twice is probably not great practice, so I'm open to suggestions to make this cleaner.

PR #1816, opened 2024-10-11
Remove deprecated `exclusive_orders_only` from schema

### LINKED ISSUE #1768: `exclusive_orders_only` is deprecated
Both `cyclus_flat.rng.in` and `cyclus.rng.in` have a comment `<!--deprecated. @TODO remove in release 1.5 -->` for the element `exclusive_orders_only` within `solver`.  It looks like the schema on the website already has this removed, should be a quick fix to remove it from the actual schemas.  @gonuke can you confirm that this is deprecated and should be removed?

bennibbelink:
Closes #1768.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,6 +38,7 @@ Since last release
 
 * Removed the ResourceBuff class and replaced its instances with ResBuf (#1755)
 * Removed reference to deprecated ``ProgTranslator::Context`` class (#1811)
+* Removed ``exclusive_orders_only`` from schema (#1816)
 
 **Fixed:**
 
--- a/share/cyclus-flat.rng.in
+++ b/share/cyclus-flat.rng.in
@@ -67,11 +67,6 @@ datatypeLibrary="http://www.w3.org/2001/XMLSchema-datatypes">
                 <data type="boolean" />
               </element>
             </optional>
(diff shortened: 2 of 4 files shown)


bennibbelink:
Seeing failing tests downstream.  Looks like it's similar to the issue of numpy discussed in https://github.com/cyclus/cymetric/pull/197, but  in this case cyclus was compiled with `numpy<2.0.0` and cymetric is installed with `numpy>=2.0.0`.  Relevant error message:
```
#10 0.715 A module that was compiled using NumPy 1.x cannot be run in
#10 0.715 NumPy 2.1.2 as it may crash. To support both 1.x and 2.x
#10 0.715 versions of NumPy, modules must be compiled with NumPy 2.0.
#10 0.715 Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.
#10 0.715 
#10 0.715 If you are a user of the module, the easiest solution will be to
#10 0.715 downgrade to 'numpy<2' or try to upgrade the affected module.
#10 0.715 We expect that some modules will need time to support NumPy 2.
```

-----------------------
Also - realizing now that we don't have CI for cycamore or cymetric on 24.04

gonuke:
Does this just need a rebase to build?

gonuke (approved):
Thanks @bennibbelink

PR #1821, opened 2024-10-25
More random distributions

nuclearkatie:
Adds binomial, negative binomial, poisson, and exponential distributions.

Adds test for negative binomial in the buy policy because I'm going to implement that into cycamore:Storage as a way to model disruptions


gonuke:
This looks ready to review, just failing some of the generic test failures.

nuclearkatie:
Yeah it is, I was going to think if any other distributions should be added now since they're pretty easy to add but this is probably fine for now

gonuke:
FYI - A rebase should result in the tests passing....

gonuke (approved):
This looks good to me.  We may consider additional distributions in the future, but this collection appears to be useful and correctly implemented.

PR #1822, opened 2024-10-25
add methods to reset/unset build policy

gonuke:
Introduce methods that allow us to reset a buy policy in case we want to initialize it with better behavior.

Also allow us to unset a specific commodity.

### DIFF
--- a/src/toolkit/matl_buy_policy.cc
+++ b/src/toolkit/matl_buy_policy.cc
@@ -34,6 +34,19 @@ MatlBuyPolicy::~MatlBuyPolicy() {
     manager()->context()->UnregisterTrader(this);
 }
 
+MatlBuyPolicy& MatlBuyPolicy::ResetBehavior() {
+  throughput_ = std::numeric_limits<double>::max();
+  quantize_ = 1;
+  fill_to_ = std::numeric_limits<double>::max();
+  req_at_ = std::numeric_limits<double>::max();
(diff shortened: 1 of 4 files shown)


gonuke:
We have some generic testing problems, but this is otherwise ready for review.

src/toolkit/matl_buy_policy.cc
```diff
@@ -34,6 +34,19 @@ MatlBuyPolicy::~MatlBuyPolicy() {
     manager()->context()->UnregisterTrader(this);
 }
 
+MatlBuyPolicy& MatlBuyPolicy::ResetBehavior() {
```
nuclearkatie, line 37: Does it makes sense to reset the capacity of the total inv tracker?

nuclearkatie: I guess that's supposed to be managed by the agent (since the buy pol isn't aware of any other buffers but `buf_`) and only set up within the buy pol when the agent passes NULL, so maybe not?

gonuke: I think that's right.  The `TotalInvTracker` has a behavior independent of the individual `ResBuf` by design, although an archetype developer will want to make sure their relationship makes sense as they change some of these things.

nuclearkatie (approved):
Nice to have this, thanks @gonuke!

PR #1829, opened 2024-11-03
Added most basic implementation of facility cost header code injection

dean-krueger:
This is a followup on the work @bam241 was doing with the position implementation using what we've been calling code injection instead of manually writing out all the header requirements for position. In this simple demo, a new common feature called facility_cost is made which gives facilities a capital_cost variable, which they can then print out to the terminal during simulations. More complex functionality is planned, but this PR mostly just shows that it's possible to do this, and roughly how to implement this type of change.

Possibly very useful for @nuclearkatie's work

Tested this with cycamore Source and it worked as described in the cycpp file (you just add the # include in the private section and it all works).

I don't think this should cause any issues, since it does nothing unless you add the include in cycamore.


src/toolkit/facility_cost.cc
```diff
@@ -0,0 +1,25 @@
...
+FacilityCost::FacilityCost() : capital_cost_(0) {}
+
+FacilityCost::FacilityCost(double capital_cost) {
+  capital_cost_ = capital_cost;
+}
+
+FacilityCost::~FacilityCost() {}
+
+void FacilityCost::EnumerateCosts(){
+    std::cout<<capital_cost_<<std::endl;
```
gonuke, line 21: Probably best for this to return a string rather than writing to stdout

gonuke (changes requested):

gonuke:
I wonder if we can test this general functionality using the simple agents that are part of cyclus?

src/toolkit/facility_cost.cc
dean-krueger: You make a good point. It may be best to just write a getter for this value and call that the test function. EnumerateCosts was just a stand-in that showed you could give it a function, but I don't know that it really serves a purpose in the grand scheme of what we'd want to do with this.

dean-krueger:
I'll try to add it to the Cyclus Agents and test with that. If I get one to work, should I add it to all of them?

src/toolkit/facility_cost.cc
gonuke: It is probably useful to have a way to dump all the data (that will eventually be there) in a nice string format.  Some times this is just a `str()` function that is called to convert to a string.  (recall this from Tricycle)

dean-krueger: Sounds good. I made EnumerateCosts return a string.

dean-krueger:
Went in and tried to test this with the Cyclus agents, but ran into a problem. If you try to use Cyclus Source you get an error saying that an outdated version of a Pandas dependency called bottleneck is installed. Wasn't sure if this was an issue I should dive into or not. I assume it just needs to be added to the Dockerfile?

src/toolkit/facility_cost.cc
```diff
@@ -0,0 +1,28 @@
...
+
+namespace cyclus {
+namespace toolkit {
+
+// Interested in thoughts on the member initializer list
+FacilityCost::FacilityCost() : capital_cost_(0) {}
+
+FacilityCost::FacilityCost(double capital_cost) {
+  capital_cost_ = capital_cost;
+}
```
gonuke, line 16: Use initialization list here, too
```suggestion
FacilityCost::FacilityCost(double capital_cost) : 
   capital_cost_(capital_cost) {};
```

gonuke, line 16: If these are just initialization list constructors, they can be defined in the header

src/toolkit/facility_cost.cc
```diff
@@ -0,0 +1,28 @@
...
+
+FacilityCost::~FacilityCost() {}
```
gonuke, line 18: This can be defined in the header.  It doesn't actually need to be defined since it has default behavior

src/toolkit/facility_cost.h
```diff
@@ -0,0 +1,43 @@
...
+class FacilityCost {
+ public:
+
+  /// The default constructor for FacilityCost. This creates an object
+  /// with all costs set to 0.0.
+  FacilityCost();
+
+  /// FacilityCost constructor with capital_cost as a double
+  /// @param capital_cost Capital cost of the facility.
+  FacilityCost(double capital_cost);
```
gonuke, line 27: Could have just one constructor with default value of 0.

gonuke (changes requested):
Don't forget to modify the Changelog file, as well.

src/toolkit/facility_cost.h
```diff
@@ -0,0 +1,45 @@
...
+/// but it will first serve as a very basic showcase of code injection into
+/// Facility header files for state variables which can be common to all
+/// Facilities.
+
+class FacilityCost {
+ public:
+
+  /// The default constructor for FacilityCost. This creates an object
+  /// with all costs set to 0.0.
+  FacilityCost() : capital_cost_(0) {}
```
gonuke, line 23: Here's what I was thinking for combining these
```suggestion
  FacilityCost(double capital_cost = 0) : capital_cost_(capital_cost) {}
```

gonuke (commented):
Just offering clarity on one of my previous suggestions

src/toolkit/facility_cost.h
dean-krueger: Oh! I was being dense. This makes way more sense than what I thought you meant.

gonuke (approved):
Thanks @dean-krueger - this looks good

PR #1827, opened 2024-10-28
Add binary distributions for doubles and integers

gonuke:
This adds a new random distribution that samples between two options with a single probability.

PR #1596, opened 2023-10-09
Buy policy can cycle back and forth between actively buying and dormant based on time step

nuclearkatie:
Two new buy policy parameters, `active` and `dormant`. Both are integers representing a number of time steps.

- For `active` part of the cycle, buying behavior is unchanged
- For `dormant`, agent will not request any new material from DRE

PR #1830, opened 2024-11-05
Active/dormant cycle lengths database table

nuclearkatie:
With all the new active/dormant cycle options in the buy policy, spurred on by random distributions (#1634, #1596,  #1827, #1821), inventory space alone does not govern whether buy-policy-implementing agents request materials at any given time step. However, we weren't recording the length of active/dormant cycles anywhere and it's not straightforward to pull from other tables. 

 I found myself wishing that I could hone in on details about active/dormant lengths. For example, when using the new binary distribution (#1827) I want to see easily if/when the simulation undergoes a disruption. It also provides a convenient way to pull the distribution of cycle lengths separate from relying on demand and transactions, which are confounded with additional information.

A broader way to approach this idea might be to have a table that records all the random numbers pulled from the simulation-managed PRNG. In fact, that might still be useful? I think that this table is useful because it can also record the relevant agent and the relevant time, which isn't necessarily the current time

This is still a draft because I'm still seeing how my own need for this info involves. But I opened a PR because I'm happy to take requests or feedback

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,7 @@ Since last release
 * Methods to reset behavaiors of MaterialBuyPolicy (#1822)
 * More random distributions for the random number generator (#1821, #1827)
 * Added example of header code injection for facility cost (#1829)
+* Table for recording active/dormant cycle periods (#1830)
 
 **Changed:**
 
--- a/src/toolkit/matl_buy_policy.cc
+++ b/src/toolkit/matl_buy_policy.cc
@@ -348,18 +348,11 @@ void MatlBuyPolicy::SetNextActiveTime() {
   int active_length = active_dist_->sample();
   next_active_end_ = active_length + manager()->context()->time();
   if (manager() != NULL) {
-    std::string adtype;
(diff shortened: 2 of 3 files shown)


katyhuff:
@nuclearkatie -- do you happen to have a quick sentence or two of motivation here?

nuclearkatie:
Didn't think anyone would be looking too closely at new draft PRs on election night so got a little lazy...

gonuke:
> Didn't think anyone would be looking too closely at new draft PRs without tagged reviewers on election night so got a little lazy... 

lol! We all need a distraction!

gonuke:
I think recording all the random numbers would be messy, requiring a lot of reconstructing the simulation. Probably better to do what you are doing here.

gonuke:
This PR reminds me how nuanced the sampling of active/dormant times are. In particular, there is some subtle assumed knowledge of the state when the sampling of the dormant times happens.  I say this because the logic allows for there to be no sampling of the dormant times under some conditions.  We may need some more comments for posterity here.

PR #1834, opened 2024-11-17
Add CI to trigger feedstock rebuild on push to `main`

bennibbelink:
The `dev` branch of `conda-forge/cyclus-feedstock` will build a conda package from the `main` branch of cyclus, but has no way of knowing when new commits are made to `cyclus/cyclus`.  This PR adds CI to automate the process of rebuilding those `dev` conda packages when our cyclus `main` branch has new commits.  

The new CI job does the following:
1. Syncs the fork of cyclus-feedstock (both main and dev branches)
2. On the dev branch of a cyclus-feedstock fork, increment the build number and rerender the feedstock
3. Commit and push these changes to a new branch
4. Create a PR into conda-forge/cyclus-feedstock:dev with the `automerge` label, so if CI passes it will get merged automatically

I tested this on my forks (using different triggers for debugging's sake) and auto-created [this PR](https://github.com/conda-forge/cyclus-feedstock/pull/97) as a POC (in draft state so it doesn't get automerged).  To implement this in the main `cyclus/cyclus` repo it would require that we create a fork of `conda-forge/cyclus-feedstock` owned by the `cyclus` organization so that all PRs to `conda-forge/cyclus-feedstock` come from this organization-owned fork.  

I also created a personal access token (with repo access) and exposed it as a repository secret `FEEDSTOCK_TOKEN` in `bennibbelink/cyclus`.  This is how I enabled the actions run in `bennibbelink/cyclus` to have permissions to write to `bennibbelink/cyclus-feedstock`.  I am not sure what the equivalent of this permissioning looks like at the organization level, but we need some way to give the workflow in `cyclus/cyclus` access to `cyclus/cyclus-feedstock`.  @gonuke do you have a better idea of what options are available at the organization level to make this work? I've looked at [these docs](https://docs.github.com/en/organizations/managing-programmatic-access-to-your-organization/about-programmatic-access-in-your-organization) but I feel like we had a similar issue when configuring package write permissions, and the methods listed in those docs don't look familiar to me.

TODO: 

- [ ] remove --draft flag
- [ ] reference cyclus/cyclus-feedstock instead of bennibbelink/cyclus-feedstock

### DIFF
--- a/.github/workflows/publish_latest.yml
+++ b/.github/workflows/publish_latest.yml
@@ -191,7 +191,6 @@ jobs:
       - name: Set up Miniconda
         uses: conda-incubator/setup-miniconda@v3
         with:
-          python-version: "3.12"
           miniforge-version: "latest"
           conda-solver: "libmamba"
           auto-update-conda: true
--- a/.github/workflows/publish_latest.yml
+++ b/.github/workflows/publish_latest.yml
@@ -227,4 +227,4 @@ jobs:
             --draft \
             --label automerge \
             --title "Cyclus commit \`main - ${SHORT_SHA}\`" \
-            --body "Auto-triggered PR from push to Cyclus `main` branch.  If CI passes this will be merged automatically."
+            --body "Auto-triggered PR from push to Cyclus \`main\` branch.  If CI passes this will be merged automatically."
--- a/.github/workflows/publish_latest.yml
+++ b/.github/workflows/publish_latest.yml
@@ -215,4 +215,4 @@ jobs:
           git add .
           git commit -m "Trigger build for Cyclus commit ${SHORT_SHA}"
           git push --set-upstream origin ${NEW_BRANCH}
-          gh pr create --repo conda-forge/cyclus-feedstock --base dev --draft --label automerge --title "Cyclus `main - ${SHORT_SHA}`"
+          gh pr create --repo conda-forge/cyclus-feedstock --base dev --draft --label automerge --title "Cyclus `main - ${SHORT_SHA}`" --fill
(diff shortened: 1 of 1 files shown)


bennibbelink:
Unfortunately, my strategy of using a personal access token doesn't work at an organization level - it is [recommended](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#about-personal-access-tokens) to use Github Apps for organization wide access.  

I can see how feasible it would be to write a Github App to manage this syncing of repo/feedstocks. IMO that would be the cleanest and most robust solution.  The other option would be to write a cronjob action to run in the `cyclus/cyclus-feedstock` fork and auto-create PRs if commits have been made since the last cronjob run.

bennibbelink:
I wrote a [cronjob workflow](https://github.com/bennibbelink/cyclus-feedstock/blob/cyclus_cronjob/.github/workflows/cyclus_cronjob.yml) to create automergeable PRs in the upstream cyclus-feedstock, but once again ran into token permission issues when just using the standard `github.token` for permissions.  In order for this approach to work, an individual would need to create a PAT and set it in a secret in the cyclus/cyclus-feedstock.  I don't think its ideal to have this infrastructure reliant on a single user.

That being said... right now I have the cronjob in my feedstock fork's `main` branch so it should run nightly and create auto-mergeable PRs if there have been cyclus commits.  This is not a permanent solution, but for now I'd like to view it as a POC for this sort of automation infrastructure.

-----------
I still think (but am not certain) that the "correct" way to implement this would be to use a Github App, but of course that adds some complexity and is another thing to maintain.

bennibbelink:
Closing this PR.  I created an [issue](https://github.com/conda-forge/cyclus-feedstock/issues/106) in the feedstock repository.

PR #1837, opened 2024-12-09
Set locale before (de)serializing data for SQLite

bennibbelink:
Closes https://github.com/cyclus/cymetric/issues/205.  

I think there must be some inconsistency with how `libc++` and `libboost` handle default locales.  This PR calls `ss.imbue()` to set the locale according to the user's environment, which seems to solve the segfault describe in https://github.com/cyclus/cymetric/issues/205 (at least when I build and test on my arm Mac machine).  We could hardcode this to the "C" locale if we want to make it more robust.  The instance where I could see issues arising is if a user runs a sim in one environment then reads the SQLite DB in a different environment with a different locale.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -52,6 +52,7 @@ Since last release
 * No longer use deprecated Numpy Cython api when Cython>=3 (#1811)
 * ResTracker Extract gave the wrong parent_id to one of the child resources (#1806)
 * Support Boost v1.86 (#1792)
+* Set locale when writing/reading from serialized SQLite data (#1837)
 
 
 v1.6.0
--- a/src/sqlite_back.cc
+++ b/src/sqlite_back.cc
@@ -2,6 +2,7 @@
 
 #include <iomanip>
 #include <sstream>
+#include <locale>
 
 #include <boost/lexical_cast.hpp>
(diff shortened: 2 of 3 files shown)


gonuke (commented):
I think we decided to explore the idea of selecting a locale and storing it in the DB so that it could be read for post-processing?  Or did we decide to leave this for a future feature?

bennibbelink:
I thought we decided to store it in the DB.  That being said however, I'm struggling with the implementation because the place that we need to read the locale from the DB is in the code interfacing with the DB.  

I'm thinking we read the locale from the environment, falling back to the C locale if not specified.  This leaves us with the potential incompatibility of users writing archetypes named using non-ASCII characters.  I can try to add tests for this and appropriate error handling so it gracefully fails in this case.

bennibbelink:
I couldn't find a character set that failed to serialize correctly, so right now all I added are a few SqliteBackTests with characters from various non-posix locales.  Maybe there is a platform/environment where this locale management isn't as robust, but I haven't been able to produce such an environment.  I was also able to run sims with various non-posix characters in the input file, and can view them correctly in the output DB via SqliteBrowser with no issues.

gonuke (approved):
This looks good to me - thanks @bennibbelink

PR #1839, opened 2024-12-16
Update mamba management in Dockerfile to reflect new mamba cli

bennibbelink:
The latest miniforge release upgrades to mamba v2, which has a slightly different CLI than v1.  Seeing the publish_latest.yml workflow fail after #1837.  This PR updates the Dockerfile to conform to both versions of the mamba CLI.  The biggest discrepancy was the method to initialize the shell (`mamba init bash` vs. `mamba shell init --shell bash`), but since we change the Dockerfile shell to run all commands within `mamba run -n cyclus /bin/bash -c ...` we don't actually have to make any changes to the `.bashrc` within the container.


gonuke (approved):
Thanks for the quick fix @bennibbelink

PR #1844, opened 2025-02-07
Cep 30 Code Implementation

dean-krueger:
This Draft PR is meant to keep those involved in reviewing CEP 30 informed about how the implementation of the code is shaping up, offer comments, etc. I am making this a Draft PR, since I expect it will need to evolve over time. Some things that are important to point out:

- This implementation ensures backwards compatibility (I think) by urging users to make pragma vars that have a default value of some sort. I do not know if there's a better way to enforce this, but as long as whoever's adding things reads the instructions, they should be good. This may not be the best way to do this, and I'm open to suggestions.
- This implementation takes advantage of the previously explored code injection to greatly simplify the management of agents which might want to use this code. Only three lines in each agent are required to add, and keep up to date with, every financial parameter which gets added.
- This implementation specifically uses getter and setters for `financial_data_` within the `Agent` class to make sure that, if the method of storing the information in financial_data_ should change for some reason, any places where EconManagers are used will not need to be revised. Only the EconManager getter and setter functions will need to be touched, and it should extend to the rest of the code.

I tested this implementation using the `cycamore::Source` agent, and everything worked well. I have not tested it using any of the other agents, but I can't think of a reason why they'd be different. 

Currently, if you wanted to add a parameter to `EconManager` that was facility specific, you'd have to add a pragma in that facility's header file, and then after calling `InitializeCosts()` in `EnterNotify()` add a line like `this->SetEconParam("<your_param>", this->your_param)` I think. I have not tested this. It would probably be sleeker to somehow add a function like `AddParam()` to the code injection which gave users the ability to add a parameter to the `econ_params` list so that they could just mess with the header instead of kludging something into `EnterNotify()`

### DIFF
--- a/src/toolkit/facility_cost.cycpp.h
+++ b/src/toolkit/facility_cost.cycpp.h
@@ -78,7 +78,8 @@ double cost_override;
 
 // This is for testing right now, need to think/talk about the "right" way to 
 // do this...
-double prev_cost = 0;
+double avg_per_unit_cost = 0;
+double total_qty_purchased = 0;
 
 
 // Must be done in a function so that we can access the user-defined values
@@ -95,11 +96,15 @@ std::unordered_map<std::string, double> InitializeParamList() const override {
   return econ_params;
 }
 
-double GetCost(double units_of_production, double input_cost) {
+// Added production_capacity_per_timestep to allow units_of_production to
+// take the actual production of that bid into account... Not totally sure
+// how I feel about all this at the moment.
+double GetCost(double production_capacity_per_timestep, 
+               double units_of_production, double input_cost) {
 
   if (cost_override > 0) {
     // Not totally sure how to levelize this yet, just sort of testing things...
-    return cost_override + input_cost;
+    return cost_override * units_of_production + input_cost;
   }
 
   // Economic Parameters (declared like this because of scoping with try{})
@@ -137,7 +142,7 @@ double GetCost(double units_of_production, double input_cost) {
   }
 
   double timesteps_per_year = kDefaultTimeStepDur * 12 / context()->dt();
-  double k = units_of_production * timesteps_per_year;
+  double k = production_capacity_per_timestep * timesteps_per_year;
   double variable_cost = labor_cost + input_cost;
 
   double property_tax = p * v;
@@ -152,7 +157,7 @@ double GetCost(double units_of_production, double input_cost) {
   double cost = w + f + c * (property_tax + delta); 
 
   // Since pref = 1/cost (for now) we CANNOT return 0
-  return cost != 0 ? cost : 1;
+  return cost != 0 ? (cost * units_of_production) : 1;
 }
 
 // Required for compilation but not added by the cycpp preprocessor. Do not
--- a/agents/null_region.cc
+++ b/agents/null_region.cc
@@ -6,6 +6,10 @@ NullRegion::NullRegion(cyclus::Context* ctx) : cyclus::Region(ctx) {}
 
 NullRegion::~NullRegion() {}
 
+void NullRegion::EnterNotify() {
+  InitializeCosts();
+}
+
 extern "C" cyclus::Agent* ConstructNullRegion(cyclus::Context* ctx) {
   return new NullRegion(ctx);
 }
--- a/agents/null_region.h
+++ b/agents/null_region.h
@@ -12,6 +12,8 @@ class NullRegion : public cyclus::Region {
   NullRegion(cyclus::Context* ctx);
   virtual ~NullRegion();
 
+  void EnterNotify();
+
   virtual std::string version() { return cyclus::version::describe(); }
 
   #pragma cyclus
@@ -20,6 +22,9 @@ class NullRegion : public cyclus::Region {
                               "institutions but exhibits null behavior. " \
                               "No parameters are given when using the " \
                               "null region."}
+
+  private:
+  #include "toolkit/region_cost.cycpp.h"
 };
 
 }  // namespace cyclus


lewisgross1296 (commented):
FYI In C++, you can multi line comment like this
```C++
/* 
This is a 
multi-line 
comment 
*/
```

EDIT: Looks like for [doxygen](https://www.mitk.org/images/1/1c/BugSquashingSeminars%242013-07-17-DoxyReference.pdf) you use `/**` or `/*!` such as 
```C++
/**
This is a 
multi-line 
comment 
that Doxygen
sees
*/
```

src/toolkit/facility_cost.cycpp.h
gonuke, line 1: Assuming we don't need this file anymore, either?

gonuke: Or maybe we do???

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,190 @@
...
+  return econ_params;
+}
+
+// Add the financial parameters to the class
+void InitializeCosts() {
+  std::unordered_map<std::string, double> econ_params = InitializeParmList();
+  for (const auto& parameter : econ_params) {
+    this->SetEconParameter(parameter.first, parameter.second);
+  }
+}
```
gonuke, line 82: This could be defined as part of the `EconomicEntity` class and require all children to implement the `InitializeParmList` method

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,190 @@
...
+
+// Helper functions for GetCost
+double DiscountedThroughputSum(double xt, double gamma, int T, double k) const {
+  double sum = 0.0;
+  for (int t = 1; t <= T; ++t) {
+    sum += std::pow(xt, t) * std::pow(gamma, t);
+  }
+  return k * sum;
+}
```
gonuke, line 91: This seems like something that could/should be defined in the `EconomicEntity` class.  We should also talk about what classes are needed and whether there are closed form solutions to these that don't require sums.

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,190 @@
...
+  for (const auto& parameter : econ_params) {
+    this->SetEconParameter(parameter.first, parameter.second);
+  }
+}
+
+// Helper functions for GetCost
+double DiscountedThroughputSum(double xt, double gamma, int T, double k) const {
+  double sum = 0.0;
+  for (int t = 1; t <= T; ++t) {
+    sum += std::pow(xt, t) * std::pow(gamma, t);
```
gonuke, line 88: Why not:
```suggestion
    sum += std::pow(xt * gamma, t);
```

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,190 @@
...
+  }
+  return k * sum;
+}
+
+double DiscountedFixedCostSum(double F, double gamma, int T) const {
+  double sum = 0.0;
+  for (int t = 1; t <= T; ++t) {
+    sum += F * std::pow(gamma, t);
+  }
+  return sum;
```
gonuke, line 98: Perhaps this?
```suggestion
  for (int t = 1; t <= T; ++t) {
    sum += std::pow(gamma, t);
  }
  return F * sum;
```

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,190 @@
...
+// Add the financial parameters to the class
+void InitializeCosts() {
+  std::unordered_map<std::string, double> econ_params = InitializeParmList();
+  for (const auto& parameter : econ_params) {
+    this->SetEconParameter(parameter.first, parameter.second);
+  }
+}
+
+// Helper functions for GetCost
+double DiscountedThroughputSum(double xt, double gamma, int T, double k) const {
```
gonuke, line 85: Except for some notation, this does essentially the same thing as `DiscountedFixedCostSum`

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,190 @@
...
+  }
+  return sum;
+}
+
+double DiscountedVariableCostSum(double variable_costs, double xt, double gamma, int T, double k) const {
+  double sum = 0.0;
+  for (int t = 1; t <= T; ++t) {
+    sum += variable_costs * k * std::pow(xt, t) * std::pow(gamma, t);
+  }
+  return sum;
```
gonuke, line 106:
```suggestion
  double xt_gamma = xt * gamma;
  for (int t = 1; t <= T; ++t) {
    sum += std::pow(xt_gamma, t);
  }
  return variable_costs * k * sum;
```

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,190 @@
...
+
+double DiscountedFixedCostSum(double F, double gamma, int T) const {
+  double sum = 0.0;
+  for (int t = 1; t <= T; ++t) {
+    sum += F * std::pow(gamma, t);
+  }
+  return sum;
+}
+
+double DiscountedVariableCostSum(double variable_costs, double xt, double gamma, int T, double k) const {
```
gonuke, line 101: Another function that is mathematically the same as the others. We should revisit the math for all this on paper to simplify.

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,190 @@
...
+
+double ComputeDelta(double depreciation_constant, int initial_book_value,
+                              int T_hat, double gamma, double alpha) const {
+  std::vector<int> book_value = {initial_book_value};
+  std::vector<double> dt;
+
+  for (int i = 0; i < T_hat - 1; ++i) {
+    int current_depreciation = static_cast<int>(-depreciation_constant * book_value[i]);
+    book_value.push_back(book_value[i] + current_depreciation);
+    dt.push_back(-static_cast<double>(current_depreciation) / book_value[0]);
```
gonuke, line 117: Why are we casting these back and forth between integers and doubles?

gonuke (changes requested):
I've recommended a few changes here, but probably worth discussing this more in person

src/toolkit/facility_cost.cycpp.h
dean-krueger: We decided to leave this as code injection for now, so these files are all necessary

src/agent.cc
```diff
@@ -235,4 +235,12 @@ void Agent::AddToTable() {
...
 }
 
+double Agent::GetEconParameter(const std::string& key) const {
+  return GetParameter(key); 
+}
+
+void Agent::SetEconParameter(const std::string& key, double value) {
+  SetParameter(key, value);
+}
+
```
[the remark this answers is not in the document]
dean-krueger: Agreed. This is an artifact from Chat not really understanding what was needed. I believe its reasoning was that this way each subclass has to consciously define the function (ie you can't forget to override and get weird behavior), but that creates duplicate code and we're only really very likely to ever use this in Agent as the "default", so it doesn't really make sense to do it that way. I'll switch it around.

src/agent.h
```diff
@@ -411,6 +412,10 @@ class Agent : public StateWrangler, virtual public Ider {
   /// Returns a list of children this agent has
   inline const std::set<Agent*>& children() const { return children_; }
 
+  /// Override for the economic_entity functions (required for implementation)
+  double GetEconParameter(const std::string& key) const override;
+  void SetEconParameter(const std::string& key, double value) override;
+
```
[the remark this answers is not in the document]
dean-krueger: With the changes above, you are correct. These are no longer necessary.

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,190 @@
...
 #pragma cyclus var { \
-"default": 0.0, \
-"uilabel": "Capital cost required to build facility", \
-"doc": "Capital cost required to build facility" \
+    "default": 0.04, \
+    "uilabel": "Compounding fractional annual increase in the cost of labor as a decimal", \
+    "doc": "Cost in dollars of labor required to produce one unit of production", \
+    "units": "Dimensionless" \
+    }
+double annual_labor_cost_increase_factor;
```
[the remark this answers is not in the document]
dean-krueger: Good catch. And agreed, but I'm not sure how to do this... For now, this should be fine since this is unlikely to change a ton and it's not TOO hard to just be careful (I have instructions in the comments at the top of the header that it needs to go in there), though obviously it's still possible to make mistakes like this.

dean-krueger:
Made changes responding to (I think) all of @gonuke's comments. Still need to add a cost_override user input and change T_hat to user input (was previously = T, changed to 11 while bug hunting), but it's getting a little late now, so I'll do that tomorrow.

src/economic_entity.h
```diff
@@ -0,0 +1,72 @@
...
+      return 0.0;
+    }
+
+    // Protects against divide by zero in discount factor
+    if (std::abs(r + 1.0) < 1e-12) {
+      return static_cast<double>(T);
+    }
+
+    double discount_factor = 1.0 / (1.0 + r);
+    double modified_discount_factor = discount_factor * cdf;
```
gonuke, line 40: I'm a fan of this `cdf` factor being tied to the annual payment rather than tied to the discount factor, but maybe it is difficult to do that mathematically and still use the closed form solution.

src/economic_entity.h
```diff
@@ -0,0 +1,72 @@
+#ifndef ECONOMIC_ENTITY_H
+#define ECONOMIC_ENTITY_H
+
+#include <stdexcept>
+#include <string>
+#include <unordered_map>
+
+class EconomicEntity {
+ public:
+  virtual double GetEconParameter(const std::string& key) const {
```
gonuke, line 10: I've forgotten (again) why we need these functions instead of just calling `GetParameter()` and `SetParameter()`?

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,159 @@
...
+    // If any of the above functions fail to get their parameters, return 1
+    // to default to old pref logic
+    LOG(cyclus::LEV_INFO1, "GetCost") << prototype() 
+                                      << "failed to get financial_data_: "
+                                      << e.what();
+    return 1;
+  }
+
+  double seconds_per_year = kDefaultTimeStepDur * 12;
+  double k = units_of_production * seconds_per_year / context()->dt();
```
gonuke, line 134:
```suggestion
  double time steps_per_year = kDefaultTimeStepDur * 12 / context()->dt();
  double k = units_of_production * timesteps_per_year;
```

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,159 @@
...
+      {"operations_and_maintenance", operations_and_maintenance},
+      {"facility_lifetime", facility_lifetime},
+      {"capacity_decline_factor", capacity_decline_factor},
+      {"per_unit_labor_cost", per_unit_labor_cost},
+      {"annual_labor_cost_increase_factor", annual_labor_cost_increase_factor}};
+
+  return econ_params;
+}
+
+double ComputeDelta(double depreciation_constant, double initial_book_value,
```
gonuke, line 77: If we're injecting this, let's put it in the `EconomicEntity` class? And maybe discuss ways to generalize it if it's not general enough already.

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,159 @@
...
+  
+  double delta_partial_sum = 0.0;
+  for (int t = 1; t < T_hat; ++t) {
+    delta_partial_sum += dt[t - 1] * std::pow(discount_factor, t); 
+  }
+
+  return (1 - alpha * -delta_partial_sum) / (1 - alpha);
+}
+
+double GetCost(double units_of_production, double input_cost) {
```
gonuke, line 98: By contrast, we should inject this (and not put it in the `EconomicEntity` class), because it depends on the specific class variables that we are injecting.

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,159 @@
...
+                                      << e.what();
+    return 1;
+  }
+
+  double seconds_per_year = kDefaultTimeStepDur * 12;
+  double k = units_of_production * seconds_per_year / context()->dt();
+  double variable_cost = labor_cost + input_cost;
+
+  double property_tax = p * v;
+  double F = OM + property_tax;
```
gonuke, line 138: Since property tax is proportional to `v` it makes more sense to me to combine this with `delta`.  (This is the way I teach it 😀 )

gonuke (changes requested):
Some more design comments

src/economic_entity.h
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  void InitializeCosts() {
+    std::unordered_map<std::string, double> econ_params = InitializeParamList();
+    for (const auto& parameter : econ_params) {
+      this->SetEconParameter(parameter.first, parameter.second);
+    }
+  }
+
+  double ComputePresentValue(double r, int T, double cdf = 1.0) const {
```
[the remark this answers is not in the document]
dean-krueger: I'm for sure in favor of a different name. I found myself trying to think of an appropriate name for long enough yesterday that I just sort of picked one and went with it. Also agreed that more financial math functions would be good here, I mostly just wanted to get the basics done (there was a weird, sort of hard to find bug in my code that took some time to find). I'll add this to my todo list for today.

src/toolkit/facility_cost.cycpp.h
dean-krueger: Sure. I see the value in this. Just to avoid push/pull I'll change this manually, but good suggestion.

src/toolkit/facility_cost.cycpp.h
dean-krueger: Yep. I had this same thought last night. It would probably also serve us and future users well to rename this.

src/toolkit/facility_cost.cycpp.h
dean-krueger: Makes sense to me! I did it this way mostly to avoid too much thinking while I was trying to match the code answer to the hand-calc (easier when the formulas match exactly), but now that they do this can get moved.

src/toolkit/facility_cost.cycpp.h
```diff
@@ -1,19 +1,159 @@
...
+  std::vector<double> book_value = {initial_book_value};
+  std::vector<double> dt;
+
+  double discount_factor = 1.0 / (1.0 + r);
+
+  for (int i = 0; i < T_hat - 1; ++i) {
+    int current_depreciation = -depreciation_constant * book_value[i];
+    book_value.push_back(book_value[i] + current_depreciation);
+    dt.push_back(current_depreciation / book_value[0]);
+  }
```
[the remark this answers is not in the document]
dean-krueger: You are correct, and also this was one of those "I just want it to work" direct solutions that avoided too much cleverness where I could potentially make mistakes. Now that it's (nominally) working, this seems like a good change to make. Also, as far as the different schemes, the "idea" was that single vs double declining balance differs by a factor of two, and then the combined scheme just does the max of those two options, so I think there's probably a clever way to do that which won't be too complicated. I'll toss it on the to-do pile for the day.

src/economic_entity.h
gonuke: I actually feel like you convinced me last time to keep them separate, but I can't remember why, and maybe those reasons have changed based on other design changes??

src/toolkit/facility_cost.cycpp.h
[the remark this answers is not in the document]
gonuke: But there is also straight line depreciation...

src/toolkit/facility_cost.cycpp.h
dean-krueger: I'll need to check the math on this, but I think xt actually shouldn't be raised to "t" in this way... I think I did it wrong from the beginning (and also wrong in my hand calc).

src/economic_entity.h
dean-krueger: I thought about the cyclus-y part of this and realized that everywhere I used this I asked the user for times in years, so the size of a cyclus tilmestep isn't playing in.

src/economic_entity.h
dean-krueger: The original reason I did it that way was to compartmentalize the calls to those functions in case they ever needed to change (the idea being that whoever was changing them could just change the functions those call and not have to change things everywhere in the code), but then at one point you brought up that even though I did that, if the way those functions work from the outside changes we'd still have to do that.

src/toolkit/facility_cost.cycpp.h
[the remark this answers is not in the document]
dean-krueger: This was a logic issue on my part. The intended functionality of this was to have the user choose between straight line (1/N), double declining (2/N) by choosing a "depreciation constant" which was either 1 or 2 and then dividing it out by `T`. When I implemented it I accidentally made it so that the user passed in the full `depreciation_constant/T_hat` value which I had set to 0.2 (which would have been correct for the specific test case I was using) and then the math "worked out" so I didn't catch it. Either way, I'll look into making this more straightforward in the way you suggested.

dean-krueger:
> A quick thought before I forget: it might be nice if transactions recorded the cost that they were transacted at. Not sure if they do this already (by just recording the preference), but this is equal parts a reminder for me to check on that and a broadcast of intentions to add that if it's not there already and to get feedback on it.

This should probably be done in both the Transactions table and the Resources Table. The Resources table should record the cost/price (whichever we decide to go with) of each Resource, and the Transaction table should record the ultimate cost of the accepted trade.

gonuke:
It occurred to me this morning that this PR may be suffering from scope creep.  The original intent was simply to implement the ability to fetch data between different agents.  I wonder if we should have a PR that simple does that, even without any code injection - just the fundamental capability.  We can then have additional PR's that add incremental pieces.

dean-krueger:
This has now been split into two separate pull requests (#1850 and #1853) and can probably now be closed (unless there's some good reason to keep it around until those go through). I have all the files locally still in my cep-30 branch so I think that should be safe though.

PR #1856, opened 2025-05-30
Fix cycstub caps

### LINKED ISSUE #1854: `cycstub` generates inconsistent files.
# Describe the bug.
The `cycstub` utility is designed to populate a directory with all of the necessary files and subdirectories to build a new archetype.  However, there appears to be an inconsistency in capitalization of naming.  

# Steps to reproduce the bug.

If one requests a new archetype as follows:

``` bash
cycstub --type facility :TutorialFacility:TutorialFacility
```

it will generate files that will install `libTutorialFacility.so` but it will generate a sample input file that refers to library `libtutorialfacility.so`.

If one requests a new archetype as:

``` bash
cycstub --type facility :tutorialfacility:TutorialFacility
```

it will generate files that will install `libtutorialfacility.so` and also a sample input that refers to library `libtutorialfacility.so`.

The expected behavior is that the example input file is always consistent with the newly installed module name.

I think we need to confirm the intended behavior, specifically:
* should the library name always be lower case, or
* should the library have user-defined case?

The former requires enforcing a lower case library name in all circumstances.  The latter requires enforcing the correct reference in the input file in all cases.

gonuke:
Fixes #1854 

When generating stub files, `cycstub` was forcing lower case names for the module library in the XML example files.  This should result in CamelCase being supported.


abachma2 (commented):
Does this PR not require any source code changes? I only see changes to example files and the CHANGELOG. Is there anything we should add to documentation for users?

Is the idea to support all lowercase or CamelCase? Do we need any new tests to make sure it's supported?

gonuke:
> Does this PR not require any source code changes? I only see changes to example files and the CHANGELOG. Is there anything we should add to documentation for users?
> 
> Is the idea to support all lowercase or CamelCase? Do we need any new tests to make sure it's supported?

Correct - no source code changes. `cycstub` is smart about replacing keywords in the stub files depending on whether the keyword itself is upper case, lower case, CamelCase, etc.  This fix only changes the stub files to use the CamelCase keyword instead of the lowercase keyword so that the correct replacement happens.  

If the user specifies a lowercase library name, the lowercase and CamelCase replacements will be the same as each other and everything will work.  If the user specifies a CamelCasel library name, the lowercase and CamelCase will be different and the correct substitution will now happen.

RE Testing: I guess I could update `test_stubs.py`.  It looks like it only every tested with pure lowercase library names.

gonuke:
I've added a test to `test_stubs.py` and have it parametrically run all the tests for different variations of the library name capitalization

dean-krueger (commented):
I feel like I'm channeling my inner Micah here, but I've noticed that in the `example_facility/inst/region.xml` files you use Pascal Case `StubFacility`, `LibName` (instead of `stubFacility` `libName` which is camel case) and the only reason I point this out is because then it looks like in the tests you actually go and use camel case `stubTest`, which maybe is going to give weird behavior (not sure how picky/smart cycstub is). This comment may be WAY too pedantic, but it may also not be? Feel free to give me a thumbs down and then ignore me if this wasn't helpful!

gonuke:
Hmmm... the tests pass, so that's good.  But maybe we should test both cases?  I think the reality is that one option just doesn't alter the case as all, and that's the option I'm really testing, so we could even use `stUbteSt`

dean-krueger (approved):
Given that tests were added (and all of them pass) as requested by @abachma2, and that my little pedantic request was taken care of, I'd say this is good to go! Unless anyone is opposed, I'll merge tomorrow (to give people some time to see this).

dean-krueger:
Saw that @gonuke made another minor change. This still looks good to me, but before merging I wanted to make sure you weren't working on adding anything else before this goes through.

dean-krueger:
For sure! I'll let you merge once you're done if that works for you

abachma2 (approved):
I didn't get a chance to look at the changes to `tests/test_stubs.py` but everything else looks good and I trust that @dean-krueger did a thorough review. Feel free to merge.

PR #1858, opened 2025-05-30
Fix rng print

### LINKED ISSUE #1857: RNG Schema always printed
# Describe the bug.
Whether or not the command-line option was selected, the RNG schema is always written to the screen.

# Steps to reproduce the bug.
Run any Cyclus problem with a valid input file and the RNG schema is written to screen immediately following the Cyclus ASCII art.

The RNG Schema should only be printed if `--rng-schema` is given on the command-line.

gonuke:
Fixes #1857 

During testing for cycstub, I noticed that the RNG Schema is always printed when I run a problem.  I traced it to an uninitialized local variable that is now initialized to `false`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -56,7 +56,7 @@ Since last release
 * Set locale when writing/reading from serialized SQLite data (#1837)
 * Support mamba v2 CLI in Dockerfile (#1839)
 * Allow master schema be to reported with all discoverable archetypes (#1846)
-* Ensure that the RNG Schema is only printed if requested ()
+* Ensure that the RNG Schema is only printed if requested (#1858)
 
 v1.6.0
 ====================


abachma2 (approved):
This seems pretty straight forward to me. Is there a way that a user can change this to True if they wanted?

gonuke:
> This seems pretty straight forward to me. Is there a way that a user can change this to True if they wanted?

If a user invokes cyclus with `--rng-print` it will print the schema.  This was always there, and invoking this option set the variable to true, but assumed it was otherwise false.  C++ std does not require it to be initialized so it might be random whether or not it was initialized to true/false.

abachma2:
Sounds good to me. Do we need a new test for when the schema gets printed?

gonuke:
We have never historically tested the semantics of the CLI.  Perhaps that can be a different issue/PR?

dean-krueger (approved):
Seems pretty cut and dry like @abachma2 was saying. I'll go ahead and merge.

PR #1861, opened 2025-06-10
Adds a new command-line option to generate a new file w/ up-to-date grammar

### LINKED ISSUE #1847: Create "default" blank input file which includes correct schema grammar
This issue stems from changes in PR #1846 which fixed Cyclus schema grammar. It was suggested that we might want some sort of default blank schema that users can generate which will include the proper hooks for including that grammar and allow the VS Code XML extension (or similar for other types of input files) to auto-populate and suggest correct syntax for (especially new) users. We'll want to do a few things:

- [ ] generate this top-level schema on demand to accommodate changes in archetype availability
- [ ] create an empty XML file that refers to that schema.
- [ ] Check whether a similar thing can be done for python (or other) types of input file

Maybe these are one step - `cyclus -new-file` will update the scheme saving it with some autogenerated unique name and then make an empty XML file that references that schema???

### LINKED ISSUE #1860: [Feature] Generate a new input file with correct grammar reference
Given the possibility of using XML editors like VS Code for guided development of input files, it is useful to be able to generate an empty input file that references an up-to-date schema from a particular host as its grammar.

A new command line option, perhaps `--new-file`, will first generate a copy of the local schema with a filename that includes the hostname and a timestamp, and then generate an empty XML file that officially registers that schema as its grammar.

Add a new command-line option to the cyclus CLI

gonuke:
Closes #1847

Also takes the opportunity to group the command-line options into logical groupings for easier user parsing/searching

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,7 @@ Since last release
 * Added example of header code injection for facility cost (#1829)
 * Table for recording active/dormant cycle periods (#1830)
 * Support for OpenMP parallelized simulations (#1709)
+* Command-line option to make a new blank input file referencing a current schema (#1861)
 
 **Changed:**
 
--- a/cli/cyclus.cc
+++ b/cli/cyclus.cc
@@ -244,9 +244,7 @@ int ParseCliArgs(ArgInfo* ai, int argc, char* argv[]) {
   general.add_options()
       ("help,h", "produce help message")
       ("version,V", "print cyclus core and dependency versions and quit")
-  #if CYCLUS_IS_PARALLEL
-      ("nthreads,j", po::value<int>(), "number of threads to use")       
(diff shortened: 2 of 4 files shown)


dean-krueger (commented):
Did a close read of the code and nothing jumped out at me (besides some of the shorter variable names like vm being a bit confusing at first), but when I went to test it out I couldn't for the life of me get it running. My guess is that I'm doing it wrong, but could you tell me the correct syntax for running this? Once I see that it works I'll gladly approve!

dean-krueger:
Also potentially linked to #1847

dean-krueger:
> My guess is that I'm being silly

Turns out I'm better at guessing than I am at remembering to build code before I try to run it. Got it to work, thanks for the hint! This PR looks good to me (and more importantly, works well on my machine)!

PR #1866, opened 2025-06-15
matl_buy/sell_policy Code injection

### LINKED ISSUE #1865: [Task] Code Inject matl_buy/sell_policy
# What is this task about?

Leftover from the work @nuclearkatie did, there are a lot of mail_buy/sell_policy pragma variables that currently need to be added manually to Agnets which wish to use them. This is antithetical to the newer direction we're trying to go with code injection, and so it would be nice if some injection files were made for those policies.

# Why is this task needed?

Adding pragma vars manually is tedious, error prone, and massively bloats the header files of agents. Code injection will standardize this process, reducing errors and clean up the headers a lot.

cyclus/cycamore#638

Making a `matl_buy_policy.cycpp.h` and `matl_sell_policy.cycpp.h` file which encompass the features of these two bits of work.

dean-krueger:
# Summary of Changes

This PR aims to capture the required pragma vars, functions, and various other bits and pieces required to use the matl_buy_policy and matl_sell_policy @nuclearkatie added as part of her dissertation in an injectable form to make it easier to add to agents, and to clean up bloat where it was added manually.

# Related CEPs and Issues

This PR is related to:

- cyclus/cycamore#638
- Closes #1865 

# Associated Developers

@nuclearkatie 
@bam241 

# Design Notes

The code injection follows the guidelines set out by @bam241 in his changes to position, and the work I did adding facility/institution/region_costs for EconomicEntities to use. 

# Testing and Validation

The code was built locally on my machine and the cycamore_unit_tests were run and all passed. I then commented out the injected parts and immediately the code wouldn't build anymore, giving me some faith that this worked, though I did not add any extra tests.

# Checklist

 - [ ]  Add or update tests.
 - [ ]  Document if needed.
 - [x]  Update the changelog.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@
 build
 .DS_Store
 debug
+core
+packages.xml
 doxygen.conf
 platform.h
 *.pdf
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,8 @@ Since last release
 
 **Added:**
 * Added code injection for matl_buy/sell_policy (#1866)
+* Added set of basic finance math function to EconomicEntity (#1864)
+* Added Composition functions to get printable material composition information (#1868)
 * Added code injection variables for economic data (#1853)
 * Agent now inherits EconomicEntity to allow limited communication across RIF (#1850)
 * Added TransportUnits (#1750, #1772, #1784)
--- a/src/composition.cc
+++ b/src/composition.cc
@@ -116,6 +116,17 @@ Composition::Composition(int prev_decay, ChainPtr decay_line)
   next_id_++;
 }
 
+std::string Composition::ToString(CompMap v) {
+  std::string comp = "";
+  for (cyclus::CompMap::const_iterator it = v.begin();
+       it != v.end(); ++it) {
+    comp += std::to_string(it->first) + std::string(": ") + 
+            std::to_string(it->second) + std::string("\n");
+  }
+
+  return comp;
+}
+
 Composition::Ptr Composition::NewDecay(int delta, uint64_t secs_per_timestep) {
   int tot_decay = prev_decay_ + delta;
   atom();  // force evaluation of atom-composition if not calculated already
--- a/src/composition.h
+++ b/src/composition.h
@@ -77,6 +77,11 @@ class Composition {
   /// not done previously).
   void Record(Context* ctx);
 
+  /// @brief Transforms a composition into a printable string, primarily for
+  /// debugging and logging.
+  /// @return A String represented as a CompMap
+  static std::string ToString(CompMap v);
+
  protected:
   /// a chain containing compositions that are a result of decay from a common
   /// ancestor composition. The key is the total amount of time a composition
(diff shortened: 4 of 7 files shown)


dean-krueger:
Immediately, once again, realized I hadn't updated the CHANGELOG. Commit incoming.

gonuke:
Looking at this reminds me that we should take this opportunity to reconsider the design.  This grew piecemeal into this set of parameters and it's a little unwieldy.  I wonder if there are ways to make it a little more streamlined and/or simpler.

gonuke:
Some thoughts....
* could we have a method that initializes a "time period" and use it once for active and once for dormant?  Maybe that's just cumbersome?
* could we break this into multiple snippets to be injected? Maybe not because of how they interact, but worth considering.
* could some of the initialization methods be moved to the buy/sell policy itself and called from here?

dean-krueger:
I find myself looking at this code and having a few thoughts:

1. What I WANT to do is to make some sort of struct that describes a policy and have all the properties that are common between the "buying", "active_buying", and "dormant_buying" versions (frequency_type, val, min, max, mean, stddev, end_probability, disruption_probability, disruption_length). That would probably be able to make this file ~60% shorter, but I don't think the cyclus pragma supports that sort of thing (we support std::pairs and combinations of things, but it's all a little finicky. Not sure how that would work with the way we do Cyclus input, but I do notice that there is a lot of "basically the same thing but with a different prefix" variables in here. **Would love to hear your thoughts on this before I dive in and spend a bunch of time messing around.** 
2. I'm not entirely sure what you're getting at with the whole time-period thing, but maybe if you elaborate I could look into that?
3. As far as the stuff that's currently in Storage's EnterNotify() function (which I think is what you're talking about with the initialization stuff), that's almost entirely Storage-specific code which dictates how that particular facility behaves. The part that has to do with the Buy/Sell policy is actually in a pretty concise form already.

As far as separating this file into multiple files, I feel like that's probably the wrong approach if we can condense it in some way like I was describing earlier, but could be possible if there's a reasonable use case for having an active part of your buy policy without a dormant part? I don't think that makes a ton of sense, but maybe @nuclearkatie has different thoughts?

gonuke:
RE: "time period"... the set of possible inputs to define an `active_buying` period and `dormant_buying` period are the same.  I was wondering if there were a way to reuse code across those two, but not sure if the pragma and post-processing actually provide that much flexibility.

gonuke:
Re: initialization - I was wondering whether some of the logic of `InitBuyPolicyParameters()` could be part of a method that was a member of the buy policy itself, and called by passing in the pragma variables that are being injected.  Something like:
```
IntDistribution::Ptr MatBuyPolicy::InitializeIntDist(std::string name, min, max, frequency_type, mean, stddev, ....)
```

And then from `InitBuyPolicyParameters()` calling:
```
active_dist_ = InitializeIntDist("active", active_buying_min, active_buying_max, etc...)
dormat_dist_ = InitializeIntDist("dormant", dormat_buying_min, dormant_buying_max, etc...)
```

gonuke:
RE: more advanced data structure - Sadly, while I share your impulse, I don't think anything is really easily accessible and I'm not sure it's worth the effort to develop something new here.  There is a lot of backend infrastructure to deal with more complicated data types.

PR #1868, opened 2025-06-16
Add Functions to Material to get Printable String Versions of a Material

### LINKED ISSUE #1739: No way to print out the full composition of a material to screen
There are some times where it might be nice from a user perspective to be able to get a string with the normalized composition of a material in it (to print, or to use otherwise). Currently, Cyclus can write this out to a file, but there's no way to do it on screen or use it as a variable. I wrote my own version of a function that does this, which I will include below to further illustrate what I mean.

```
std::string Reactor::GetComp(cyclus::Material::Ptr mat) {
  std::string comp = "{";
  cyclus::CompMap c = mat->comp()->atom();
  cyclus::compmath::Normalize(&c, 1);
  for (std::map<const int, double>::const_iterator it = c.begin();
       it != c.end(); ++it) {
    comp = comp + std::string("{") + std::to_string(it->first) +
           std::string(",") + std::to_string(it->second) + std::string("},");
  }
  comp.pop_back();
  comp = comp + std::string("}");
  return comp;
}
```

dean-krueger:
# Summary of Changes

Two functions were added to Material, each returns a string with a breakdown of the material's composition (one in atom and one in mass basis). This was a function I really wanted when I was first learning Cyclus because there was, so far as I could tell at the time, no easy way to do this to check that I was doing things correctly.

# Related CEPs and Issues

This PR is related to:

- Closes #1739 

# Associated Developers

None.

# Design Notes

# Testing and Validation

The code was built locally on my machine, and all Cyclus tests passed after adding two new tests: one for each function.

# Checklist

 - [x]  Add or update tests.
 - [ ]  Document if needed.
 - [x]  Update the changelog.

### DIFF
--- a/src/composition.cc
+++ b/src/composition.cc
@@ -116,6 +116,18 @@ Composition::Composition(int prev_decay, ChainPtr decay_line)
   next_id_++;
 }
 
+std::string Composition::ToString(CompMap v) {
+  std::string comp = "{";
+  for (cyclus::CompMap::const_iterator it = v.begin();
+       it != v.end(); ++it) {
+    comp += std::string("{") + std::to_string(it->first) +
+            std::string(",") + std::to_string(it->second) + std::string("},");
+  }
+  comp.pop_back();
+  comp = comp + std::string("}");
+  return comp;
+}
+
 Composition::Ptr Composition::NewDecay(int delta, uint64_t secs_per_timestep) {
   int tot_decay = prev_decay_ + delta;
   atom();  // force evaluation of atom-composition if not calculated already
--- a/src/composition.cc
+++ b/src/composition.cc
@@ -117,14 +117,13 @@ Composition::Composition(int prev_decay, ChainPtr decay_line)
 }
 
 std::string Composition::ToString(CompMap v) {
-  std::string comp = "{";
+  std::string comp = "";
   for (cyclus::CompMap::const_iterator it = v.begin();
        it != v.end(); ++it) {
-    comp += std::string("{") + std::to_string(it->first) +
-            std::string(",") + std::to_string(it->second) + std::string("},");
+    comp += std::to_string(it->first) + std::string(": ") + 
+            std::to_string(it->second) + std::string("\n");
   }
-  comp.pop_back();
-  comp = comp + std::string("}");
+
   return comp;
 }
 
--- a/src/material.cc
+++ b/src/material.cc
@@ -53,34 +53,6 @@ void Material::Record(Context* ctx) const {
   comp_->Record(ctx);
 }
 
-std::string Material::GetNormalizedCompAtom() const {
-  std::string comp = "{";
-  cyclus::CompMap c = comp_->atom();
-  cyclus::compmath::Normalize(&c, 1);
-  for (std::map<const int, double>::const_iterator it = c.begin();
-       it != c.end(); ++it) {
-    comp = comp + std::string("{") + std::to_string(it->first) +
-           std::string(",") + std::to_string(it->second) + std::string("},");
-  }
-  comp.pop_back();
-  comp = comp + std::string("}");
-  return comp;
-}
-
-std::string Material::GetNormalizedCompMass() const {
-  std::string comp = "{";
-  cyclus::CompMap c = comp_->mass();
-  cyclus::compmath::Normalize(&c, 1);
-  for (std::map<const int, double>::const_iterator it = c.begin();
-       it != c.end(); ++it) {
-    comp = comp + std::string("{") + std::to_string(it->first) +
-           std::string(",") + std::to_string(it->second) + std::string("},");
-  }
-  comp.pop_back();
-  comp = comp + std::string("}");
-  return comp;
-}
-
 std::string Material::units() const {
   return "kg";
 }
--- a/src/material.cc
+++ b/src/material.cc
@@ -53,6 +53,34 @@ void Material::Record(Context* ctx) const {
   comp_->Record(ctx);
 }
 
+std::string Material::GetNormalizedCompAtom() const {
+  std::string comp = "{";
+  cyclus::CompMap c = comp_->atom();
+  cyclus::compmath::Normalize(&c, 1);
+  for (std::map<const int, double>::const_iterator it = c.begin();
+       it != c.end(); ++it) {
+    comp = comp + std::string("{") + std::to_string(it->first) +
+           std::string(",") + std::to_string(it->second) + std::string("},");
+  }
+  comp.pop_back();
+  comp = comp + std::string("}");
+  return comp;
+}
+
+std::string Material::GetNormalizedCompMass() const {
+  std::string comp = "{";
+  cyclus::CompMap c = comp_->mass();
+  cyclus::compmath::Normalize(&c, 1);
+  for (std::map<const int, double>::const_iterator it = c.begin();
+       it != c.end(); ++it) {
+    comp = comp + std::string("{") + std::to_string(it->first) +
+           std::string(",") + std::to_string(it->second) + std::string("},");
+  }
+  comp.pop_back();
+  comp = comp + std::string("}");
+  return comp;
+}
+
 std::string Material::units() const {
   return "kg";
 }
--- a/tests/material_tests.cc
+++ b/tests/material_tests.cc
@@ -443,17 +443,17 @@ TEST_F(MaterialTest, DecayHeatTest) {
 TEST_F(MaterialTest, GetNormalizedCompAtom) {
   double val = 1.5 * units::kg;
   Material::Ptr m1 = Material::CreateUntracked(val, diff_comp_);
-  cyclus::CompMap c = m1->comp()->atom();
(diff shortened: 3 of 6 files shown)


src/material.cc
```diff
@@ -53,6 +53,34 @@ void Material::Record(Context* ctx) const {
   comp_->Record(ctx);
 }
 
+std::string Material::GetNormalizedCompAtom() const {
```
gonuke, line 56: The bulk of this implementation should probably for a `Composition` that gets called from this `Material` method.

A method like this is commonly named `to_string()`, but maybe in this case `to_atom_string()` or `to_string(atom|mass)`?

src/material.cc
```diff
@@ -53,6 +53,34 @@ void Material::Record(Context* ctx) const {
...
+  cyclus::CompMap c = comp_->atom();
+  cyclus::compmath::Normalize(&c, 1);
+  for (std::map<const int, double>::const_iterator it = c.begin();
+       it != c.end(); ++it) {
+    comp = comp + std::string("{") + std::to_string(it->first) +
+           std::string(",") + std::to_string(it->second) + std::string("},");
+  }
+  comp.pop_back();
+  comp = comp + std::string("}");
+  return comp;
```
gonuke, line 67: Most of the code for the atom and mass versions are identical, with just a different `CompMap`, so probably want to combine those into a single method that just gets called with a different `CompMap`.

src/material.cc
```diff
@@ -53,6 +53,34 @@ void Material::Record(Context* ctx) const {
   comp_->Record(ctx);
 }
 
+std::string Material::GetNormalizedCompAtom() const {
+  std::string comp = "{";
+  cyclus::CompMap c = comp_->atom();
+  cyclus::compmath::Normalize(&c, 1);
+  for (std::map<const int, double>::const_iterator it = c.begin();
+       it != c.end(); ++it) {
+    comp = comp + std::string("{") + std::to_string(it->first) +
```
gonuke, line 62:
```suggestion
    comp += std::string("{") + std::to_string(it->first) +
```

src/material.cc
```diff
@@ -53,6 +53,34 @@ void Material::Record(Context* ctx) const {
   comp_->Record(ctx);
 }
 
+std::string Material::GetNormalizedCompAtom() const {
+  std::string comp = "{";
+  cyclus::CompMap c = comp_->atom();
+  cyclus::compmath::Normalize(&c, 1);
+  for (std::map<const int, double>::const_iterator it = c.begin();
```
gonuke, line 60:
```suggestion
  for (cyclus::CompMap::const_iterator it = c.begin();
```

dean-krueger:
Good suggestions, I was definitely shooting from the hip when making these functions (especially when naming them). Before I go and fix them up, though, your comment about where I'd want to use them suggests that there's maybe a slightly larger discussion to be had here about how 'necessary' these are, which I can add to the agenda for our fuel cycles meeting tomorrow so we can avoid the back and forth. After that, if we decide to go forward, I can run through with a round of clean-up (or hit the delete key a bunch).

gonuke:
I ask about where to use them because Cyclus is not designed to have much output to screen, and have it all go to the DB files.  To be sure, the log can be used for different levels of debugging and output, so it could be useful there.   Also, historically, we are used to compositions either being very simple (fresh fuel), or too complex to want to view this way (spent fuel).  None of these things mean we should NOT include these functions, and I'm sure someone will find them valuable in the future.

I think you can go ahead and make some changes for another round of review.

src/composition.cc
```diff
@@ -116,6 +116,18 @@ Composition::Composition(int prev_decay, ChainPtr decay_line)
   next_id_++;
 }
 
+std::string Composition::ToString(CompMap v) {
```
gonuke, line 119: Most standard libraries for this kind of function name this `to_string()` - not sure how to reconcile that with our style practices

src/composition.cc
```diff
@@ -116,6 +116,18 @@ Composition::Composition(int prev_decay, ChainPtr decay_line)
...
+  std::string comp = "{";
```
gonuke, line 120: Slight readability improvements of the final product, without bloating it too much.  Adding a space and using a different delimiter than each of the nuclide entries
```suggestion
  std::string comp = "[ ";
```

src/composition.cc
```diff
@@ -116,6 +116,18 @@ Composition::Composition(int prev_decay, ChainPtr decay_line)
...
+  for (cyclus::CompMap::const_iterator it = v.begin();
+       it != v.end(); ++it) {
+    comp += std::string("{") + std::to_string(it->first) +
+            std::string(",") + std::to_string(it->second) + std::string("},");
+  }
+  comp.pop_back();
+  comp = comp + std::string("}");
```
gonuke, line 127: readability
```suggestion
  comp = comp + std::string(" ]");
```

src/composition.cc
```diff
@@ -116,6 +116,18 @@ Composition::Composition(int prev_decay, ChainPtr decay_line)
   next_id_++;
 }
 
+std::string Composition::ToString(CompMap v) {
+  std::string comp = "{";
+  for (cyclus::CompMap::const_iterator it = v.begin();
+       it != v.end(); ++it) {
+    comp += std::string("{") + std::to_string(it->first) +
+            std::string(",") + std::to_string(it->second) + std::string("},");
```
gonuke, line 124: Some space and a different delimiter for readability
```suggestion
            std::string(" : ") + std::to_string(it->second) + std::string("},");
```

tests/material_tests.cc
```diff
@@ -440,4 +440,20 @@ TEST_F(MaterialTest, DecayHeatTest) {
   ASSERT_NEAR(3.614E-14 , dec_heat, 0.0005);
 }
 
+TEST_F(MaterialTest, GetNormalizedCompAtom) {
+  double val = 1.5 * units::kg;
+  Material::Ptr m1 = Material::CreateUntracked(val, diff_comp_);
+  cyclus::CompMap c = m1->comp()->atom();
+  cyclus::compmath::Normalize(&c);
+  EXPECT_EQ(Composition::ToString(c), "{{380890000,0.459861},{822080000,0.196585},{922350000,0.173947},{952410000,0.169608}}");
```
gonuke, line 448: Update for above readability suggestions
```suggestion
  EXPECT_EQ(Composition::ToString(c), "[ {380890000 : 0.459861},{822080000 : 0.196585},{922350000 : 0.173947},{952410000 : 0.169608} ]");
```

tests/material_tests.cc
```diff
@@ -440,4 +440,20 @@ TEST_F(MaterialTest, DecayHeatTest) {
...
+}
+
+TEST_F(MaterialTest, GetNormalizedCompMass) {
+  double val = 1.5 * units::kg;
+  Material::Ptr m1 = Material::CreateUntracked(val, diff_comp_);
+  cyclus::CompMap c = m1->comp()->mass();
+  cyclus::compmath::Normalize(&c);
+  EXPECT_EQ(Composition::ToString(c), "{{380890000,0.250000},{822080000,0.250000},{922350000,0.250000},{952410000,0.250000}}");
```
gonuke, line 456: Update for above readability suggestions
```suggestion
  EXPECT_EQ(Composition::ToString(c), "[ {380890000 : 0.250000},{822080000 : 0.250000},{922350000 : 0.250000},{952410000 : 0.250000}] ");
```

tests/material_tests.cc
```diff
@@ -440,4 +440,20 @@ TEST_F(MaterialTest, DecayHeatTest) {
   ASSERT_NEAR(3.614E-14 , dec_heat, 0.0005);
 }
 
+TEST_F(MaterialTest, GetNormalizedCompAtom) {
+  double val = 1.5 * units::kg;
+  Material::Ptr m1 = Material::CreateUntracked(val, diff_comp_);
+  cyclus::CompMap c = m1->comp()->atom();
+  cyclus::compmath::Normalize(&c);
```
gonuke, line 447: I think (?) the original pattern will result in the composition for material `m1` being changed because `atom()` returns a reference to that composition and `Normalize()` changes the values in place.  Maybe this is more `const` w.r.t. `m1`? 
```suggestion
  cyclus::CompMap c(m1->comp()->atom());
  cyclus::compmath::Normalize(&c);
```
It doesn't matter for testing purposes, but it may be useful to demonstrate a `const` pattern

gonuke (changes requested):
Thanks @dean-krueger.  These are mostly readability suggestions for the output - it might also be helpful to add spaces around the commas between nuclides (?)

src/composition.cc
dean-krueger: The consensus of the wise and all knowing Chat was:

- to_string() is better for free functions which are attempting to conform to something called the STL interface
- ToString() is better for internal functions assuming the rest of your code uses Pascal Case

Since this is an internal, and not a free function, it says to stick to Pascal Case to match the rest of the code. Not sure if this is a satisfying answer, happy to do it the other way if you still think it's better.

PR #1853, opened 2025-05-28
Add financial data code injection for CYCLUS agents

dean-krueger:
# Summary of Changes

This pull request is the second half of splitting #1844 (The other half being #1850) and deals with adding infrastructure for injecting financial information into the headers of CYCLUS (+Cycamore +others) agents. It comes with three main files, called "facility/institution/region_cost.cycpp", which include all the required code for injection, as well as an example of injection in the null region agent (.cc and .h files). These changes address and close #1851. Finally, there is a change made to the trade_executor.h file which adds a line to the Transactions table showing the Cost (1/pref) of the successful bid associated with that transaction. This closes #1852.

# Related CEPs and Issues

- Cyclus CEP-30 (cyclus/cyclus.github.com#396
- Cyclus Issue #1851 
- Cyclus Issue #1852 
- Cyclus PR #1844 
- Cyclus PR #1850 

# Associated Developers

@gonuke 

# Design Notes

There was a decent amount of thought that went into the design of this code, but I have a feeling that a few more tweaks will probably be requested before this gets merged.

# Testing and Validation

I have tested this on my local machine (alongside the changes in #1850 ) and it works there. However, like in #1850 there are currently **no** official tests associated with this code. Those will need to be added before this is merged.

# Checklist

 - [x] Read the [Contributing to Cyclus](https://fuelcycle.org/kernel/contributing_to_cyclus.html) guide.
 - [x]  Compile and run locally.
 - [ ]  Add or update tests.
 - [ ]  Document if needed.
 - [x]  Follow style guidelines.
 - [x]  Update the changelog.
 - [ ]  Address all review comments.
Reviewers, please refer to the Cyclus [Guide for Reviewers](https://fuelcycle.org/kernel/pr_review.html).

PR #1872, opened 2025-06-25
Position Code Injection Variable Initialization Changes

### LINKED ISSUE #1871: [Task] Rework position.cycpp.h to initialize variables more elegantly
# What is this task about?

@gonuke pointed out in cyclus/cycamore#502 that the way @bam241 was initializing things perhaps wasn't in line with the way I had done so in #1853. This issue is to take a look at the code injection and perhaps rework it so that it's more in line with the way things were done in #1853 for consistency.

# Why is this task needed?

Having a consistent way to initialize variables in your code base is good. 

#1510

This issue can be closed once everyone (namely @gonuke and @bam241 ) are happy with the state of the position code injection implementation. Nothing necessarily needs to change, but it may require changes to `position.h`, `position.cc`, or `position.cycpp.h`.

dean-krueger:
# Summary of Changes
This PR primarily serves to reopen discussion on the "correct" way to do code injection through the .cycpp system, specifically related to the initialization of variables in position. A few minor changes were made to the way @bam241 did things, mostly to bring the code in line with some of the in-person talks @gonuke and I had when talking about cyclus/cycamore#502.

# Related CEPs and Issues

This PR is related to:

- cyclus/cycamore#502
- #1853 
- #1510 
- Closes #1871   

# Associated Developers

@bam241 
@gonuke 

# Design Notes

I attempted to remove all initialization from the constructor (the `latitude(0.0), longitude(0.0), coordinates(0, 0)` bit), but because of limitations with the `#pragma cyclus var` system, it is impossible to inject something like `double latitude = 0.0` (it throws an error). I was able to remove the `coordinates(0, 0)` bit, and to condense the initialization into an injected function, however. 

## TODO:
- [x] Change the documentation in the .cycpp.h file
- [x] Update the changelog (oops)

# Testing and Validation

The code was built locally on my machine. No additional tests were added, but all existing tests passed on my machine. 

# Checklist

 - [x]  Add or update tests.
 - [x]  Document if needed.
 - [x]  Update the changelog.

### DIFF
--- a/src/toolkit/position.cc
+++ b/src/toolkit/position.cc
@@ -28,6 +28,11 @@ double Position::longitude() const {
 void Position::latitude(double lat) {
   if (!ValidLatitude(lat)) {
     latitude_ = kInvalidLatitude;
+    std::stringstream msg;
+    msg << "The provided latitude (" << lat
+        << ") is outside the acceptable range "
+        << "[-90, 90]. Setting to " << kInvalidLatitude;
+    cyclus::Warn<cyclus::VALUE_WARNING>(msg.str());
   } else { 
     latitude_ = SetPrecision(lat * CYCLUS_DECIMAL_SECOND_MULTIPLIER, 1);
   }
@@ -36,6 +41,11 @@ void Position::latitude(double lat) {
 void Position::longitude(double lon) {
   if (!ValidLongitude(lon)) {
     longitude_ = kInvalidLongitude;
+    std::stringstream msg;
+    msg << "The provided longitude (" << lon
+        << ") is outside the acceptable range "
+        << "[-180, 180]. Setting to " << kInvalidLongitude;
+    cyclus::Warn<cyclus::VALUE_WARNING>(msg.str());
   } else {
     longitude_ = SetPrecision(lon * CYCLUS_DECIMAL_SECOND_MULTIPLIER, 1);
   }
@@ -58,11 +68,6 @@ void Position::RecordPosition(Agent* agent) {
 }
 bool Position::ValidLatitude(double lat) {
   if (lat > 90 || lat < -90) {
-    std::stringstream msg;
-    msg << "The provided latitude (" << lat
-        << ") is outside the acceptable range "
-        << "[-90, 90]. Setting to NaN.";
-    cyclus::Warn<cyclus::VALUE_WARNING>(msg.str());
     return false;
   }
   return true;
@@ -70,11 +75,6 @@ bool Position::ValidLatitude(double lat) {
 
 bool Position::ValidLongitude(double lon) {
   if (lon > 180 || lon < -180) {
-    std::stringstream msg;
-    msg << "The provided longitude (" << lon
-        << ") is outside the acceptable range "
-        << "[-180, 180]. Setting to NaN.";
-    cyclus::Warn<cyclus::VALUE_WARNING>(msg.str());
     return false;
   }
   return true;
--- a/src/toolkit/position.cc
+++ b/src/toolkit/position.cc
@@ -60,7 +60,7 @@ void Position::LatCheck(double lat) {
     msg << "The provided latitude (" << lat
         << ") is outside the acceptable range. "
         << "[-90, 90]";
-      cyclus::Warn<cyclus::VALUE_WARNING>(msg.str());
+    cyclus::Warn<cyclus::VALUE_WARNING>(msg.str());
   }
 }
 
--- a/src/toolkit/position.cc
+++ b/src/toolkit/position.cc
@@ -44,7 +44,7 @@ void Position::longitude(double lon) {
     std::stringstream msg;
     msg << "The provided longitude (" << lon
         << ") is outside the acceptable range "
-        << "[-180, 180]. Setting to " << kInvalidLongitude;
+        << "[-180, 180]. Longitude has been set to " << longitude_;
     cyclus::Warn<cyclus::VALUE_WARNING>(msg.str());
   } else {
     longitude_ = SetPrecision(lon * CYCLUS_DECIMAL_SECOND_MULTIPLIER, 1);
--- a/src/toolkit/position.cc
+++ b/src/toolkit/position.cc
@@ -11,10 +11,14 @@ namespace toolkit {
 Position::Position() : latitude_(0), longitude_(0) {}
 
 Position::Position(double decimal_lat, double decimal_lon) {
-  LatCheck(decimal_lat);
-  LonCheck(decimal_lon);
-  latitude_ = SetPrecision(decimal_lat * CYCLUS_DECIMAL_SECOND_MULTIPLIER, 1);
-  longitude_ = SetPrecision(decimal_lon * CYCLUS_DECIMAL_SECOND_MULTIPLIER, 1);
+  if (ValidLatitude(decimal_lat) and ValidLongitude(decimal_lon)) {
+    latitude_ = SetPrecision(decimal_lat * CYCLUS_DECIMAL_SECOND_MULTIPLIER, 1);
+    longitude_ = SetPrecision(decimal_lon * CYCLUS_DECIMAL_SECOND_MULTIPLIER, 1);
+  } else {
+    // Set to some legal value. 0.0 is chosen as the default.
+    latitude_ = 0.0; 
+    longitude_ = 0.0;
+  }
 }
 
 Position::~Position() {}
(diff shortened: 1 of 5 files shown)


src/toolkit/position.cycpp.h
```diff
@@ -28,5 +25,14 @@ std::vector<int> cycpp_shape_latitude = 0;
        "be expressed in degrees as a double." \
 }
 double longitude;
+
+protected:
+
+  void InitializePosition(cyclus::Agent* agent) {
```
gonuke, line 31: This injected function becomes a member function of the agent, so we probably don't need to pass in an agent pointer an can just assume `this` and pass that to `RecordPosition`

gonuke (changes requested):
This change looks like the right way to go - thanks @dean-krueger.  I only have one small comment here - maybe more on the Cycamore side

src/toolkit/position.cycpp.h
```diff
@@ -9,7 +8,7 @@
 //   longitude);'
 //   - call the record method: 'coordinates.RecordPosititon(this);'
```
bam241, line 9: Maybe add some documentation there about the InitializePosition.

my guess is the InitializePosition would need to be called in the archetype.

src/toolkit/position.cycpp.h
```diff
@@ -28,5 +25,14 @@ std::vector<int> cycpp_shape_latitude = 0;
        "be expressed in degrees as a double." \
 }
 double longitude;
+
+protected:
```
bam241, line 29: Shall we also enforce the visibility of the `coordinates`, `latitude ` and `longitude` variable ?

i.e. if `InitializePosition` is `protected`, shall not `coordinates` to also be `private` ?
What about `lat` and `long` ?

bam241 (approved):
LGTM!

some quick comments about docs and exposure of the injected variables in the "parasited" archetype.

src/toolkit/position.cycpp.h
bam241: In my opinion she shall define the injection as "protected/private/public" in the archetype, as:
```
private:
   // Code Injection:
   #include "toolkit/position.cycpp.h"
```
this would allow the developer of the archetype to control this and not be enforced and potentially have impact on the rest of the archetype members

src/toolkit/position.cycpp.h
dean-krueger: Definitely. I made these changes late at night, and only realized I had forgotten to do this after I made all my pushes before bed. I'll modify the documentation today. Thanks for pointing this out!

src/toolkit/position.cycpp.h
```diff
@@ -1,15 +1,14 @@
-
 // This includes the required header to add geographic coordinates to a
 // archetypes.
-// One only need to:
-// - '#include "toolkit/position.cycpp.h"' in the core of the archetype class (as
+// One only needs to:
+// - '#include "toolkit/position.cycpp.h"' in the header of the archetype class (as
 // private)
```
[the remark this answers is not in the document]
dean-krueger: Doing this on my end since you missed some `//`'s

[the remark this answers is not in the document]
dean-krueger: I also added the word "strongly", since that felt more forceful.

dean-krueger:
Had to change a few of the tests to work with the new way of doing things (changed throws to warnings in `CheckLat` and `CheckLon` functions so that we could pass tests). Let me know if anything jumps out as a "bad idea" or anything like that, but I think this _should_ be good to go now!

tests/toolkit/position_tests.cc
```diff
@@ -1,10 +1,23 @@
 #include <gtest/gtest.h>
 #include <iostream>
 #include <string>
+#include <functional>
 
 #include "toolkit/position.h"
 using cyclus::toolkit::Position;
 
+std::string CaptureCerr(std::function<void()> f) {
```
gonuke, line 9: Did you look for other examples of test that check for warning messages? Is this how they do it, too?

gonuke (changes requested):

tests/toolkit/position_tests.cc
dean-krueger: As you probably guessed, this was a Chat suggestion, but I spent a little bit of time just now digging around to see if it was a hallucination or not, and it seems like the pattern is fairly common. See [this](https://stackoverflow.com/questions/3803465/how-to-capture-stdout-stderr-with-googletest) stack overflow post where I'm sure chat got its "idea" from (it's got a very similar style). The main thing here is that since the function no longer throws, we can't use the GTEST EXPECT_THROW(), but we still need some way to catch that the warning got produced. The way to do this seems to be to capture the cerr output like this, and then the "extra" bit is that you need to be careful to restore the cerr buffer (not sure if that's the right word for it) or else it will get confused and the error will compound later down the line. This is just one way of doing that, but it seems pretty standard to me (not sure if there's truly a "right" way to do this, and GTEST doesn't offer any advice that I could find).

src/toolkit/position.cc
```diff
@@ -70,7 +70,7 @@ void Position::LonCheck(double lon) {
     msg << "The provided longitude (" << lon
         << ") is outside the acceptable range."
         << "[-180, 180]";
-    throw ValueError(msg.str());
```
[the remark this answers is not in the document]
dean-krueger: From google:

> Exceptions are events that occur during the execution of a program, indicating an unexpected or exceptional  condition that prevents normal program flow. If an exception is thrown and not caught by a suitable catch block, it will typically lead to program termination.

> Warnings are messages issued by the compiler during the compilation phase. They highlight potential problems, non-standard practices, or areas that might lead to runtime issues, even if the code is syntactically valid. Warnings typically require the developer to review and potentially modify the code to address the underlying issue or suppress the warning if it's deemed irrelevant. Examples include unused variables, implicit type conversions, or potential uninitialized variables.


Seems like warning is the way to go, but happy to talk about this more or change it if you'd prefer.

src/toolkit/position.cc
```diff
@@ -60,7 +60,7 @@ void Position::LatCheck(double lat) {
     msg << "The provided latitude (" << lat
         << ") is outside the acceptable range. "
         << "[-90, 90]";
-    throw ValueError(msg.str());
+      cyclus::Warn<cyclus::VALUE_WARNING>(msg.str());
```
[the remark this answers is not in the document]
dean-krueger: Fixed! Good catch.

src/toolkit/position.cc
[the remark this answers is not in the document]
gonuke: If we warn, we should maybe leave this in a more reasonable state, e.g. change the value of the lat/long?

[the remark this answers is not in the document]
dean-krueger: Yeah, I was thinking about that. I'm sort of stuck between two options, though. On the one hand, we use a default of `0.0` for when a value isn't provided, so that's the one that makes the most sense if we're setting it to some "legal" value, but on the other hand, it may mask bugs by being set to something "legal", so it may make sense to set it to something like `quiet_nan()` (which would still be "valid" for math, but cascade through as nan without raising errors), so it was easy to see that something was going wrong if you looked at your output and weren't paying attention to the terminal while it was running (we print a lot of stuff, sometimes warnings get suppressed with the warn limit, etc). I'm LEANING towards `0.0` because it'd be "easiest", but I can see why `quiet_nan()` might be good too. Interested in your thoughts.

dean-krueger:
So it turns out that `(0.0, 0.0)` is specifically a thing in mapping software that people use to trap errors, since it's in the middle of the ocean. Fun read [here](https://en.wikipedia.org/wiki/Null_Island), but I think this is a good default for us to set things to instead of `quiet_nan()`, since it's clearly not "anywhere".

tests/toolkit/position_tests.cc
dean-krueger: Oh! I thought this had to be done in the input file somehow, so I didn't even consider testing out the warn as error thing. I suppose that's a better way to do it, since that functionality is built in to the warning system and can be used like that. It's also going to be way less verbose. I'll check that out! Good suggestion.

src/toolkit/position.cc
```diff
@@ -28,20 +32,29 @@ double Position::longitude() const {
...
-  LonCheck(lon);
-  longitude_ = SetPrecision(lon * CYCLUS_DECIMAL_SECOND_MULTIPLIER, 1);
+  if (ValidLongitude(lon)) {
+    longitude_ = SetPrecision(lon * CYCLUS_DECIMAL_SECOND_MULTIPLIER, 1);
+  } else {
+    longitude_ = 0.0;
+  }
 }
 
 void Position::set_position(double lat, double lon) {
```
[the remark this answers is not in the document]
dean-krueger: As not the person who developed this, I'm not SUPER sure what the value of the precision part of this is, but I'll take a look at this and see what makes sense and come back with something more sensible (or at least an explanation for why it makes sense to leave it as it is).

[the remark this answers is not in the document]
gonuke: But those two functions use the same precision setting as this one does.  In essence, there are currently 4 functions that all set some combination of lat/long with the same approach.

gonuke (changes requested):
Thanks for cleaning this up @dean-krueger - I've got one last request to move the warning messages to the place where the actually value is being set to the invalid default.  This should provide extra insurance against the message and the action becoming out of sync.

dean-krueger:
Done! Let me know if there's anything else, happy to continue to clean things up if needed

src/toolkit/position.cc
```diff
@@ -28,20 +26,34 @@ double Position::longitude() const {
...
 
 void Position::latitude(double lat) {
-  LatCheck(lat);
-  latitude_ = SetPrecision(lat * CYCLUS_DECIMAL_SECOND_MULTIPLIER, 1);
+  if (!ValidLatitude(lat)) {
+    latitude_ = kInvalidLatitude;
+    std::stringstream msg;
+    msg << "The provided latitude (" << lat
+        << ") is outside the acceptable range "
+        << "[-90, 90]. Setting to " << kInvalidLatitude;
```
gonuke, line 34:
```suggestion
        << "[-90, 90]. Latitude has been set to " << latitude_;
```

src/toolkit/position.cc
```diff
@@ -28,20 +26,34 @@ double Position::longitude() const {
...
 
 void Position::longitude(double lon) {
-  LonCheck(lon);
-  longitude_ = SetPrecision(lon * CYCLUS_DECIMAL_SECOND_MULTIPLIER, 1);
+  if (!ValidLongitude(lon)) {
+    longitude_ = kInvalidLongitude;
+    std::stringstream msg;
+    msg << "The provided longitude (" << lon
+        << ") is outside the acceptable range "
+        << "[-180, 180]. Setting to " << kInvalidLongitude;
```
gonuke, line 47:
```suggestion
        << "[-180, 180]. Longitude has been set to " << longitude_;
```

gonuke (changes requested):
small change to warnings to report the value that it has already been changed to

PR #1874, opened 2025-07-02
Allow XML to have multiple `archetype` elements

gonuke:
The previous XML schema only allowed a single `archetype` block to exist.  This change allows multiple `archetype` blocks, as long as there is at least one.

A motivating use case for this is to facilitate including XML snippets that contain long lists of facility prototypes that may be useful in many simulations.  Without this change, the top level input file must include all the archetype specs for the archetypes that are used in those facilities.  With this change, the archetype specs that are used in that included file can be included there to make sure they are available.

Fortunately, this required no code changes and only a change in our XML schema.  We already processed archetypes in a way that didn't care how many blocks there were, nor cared if there were duplicates of the same archetype.

A test has been added that demonstrates how to use this kind of inclusion, which requires the `xpointer` metadata in the include line.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,7 @@ Since last release
 * Table for recording active/dormant cycle periods (#1830)
 * Support for OpenMP parallelized simulations (#1709)
 * Command-line option to make a new blank input file referencing a current schema (#1861)
+* Allow multiple archetype blocks to facilitate includes ()
 
 **Changed:**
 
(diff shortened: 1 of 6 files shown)


abachma2 (approved):
Thanks for making these changes @gonuke. I'm approving and will merge once I see the Cyclus tests passing.

dean-krueger (approved):
Had to go fast to sneak in before Amanda could merge, but this looks good to me as well. Excited to try this out sometime!

gonuke:
> Had to go fast to sneak in before Amanda could merge, but this looks good to me as well. Excited to try this out sometime!

I think @trysarahtops will be eager to try this out to make her input files simpler!

abachma2:
All the tests passed. Merging now.

Issue #1879, opened 2025-07-14
[Task] Run clang-format on Cyclus


dean-krueger:
# What is this task about?

clang-format needs to be run on Cyclus, then those formatted files need to be reviewed and pushed

# Why is this task needed?

Cyclus is messy, and #1674 wants to add a workflow to check the style. This needs to be done to fix the former and prepare for the latter.

# Relevant resources or references

#1674 
#1878 

# How can this issue be closed?

This can be closed once clang-format has been run on all files in the Cyclus `src` directory and all tests pass without erroring out.

PR #1880, opened 2025-07-15
Add clang-format protections to Cyclus

### LINKED ISSUE #1878: [Task] Protect Cyclus from clang-format
# What is this task about?

Some parts of Cyclus will break if clang-format is run without any guards. These guards need to be added.

# Why is this task needed?

Specifically, any `#pragma` blocks in `.cycpp.h` files, as well as the order of `#include` statements will get messed up if you just run clang-format, so these two (as well as potentially other) guards need to be added before clang-format is run and #1674 is merged.

#1674 
[Protect include blocks](https://stackoverflow.com/questions/60334299/clang-format-disable-ordering-includes)

This issue will be good to close with a PR that runs clang-format on the entirety of Cyclus' `src` directory and doesn't error out.

dean-krueger:
# Summary of Changes

In preparation for #1674 some protections against clang-format need to be added to Cyclus. Specifically, the `#pragma` blocks in all the .cycpp.h files needed to be protected with `//clang-format off/on` guards, and an extra line needed to be added to the .clang-format file to make it not reorder the `#include` statements, since some of them depend on each other.

# Related CEPs and Issues

This PR is related to:

- #1674 
- Closes #1878 
- #1879  

# Associated Developers

[ahnaf-tahmid-chowdhury](https://github.com/ahnaf-tahmid-chowdhury)

# Design Notes

Besides the .cycpp.h files, the `query_backend.h` file needed to be protected in its entirety, since a python script reads it and looks for some very specific formatting that doesn't play nicely with clang-format. 

# Testing and Validation

Built and tested Cyclus and Cycamore on my local machine with Clean Builds. No additional unit testing added. clang-format was run on all files in src and then a clean-build was performed, which succeeded without error.

# Checklist

 - [x]  Add or update tests.
 - [ ]  Document if needed.
 - [x]  Update the changelog.

### DIFF
--- a/src/query_backend.h
+++ b/src/query_backend.h
@@ -1,3 +1,4 @@
+// clang-format off
 #ifndef CYCLUS_SRC_QUERY_BACKEND_H_
 #define CYCLUS_SRC_QUERY_BACKEND_H_
 
@@ -1044,3 +1045,4 @@ class Sha1 {
 }  // namespace cyclus
 
 #endif  // CYCLUS_SRC_QUERY_BACKEND_H_
+// clang-format on
--- a/.clang-format
+++ b/.clang-format
@@ -1,13 +1,12 @@
----
 Language:        Cpp
-BasedOnStyle:  Google
-# only allow single line functions if they are inlined in class def'n
+BasedOnStyle:    Google
+# Only allow single-line functions if they are inlined in class definition
 AllowShortFunctionsOnASingleLine: Inline
-# do not force new line after template declarations
+# Do not force new line after template declarations
 AlwaysBreakTemplateDeclarations: false
-# Automatically detect whether a given fuction's arguments are one-per-line
+# Automatically detect whether a given function's arguments are one-per-line
 ExperimentalAutoDetectBinPacking: true
 # Use C++11 standard
-Standard:        Cpp11
-...
-
+Standard: Cpp11
+# Prevent automatic include sorting
+SortIncludes: Never
--- a/src/toolkit/facility_cost.cycpp.h
+++ b/src/toolkit/facility_cost.cycpp.h
@@ -15,6 +15,7 @@
 ///    file with the other ones, reaplcing <param_name> with the name you put
 ///    in the econ_params array (again, must match exactly).
 
+// clang-format off
 #pragma cyclus var {"default" : 0.0,                                       \
                    "uilabel" : "Capital cost required to build facility", \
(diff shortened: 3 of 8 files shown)


src/query_backend.h
gonuke, line 1: Since this is the whole file, is there a way to simply exclude this file from being formatted?

Also curious to see what clang does to this file and whether we can make the python script more flexible....  Maybe a separate issue/PR.

gonuke (changes requested):
This looks pretty straightforward - thanks @dean-krueger.

I do have one question about ignoring whole files: https://clang.llvm.org/docs/ClangFormat.html#clang-format-ignore

.clang-format-ignore
```diff
@@ -0,0 +1,3 @@
+src/eventhooks.*
```
gonuke, line 1: Aren't these files python? Does clang-format touch them?

.clang-format-ignore
```diff
@@ -0,0 +1,3 @@
...
+src/decay.*
```
gonuke, line 2: I don't see any files matching this glob

.clang-format-ignore
```diff
@@ -0,0 +1,3 @@
...
+src/query_backend.h
```
gonuke, line 3: Let's add a comment (if allowed) about why this file is ignored.  Something like:
```suggestion
# query_backend.h is parsed by dbtpyes_gen.py that
# expects strict formatting of code and comments to 
# generate code for handling different data types
# in the backends
src/query_backend.h
```

gonuke (changes requested):
A few comments/questions about the ignored files, but otherwise looking to merge soon!

.clang-format-ignore
dean-krueger: They've got the "correct" file type for clang (`eventhooks_api.h`, `eventhooks.cc`, `eventhooks.h`, etc, so I figured that even though they get autogenerated every time you build cyclus, if someone was working on cyclus and formatting everything as they go they'd maybe run into issues with these files since they're autogenerated and some other file checks to make sure they're in the right format. Best to just leave them out.

.clang-format-ignore
dean-krueger: Sounds good. I can do that. Should I add comments to all of the ones I ignore, or just this one?

dean-krueger:
Should be good for another look @gonuke

PR #1894, opened 2025-07-25
Add unit_value to Resource Objects and Children

### LINKED ISSUE #1885: [Feature] Resource objects should carry a value
As realistic costs and prices are introduced, it is useful for discrete objects to carry an embodied value with them as they move through the cycle.

Add a new member variable to a `Resource` object that holds the unit value (or total value - TBD) of that object.

Methods that cause `Resource` objects to be modified should invoke an appropriate default behavior to updating the value.  E.g. `absorb()` should perform a mass-weighted average of the value.  However, `extract()` has no obvious default; in the real world extraction typically results in two streams of different value and archetype developers should be encouraged to manage this according to the economics and physics of their archetype.

Every `Resource` object has a value member variable that is treated consistently and correctly throughout its standard usage.

dean-krueger:
# Summary of Changes

A member variable "unit_value_" has been added to all resource-type objects (`Resource`, `Product`, and `Material`) and appropriate functions (`Absorb`, `Clone`, `Extract`) have been updated to have reasonable behavior concerning the new `unit_value_` member.

# Related CEPs and Issues

This PR is related to:

- Closes #1885 

# Associated Developers

# Design Notes

Wasn't sure how to handle packaging. May need to discuss in person, or here, either works.

# Testing and Validation

Built and tested Cyclus on my local machine. Added unit tests. Clean Build Cyclus and Cycamore, all tests for both passing.

# Checklist

 - [x]  Add or update tests.
 - [x]  Document if needed.
 - [x]  Update the changelog.
 - [x]   Run clang-format

### DIFF
--- a/src/material.cc
+++ b/src/material.cc
@@ -17,13 +17,15 @@ Material::~Material() {}
 Material::Ptr Material::Create(Agent* creator, double quantity,
                                Composition::Ptr c, std::string package_name,
                                double unit_value) {
-  Material::Ptr m(new Material(creator->context(), quantity, c, package_name, unit_value));
+  Material::Ptr m(
+      new Material(creator->context(), quantity, c, package_name, unit_value));
   m->tracker_.Create(creator);
   return m;
 }
 
 Material::Ptr Material::CreateUntracked(double quantity, Composition::Ptr c) {
-  Material::Ptr m(new Material(NULL, quantity, c, Package::unpackaged_name(), 0.0));
+  Material::Ptr m(
+      new Material(NULL, quantity, c, Package::unpackaged_name(), 0.0));
   return m;
 }
 
@@ -87,7 +89,8 @@ Material::Ptr Material::ExtractComp(double qty, Composition::Ptr c,
   }
 
   qty_ -= qty;
-  Material::Ptr other(new Material(ctx_, qty, c, Package::unpackaged_name(), UnitValue()));
+  Material::Ptr other(
+      new Material(ctx_, qty, c, Package::unpackaged_name(), UnitValue()));
 
   // Decay called on the extracted material should have the same dt as for
   // this material regardless of composition.
@@ -119,7 +122,8 @@ void Material::Absorb(Material::Ptr mat) {
     prev_decay_time_ = mat->prev_decay_time_;
   }
   double tot_mass = qty_ + mat->quantity();
-  double avg_unit_value = (qty_ * UnitValue() + mat->quantity() * mat->UnitValue()) / tot_mass;
+  double avg_unit_value =
+      (qty_ * UnitValue() + mat->quantity() * mat->UnitValue()) / tot_mass;
   SetUnitValue(avg_unit_value);
   qty_ = tot_mass;
   mat->qty_ = 0;
--- a/src/material.cc
+++ b/src/material.cc
@@ -153,7 +153,8 @@ Resource::Ptr Material::PackageExtract(double qty,
   }
 
   qty_ -= qty;
-  Material::Ptr other(new Material(ctx_, qty, comp_, new_package_name));
+  Material::Ptr other(
+      new Material(ctx_, qty, comp_, new_package_name, UnitValue()));
 
   // Decay called on the extracted material should have the same dt as for
   // this material regardless of composition.
--- a/src/material.cc
+++ b/src/material.cc
@@ -15,14 +15,15 @@ const ResourceType Material::kType = "Material";
 Material::~Material() {}
 
 Material::Ptr Material::Create(Agent* creator, double quantity,
-                               Composition::Ptr c, std::string package_name) {
-  Material::Ptr m(new Material(creator->context(), quantity, c, package_name));
+                               Composition::Ptr c, std::string package_name,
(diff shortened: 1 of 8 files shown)


dean-krueger:
Need to do changelog and clang-format, but need to run to something real quick, so I'll do those later today.

src/resource.h
```diff
@@ -112,6 +119,8 @@ class Resource {
   static int nextstate_id_;
   static int nextobj_id_;
   int state_id_;
+  // Unit value of the resource.
+  double unit_value_;
```
gonuke, line 123: (Pedantic) let's put this either after or before all the ID variables.

src/product.cc
```diff
@@ -78,6 +86,7 @@ std::string Product::package_name() {
   return package_name_;
 }
 
+// Not sure what to do here with the unit value and packaging...
```
gonuke, line 89: This should behave the same way as `Extract` from the point if view of the unit value

gonuke: But I think I'll need to remember why this extract needs to be special and add better comments for why

src/material.cc
```diff
@@ -86,7 +89,8 @@ Material::Ptr Material::ExtractComp(double qty, Composition::Ptr c,
   }
 
   qty_ -= qty;
-  Material::Ptr other(new Material(ctx_, qty, c, Package::unpackaged_name()));
+  Material::Ptr other(
```
gonuke, line 92: Need to add similar to `PackageExtract`

tests/resource_tests.cc
```diff
@@ -72,17 +76,29 @@ TEST_F(ResourceTest, MaterialAbsorbGraphid) {
 
 TEST_F(ResourceTest, MaterialExtractTrackid) {
   int obj_id = m1->obj_id();
-  Material::Ptr m3 = m1->ExtractQty(2);
+  Material::Ptr m4 = m1->ExtractQty(2);
```
gonuke, line 79: I wonder if we should rename all the local objects to avoid potential name collisions in the future
```suggestion
  Material::Ptr lm1 = m1->ExtractQty(2);
```
Add 'l' for "local" and restart numbering at 1?

gonuke (changes requested):
This looks like a good start. In my next review, I'll more carefully look for places where changes should exist by may be forgotten.

src/resource.h
dean-krueger: Oops, thought I did, but I guess I missed the `int obj_id_` below. My bad!

src/material.cc
dean-krueger: Done. Good catch.

dean-krueger:
So in my search I found the following method in `res_tracker.cc`:

```
void ResTracker::Record(bool bumpId) {
  if (bumpId) {
    res_->BumpStateId();
  }
  ctx_->NewDatum("Resources")
      ->AddVal("ResourceId", res_->state_id())
      ->AddVal("ObjId", res_->obj_id())
      ->AddVal("Type", res_->type())
      ->AddVal("TimeCreated", ctx_->time())
      ->AddVal("Quantity", res_->quantity())
      ->AddVal("Units", res_->units())
      ->AddVal("QualId", res_->qual_id())
      ->AddVal("PackageName", res_->package_name())
      ->AddVal("Parent1", parent1_)
      ->AddVal("Parent2", parent2_)
      ->Record();
  res_->Record(ctx_);
}
```

I'm not totally sure if this is the one you were talking about with the `Tracker` bit of what you were saying, but that __appears__ to be the only place in the code where the `Resources` table gets specifically written to (that method gets called when resources are created, modified, split, combined, or repackaged, though, through the `tracker_` object methods). It LOOKS like this table is the one that records all the other properties of a resource (id, type, units, etc), so to me it seems like this is where we would want to write out its value (especially since this method gets called whenever resources are modified, too.

Interested to hear your thoughts!

gonuke:
> So in my search I found the following method in `res_tracker.cc`:

Good find!  I think that's the place we want to do it, for sure.

dean-krueger:
On it!

src/material.cc
```diff
@@ -15,14 +15,17 @@ const ResourceType Material::kType = "Material";
...
-  Material::Ptr m(new Material(creator->context(), quantity, c, package_name));
+                               Composition::Ptr c, std::string package_name,
+                               double unit_value) {
+  Material::Ptr m(
+      new Material(creator->context(), quantity, c, package_name, unit_value));
   m->tracker_.Create(creator);
   return m;
 }
 
 Material::Ptr Material::CreateUntracked(double quantity, Composition::Ptr c) {
```
gonuke, line 26: Should we allow untracked materials to be given a `unit_value` upon creation?  One can always call `SetUnitValue()` after the fact, but it might be more convenient if we allow it directly during creation?

gonuke (changes requested):
One last comment and then I think I'm ready to move forward and address anything else as it comes.

gonuke (approved):
I think this looks good now.  It will be hard to exhaustively explore all the places that this will matter, so in the future, we'll need to carefully look for places where the unit value gets modified or reset to default accidentally.

PR #1895, opened 2025-08-01
Add a Warning When a Facility Trades with itself

dean-krueger:
# Summary of Changes

It was identified as potentially undesirable while testing the Tricycle code for Fusion in Cyclus to have a facility trading with itself. However, sometimes facilities actually want to do this. The proposed solution was, then, to simply warn the user whenever this was happening. This PR adds that warning by throwing it during the RecordTrades() function in the `trade_executor`. It also adds some AI generated tests to cover this behavior.

# Related CEPs and Issues

This PR is related to:

- Closes cyclus/cycamore#645

# Associated Developers

Cursor AI

# Design Notes

Implementation is fairly straightforward. Mainly look at `trade_executor.h` for the "meat" of this PR (though most new lines are in the test file)

# Testing and Validation

Built and tested Cyclus on my local machine.  Had Cursor generate the tests (though I read through them and checked that they were doing the right thing, then manually verified the behavior with a test file)

Screenshot 2025-07-31 at 11 21 39 PM

Screenshot 2025-07-31 at 11 12 48 PM

Note that the warning in the second image is only given when the Storage facility in my input file trades with itself, and not when the Conversion facility trades with the Storage facility or the Mine trades with Conversion.

All Cyclus and Cycamore unit tests pass. Code builds, *should* be fine.

# Checklist

 - [x]  Add or update tests.
 - [x]  Document if needed.
 - [x]  Update the changelog.
 - [x]   Run clang-format


tests/self_trading_warning_tests.cc
gonuke, line 1: Let's move these tests into the tests for `trade_executor`, please

gonuke (changes requested):
Reminder to move tests

tests/self_trading_warning_tests.cc
dean-krueger: Thanks for the reminder! I'll do that now.

tests/trade_executor_tests.cc
```diff
@@ -187,3 +188,225 @@ TEST_F(TradeExecutorTests, NoThrowWriting) {
 //   delete s;
 //   delete r;
 // }
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+// Simple test facility for self-trading tests
+class SelfTradingTestFacility : public cyclus::Facility {
```
gonuke, line 194: All of these test seem to be much more heavy duty than the existing trade executor tests. Those tests use the `TestTrader`, create a number of agents in `Setup` and establish a number of traders that are then subjected to different conditions.  Could we just subject some of those `TestTrader`s to the condition we are testing?

gonuke (changes requested):
These tests seem to be more than what's necessary compared to other trade executor tests, but maybe I'm wrong...

tests/trade_executor_tests.cc
dean-krueger: It's also possible that I missed the point of your original comment, in which case: could you expand a bit on what you mean?

gonuke: > I took a look at `test_trader.h`, and I see a few potential problems with doing that. Specifically, the big one is that `TestTrader`s have an `is_requester()` flag which controls whether it requests material (`is_requester = true`) or offers material (`is_requester = false`). Then, in its implementation of `GetMatlBids` and `GetMatlRequests` it checks whether `is_requester` is true or not, and depending on the answer returns empty sets for the corresponding function:
> 
> 
> 
> ```
> 
> virtual std::set<BidPortfolio<Material>::Ptr> GetMatlBids(CommodMap<Material>::type& commod_requests) {
> 
>     ...
> 
>     if (obj_fac == NULL || is_requester) {
> 
>       return std::set<BidPortfolio<Material>::Ptr>();
> 
>     } else {
> 
>       ...
> 
>     }
> 
>   }
> 
>   ```
> 
> 
> 
> (the inverse happens in `GetMatlRequests`).
> 
> 
> 
> Since the purpose of these tests is to see when a facility trades with itself, it *needs* to be able to do both of these at the same time (otherwise the warning won't be triggered). It would probably be possible to refactor `test_trader.h` to remove this flag and implement both functions at the same time, BUT I have no idea what sort of downstream effects that would have. Additionally, the `TestTrader` uses a bunch of factory methods which I certainly do not understand (though I plan on learning how these work eventually...). This adds a bunch of complexity to how they work.
> 
> 
> 
> TLDR: `TestTrader` doesn't work as written. I could probably refactor, but that might cause all sorts of downstream trouble. `TestTrader` is kind of complex. The `SelfTradingTestFacility` is specifically designed to test this exact warning. 
> 
> 
> 
> Happy as always to spend time doing whatever you think is best (honestly)!

Great insight here!  Let me take another look, but likely to leave things largely as you proposed.

(Interesting that `TestTrader` explicitly avoids self-trading 😀 )

gonuke (changes requested):
I put this side-by-side with the `TestTrader` definition and related tests and came up with a number of suggestions.

tests/trade_executor_tests.cc
```diff
@@ -187,3 +188,225 @@ TEST_F(TradeExecutorTests, NoThrowWriting) {
...
+ protected:
+  std::unique_ptr<cyclus::TestContext> tc_;
+  SelfTradingTestFacility* facility_;
+};
+
+TEST_F(SelfTradingWarningTest, SelfTradingWarningIssued) {
+  // Capture stderr to check for warning messages
+  std::stringstream captured_stderr;
+  std::streambuf* original_stderr = std::cerr.rdbuf();
+  std::cerr.rdbuf(captured_stderr.rdbuf());
```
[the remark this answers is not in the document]
dean-krueger: You're totally right! I remembered that we had found some way to avoid doing all this, but couldn't remember what it was. Thanks for the reminder. Hopefully this time it'll stick

tests/trade_executor_tests.cc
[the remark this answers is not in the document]
dean-krueger: This should have mostly been taken care of now, but I've still got some clean up tasks left so I'll give it another look before I re-request review!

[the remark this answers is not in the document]
dean-krueger: I think these have all been taken care of. I wasn't TOTALLY sure what you meant by 

> If you include some specific headers

But I just changed the `using cyclus Composision ...` bit to `using namespace cyclus` since there were like 15 of them and I figured that was a. fine to do since we were just about including them all anyway, and b. a good solution to get rid of all the `cyclus::...` bits. If that's not what you meant then I'm happy to do something different!

tests/trade_executor_tests.cc
```diff
@@ -187,3 +188,225 @@ TEST_F(TradeExecutorTests, NoThrowWriting) {
...
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+// Simple test facility for self-trading tests
+class SelfTradingTestFacility : public cyclus::Facility {
+ public:
+  SelfTradingTestFacility(cyclus::Context* ctx) : cyclus::Facility(ctx) {}
+  virtual ~SelfTradingTestFacility() {}
+
+  virtual cyclus::Agent* Clone() { 
+    return new SelfTradingTestFacility(context());
```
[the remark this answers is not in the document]
dean-krueger: Done! Good catch.

dean-krueger:
Think this should be good now for another round of review!

tests/trade_executor_tests.cc
```diff
@@ -168,22 +162,124 @@ TEST_F(TradeExecutorTests, NoThrowWriting) {
...
+  }
+
+  virtual void TearDown() {
+    delete facility_;
+  }
+
+ protected:
+  std::unique_ptr<TestContext> tc_;
+  SelfTradingTestFacility* facility_;
+  Composition::Ptr test_comp_;
```
gonuke, line 192: I think this is only used in `Setup()` so can be a local variable there.

tests/trade_executor_tests.cc
```diff
@@ -168,22 +162,124 @@ TEST_F(TradeExecutorTests, NoThrowWriting) {
...
+  delete bid;
+  delete req;
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+TEST_F(SelfTradingWarningTest, NoWarningForDifferentAgents) {
+  // Create a second test facility
+  SelfTradingTestFacility* facility2 = new SelfTradingTestFacility(tc_->get());
+  facility2->Build(nullptr);
+  facility2->EnterNotify();
```
gonuke, line 228: Why not just make both facilities in `Setup()`?

tests/trade_executor_tests.cc
```diff
@@ -168,22 +162,124 @@ TEST_F(TradeExecutorTests, NoThrowWriting) {
...
+  Composition::Ptr test_comp_;
+  Material::Ptr test_mat_;
+  
+  // Test constants
+  static constexpr double kTestTradeAmount = 50.0;
+};
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+TEST_F(SelfTradingWarningTest, SelfTradingWarningIssued) {
+  warn_as_error = true;
```
gonuke, line 201: Maybe do this as late as possible, I.e. immediately before the call that throws the warning

tests/trade_executor_tests.cc
```diff
@@ -168,22 +162,124 @@ TEST_F(TradeExecutorTests, NoThrowWriting) {
...
+}
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+TEST_F(SelfTradingWarningTest, NoWarningForDifferentAgents) {
+  // Create a second test facility
+  SelfTradingTestFacility* facility2 = new SelfTradingTestFacility(tc_->get());
+  facility2->Build(nullptr);
+  facility2->EnterNotify();
+  
+  warn_as_error = true;
```
gonuke, line 230: Ibid

tests/trade_executor_tests.cc
```diff
@@ -168,22 +162,124 @@ TEST_F(TradeExecutorTests, NoThrowWriting) {
...
+  std::vector<Trade<Material>> trades;
+  trades.push_back(trade);
+  
+  
+  TradeExecutor<Material> executor(trades);
+
+  // Use a try/catch to check for the error message
+  try {
+  executor.ExecuteTrades(tc_->get());
+  FAIL() << "Expected error to be thrown";
```
gonuke, line 274: What does this line do?

gonuke (changes requested):
This looks pretty good to me now... just a few minor requests

tests/trade_executor_tests.cc
dean-krueger: This line says that if the trade went through (we typically warn but here we're throwing), fail the test and then give a reason "Expected error to be thrown" in the test output.

gonuke (changes requested):
Sorry for a new round of comments - most things I should have raised sooner.  I think this could all be streamlined into a single test for the new condition you added, and add a new macro into an existing test for the "no throw" condition.

tests/trade_executor_tests.cc
```diff
@@ -215,9 +216,83 @@ TEST_F(TradeExecutorTests, NoThrowWriting) {
...
+    facility2_ = new SelfTradingTestFacility(tc_->get());
+    facility2_->Build(nullptr);
+    facility2_->EnterNotify();
+    
+    CompMap v;
+    v[id("u235")] = 1;
+    double trade_amt = 100;
+
+    Composition::Ptr test_comp_ = Composition::CreateFromAtom(v);
+    test_mat_ = Material::CreateUntracked(trade_amt, test_comp_);
```
gonuke: Maybe I'm missing something, but can't you rely on the `obj_fac` to provide a material for you and don't need to actually build this material?

gonuke:
The more I look at this, the more I wonder why we need all the infrastructure of a new testing class, and can't just add a single new test based on the existing `TestTrader`?

gonuke:
> The more I look at this, the more I wonder why we need all the infrastructure of a new testing class, and can't just add a single new test based on the existing `TestTrader`?

* `TradeExecutorTests` has an additional member variable `self_trades` of type `Trades`
* `TradeExecutorTests::SetUp()` populates `self_trades` in the same way it current populates `trades`
* a new test function looks like the `WholeShebang` test, but just tests for the warning.

dean-krueger:
Refactored things to use the regular `TestTrader`, then got rid of the `SelfTradeTestTrader`. Should be good for another look.

tests/trade_executor_tests.cc
```diff
@@ -215,9 +227,83 @@ TEST_F(TradeExecutorTests, NoThrowWriting) {
...
+  
+  // Test that trades between different agents don't trigger warnings
+  warn_as_error = true;
+  TradeExecutor<Material> warning_exec(trades);
+  EXPECT_NO_THROW(warning_exec.ExecuteTrades(tc.get()));
+  warn_as_error = false;
 }
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+class SelfTradingWarningTest : public ::testing::Test {
```
gonuke: It's been a while, but I wonder why you can't build the necessary new test within the structure of the existing test class?  It already seems to have the ability to set up requests, bids and execute trades...

gonuke (changes requested):
I'm still not sure we need so much new testing infrastructure here.

tests/trade_executor_tests.cc
dean-krueger: I probably can! This was just the more minimal change for me yesterday 😬

tests/trade_executor_tests.cc
dean-krueger: it's possible! I didn't look too hard into this yesterday, but I'll take another look more holistically at the whole "testing infrastructure" and see if I can't reduce it while still adding the necessary tests.

dean-krueger:
Assuming the tests pass here like they did on my machine this should be good for another look. Since this is a legacy Cursor job (and quite honestly I was feeling kind of lazy) I just had Codex refactor this. It made a few decisions that were interesting, but it defended them satisfactorily for me:

1. removing pyne - this was added by Cursor to make a material that you referenced in your comment asking why not just use the built in ones. Codex seemed to agree, and so it dropped the Pyne requirement.
2. removing the test macro - this was done because it determined the test was better if we actually checked that the ID given by that warning matched the ID we expected, so it wanted to be able to read the message. The old macro didn't allow this (it could only see that it threw an error. I asked it if there was a way to just do that with a macro and it said "yeah!" and then started speaking the C++ equivalent of Bulgarian as far as I could tell so I figured it was being too clever for me and ditched the macro.
3. `catch (...) {` - I had never seen this pattern before, but apparently you can do this like a wildcard catch and it chose to do it just so that if anything mysterious or unexpected happened that wasn't the error we expect it didn't leave `warn_as_error` stuck on true (a little hygiene inclusion which I support). I left a comment (hopefully) explaining as much.
4. using try/catch at all - this was done so that it could look at the error and test that it has the right ID in it. I support, though I know you often don't love this pattern.


and then a reminder that `std::string::npos` is like "we didn't find anything matching this" so that line: 

`EXPECT_TRUE(warning_msg.find(expected_id) != std::string::npos);`

is just "expect that we found the `expected_id` inside warning_msg". I say this because I feel like I remember us having a back and forth about what that was before (though maybe I'm misremembering).

gonuke (approved):
LGTM - thanks for all the iteration @dean-krueger

PR #1897, opened 2025-08-06
Allow Cyclus to Record the Adjusted Preference/Cost of Trades

### LINKED ISSUE #1896: [Bug] The RecordTrade() function doesn't actually have access to the values the DRE uses for trades
# Describe the bug.

A recent feature that was added to cyclus was the ability for the Transactions table to record the "cost" associated with each trade that was executed during a simulation. This is going to be important for future economic features in cyclus. However, the cost it has access to here is blind to any preference adjustments that go on "behind the scenes" in the exchange context of cyclus. This means that any future attempts to create things like tariffs will not be represented in this output.

# Steps to reproduce the bug.

If you give a source bid a pref and then have another agent adjust the incoming bid preference Transactions doesn't record the change that gets made, and only records the original bid cost.

 - Operating System:
 - Cyclus Version:
 - Python Version:

# Additional context.

dean-krueger:
# Summary of Changes

While playing with something else, I noticed that when I was adjusting the preference of trades the `RecordTrades()` function, which writes to the Transactions table in the database wasn't reflecting the adjusted cost, but was instead simply recording the original bid cost *before* the adjustment. I am not certain that we will necessarily want to represent this cost as "real", but it would be nice to be able to have access to it. For now, I had Cursor go in and give that function access to the exchange context of the DRE so it can see the adjusted cost, and changed "RecordTrades" to record this cost (if it was adjusted, and otherwise just the normal cost). I looked at what it did, and it more or less made sense and didn't seem too wild, so I figured I'd make a PR since this is a pretty limited change.

# Related CEPs and Issues

This PR is related to:

- Closes #1896 

# Associated Developers

Cursor AI

# Design Notes

Again, not totally sure if the desired behavior is to always record this as the "cost", but I found myself personally wanting access to it. Maybe adding a column to the `RecordTrades()` Transactions table to have a "bid" and an "adjusted cost" section would be better? Not totally sure...

# Testing and Validation

Built and tested Cyclus on my local machine.  All CURRENT unit tests pass. 

NO NEW UNIT TESTS HAVE BEEN ADDED (YET)

# Checklist

 - [x]  Add or update tests.
 - [x]  Document if needed.
 - [x]  Update the changelog.
 - [x]   Run clang-format

### DIFF
--- a/src/trade_executor.h
+++ b/src/trade_executor.h
@@ -119,14 +119,18 @@ template <class T> class TradeExecutor {
           // If we have access to the exchange context, use the adjusted
           // preference that was actually used by the solver
           if (ex_ctx) {
-            try {
-              adjusted_preference = ex_ctx->trader_prefs.at(
-                  trade.request->requester())[trade.request][trade.bid];
-            } catch (const std::out_of_range&) {
-              // If the preference is not found in the exchange context, fall
-              // back to the original preference This could happen if the trade was
-              // not part of the original exchange
+            auto trader_it = ex_ctx->trader_prefs.find(trade.request->requester());
+            if (trader_it != ex_ctx->trader_prefs.end()) {
+              auto request_it = trader_it->second.find(trade.request);
+              if (request_it != trader_it->second.end()) {
+                auto bid_it = request_it->second.find(trade.bid);
+                if (bid_it != request_it->second.end()) {
+                  adjusted_preference = bid_it->second;
+                }
+              }
             }
+            // If any of the keys are not found, adjusted_preference remains
+            // the original preference
           }
 
           ctx->NewDatum("Transactions")
--- a/src/trade_executor.h
+++ b/src/trade_executor.h
@@ -71,7 +71,8 @@ template <class T> class TradeExecutor {
     SendTradeResources(trade_ctx_);
   }
 
-  /// @brief execute all trades with access to exchange context for adjusted preferences
+  /// @brief execute all trades with access to exchange context for adjusted
+  /// preferences
   void ExecuteTrades(Context* ctx, ExchangeContext<T>* ex_ctx) {
     GroupTradesBySupplier(trade_ctx_, trades_);
     GetTradeResponses(trade_ctx_);
@@ -85,15 +86,15 @@ template <class T> class TradeExecutor {
   ///
   /// @param ctx the Context through which communication with backends will
   /// occur
-  void RecordTrades(Context* ctx) {
-    RecordTrades(ctx, NULL);
-  }
+  void RecordTrades(Context* ctx) { RecordTrades(ctx, NULL); }
 
-  /// @brief Record all trades with the appropriate backends, using adjusted preferences
+  /// @brief Record all trades with the appropriate backends, using adjusted
+  /// preferences
   ///
   /// @param ctx the Context through which communication with backends will
   /// occur
-  /// @param ex_ctx the ExchangeContext containing the adjusted preferences used by the solver
+  /// @param ex_ctx the ExchangeContext containing the adjusted preferences used
+  /// by the solver
   void RecordTrades(Context* ctx, ExchangeContext<T>* ex_ctx) {
     // record all trades
     typename std::map<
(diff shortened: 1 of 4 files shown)


src/trade_executor.h
```diff
@@ -90,14 +112,36 @@ template <class T> class TradeExecutor {
...
+          double adjusted_preference = trade.bid->preference();
+
+          // If the bid has NaN preference, use the request preference
+          if (std::isnan(adjusted_preference)) {
+            adjusted_preference = trade.request->preference();
+          }
+
+          // If we have access to the exchange context, use the adjusted
+          // preference that was actually used by the solver
+          if (ex_ctx != NULL) {
```
gonuke, line 126: I think this is the same thing ...
```suggestion
          if (ex_ctx) {
```

gonuke (changes requested):
Thanks @dean-krueger - Cursor seems to understand Cyclus better than I do 😀 

1. Some additional tests would be great. (Maybe we should add that to the checkboxes in the PR template)

2. Do you think it will be useful to record both preferences in some cases, so that we know the adjustment?  Or maybe even develop a way to record all the adjustments?

src/trade_executor.h
```diff
@@ -69,11 +71,31 @@ template <class T> class TradeExecutor {
     SendTradeResources(trade_ctx_);
   }
 
+  /// @brief execute all trades with access to exchange context for adjusted
+  /// preferences
+  void ExecuteTrades(Context* ctx, ExchangeContext<T>* ex_ctx) {
```
gonuke, line 76: Can't the above (original) `ExecuteTrades` method just wrap this like you did with `Record`?

src/trade_executor.h
```diff
@@ -69,11 +71,31 @@ template <class T> class TradeExecutor {
...
+    GroupTradesBySupplier(trade_ctx_, trades_);
+    GetTradeResponses(trade_ctx_);
+    if (ctx != NULL) {
```
gonuke, line 79:
```suggestion
    if (ctx) {
```

gonuke (changes requested):
Found a couple of additional suggestions

gonuke:
> (Maybe we should add that to the checkboxes in the PR template)

Apparently it already is! 🤦‍♂️

src/trade_executor.h
dean-krueger: yes.

dean-krueger:
> Do you think it will be useful to record both preferences in some cases, so that we know the adjustment? 

yes, and also this sort of opens the doors to this whole idea of "is an adjustment to the cost 'real', or not? Sometimes yes sometimes no? How do we deal with THAT?" Either way, here's a little screenshot below showing both at the same time.

Screenshot 2025-08-09 at 2 25 43 PM

dean-krueger:
Tests added! Think this is ready for another round of review. I went through and tried to clean things up as much as possible, but I may have missed some things, so feel free to critique!

dean-krueger:
Another note: since I changed how the Transaction table works, if this gets merged an issue needs to be made to go fix Cycamore again in the same way you did earlier (by having it ignore the "Cost" column, which now needs to be the "BidCost" and "AdjustedCost" columns)

src/trade_executor.h
```diff
@@ -90,14 +105,39 @@ template <class T> class TradeExecutor {
...
+          // preference that was actually used by the solver
+          if (ex_ctx) {
+            try {
+              adjusted_preference = ex_ctx->trader_prefs.at(
+                  trade.request->requester())[trade.request][trade.bid];
+            } catch (const std::out_of_range&) {
+              // If the preference is not found in the exchange context, fall
+              // back to the original preference This could happen if the trade was
+              // not part of the original exchange
+            }
```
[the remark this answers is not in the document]
dean-krueger: There sure is! It gets rid of the whole "try/catch" thing too. Good suggestion.

dean-krueger:
Things (especially the tests) have been cleaned up and your comments have been responded to. I'll do one more thorough look through before re-requesting review (and while I wait for the tests to finish running), but I think this should be pretty close now!

src/trade_executor.h
```diff
@@ -90,14 +105,43 @@ template <class T> class TradeExecutor {
...
+          if (ex_ctx) {
+            auto trader_it = ex_ctx->trader_prefs.find(trade.request->requester());
+            if (trader_it != ex_ctx->trader_prefs.end()) {
+              auto request_it = trader_it->second.find(trade.request);
+              if (request_it != trader_it->second.end()) {
+                auto bid_it = request_it->second.find(trade.bid);
+                if (bid_it != request_it->second.end()) {
+                  adjusted_preference = bid_it->second;
+                }
+              }
```
gonuke, line 130: (Not for this PR) Might want to make an issue to explore adding a method to the `ExchangeContext` to find something.  This may be the first time we've done this, but it seems like it might become useful in other cases.

PR #1903, opened 2025-08-22
TEST PR (conda-forge tests are failing)

dean-krueger:
This PR is a test. Do NOT merge. Hopefully this passes all tests.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -15,6 +15,7 @@ RUN apt update -y && apt install -y \
         make \
         libxml2-dev \
         libxml++2.6-dev \
+        libblas-dev \
         liblapack-dev \
         pkg-config \
         coinor-libcbc-dev \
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -60,13 +60,15 @@ RUN mamba install -y "python<3.13" --no-pin && \
     mamba update -y --all && \
     mamba install -y \
                compilers \
-               cmake \
+               "cmake=4.0.3" \
                make \
                git \
                glib \
                libxml2 \
+               "libblas=3.11.0" \
```
gonuke, line 68: I don't think this version exists, and looking closer ... the current condo-forget version (3.9.0) is many months old and probably not causing this

gonuke (changes requested):

docker/Dockerfile
```diff
@@ -60,13 +60,15 @@ RUN mamba install -y "python<3.13" --no-pin && \
     mamba update -y --all && \
     mamba install -y \
                compilers \
-               cmake \
+               "cmake=4.0.3" \
```
gonuke, line 63: Try
```suggestion
               "cmake<4.1" \
```

gonuke (changes requested):

docker/Dockerfile
dean-krueger: I found it going back on the conda-forge page [here](https://anaconda.org/conda-forge/libblas/files?page=7). It seemed like that last "version" (not build) so I tried it.

PR #1907, opened 2025-08-25
Timeshift

gonuke:
Allow archetype developers to calculate a shift in time between the simulation start time and some arbitrary time stamp specified as a year and month.

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -315,6 +315,10 @@ int Context::time() {
   return ti_->time();
 }
 
+int Context::timeshift(int year, int month) {
+  return ti_->timeshift(year, month);
+}
+
 int Context::random() {
   return rng_->random();
 }
--- a/src/context.cc
+++ b/src/context.cc
@@ -315,8 +315,8 @@ int Context::time() {
   return ti_->time();
 }
 
-int Context::timeshift(int year, int month) {
-  return ti_->timeshift(year, month);
+int Context::CalcTimeDiff(int year, int month) {
+  return ti_->CalcTimeDiff(year, month);
 }
 
 int Context::random() {
--- a/src/context.cc
+++ b/src/context.cc
@@ -50,21 +50,6 @@ SimInfo::SimInfo(int dur, int y0, int m0, std::string handle)
       seed(kDefaultSeed),
       stride(kDefaultStride) {}
 
-SimInfo::SimInfo(int dur, int y0, int m0, std::string handle, int timestep)
-    : duration(dur),
-      y0(y0),
-      m0(m0),
-      dt(timestep),
-      decay("manual"),
-      branch_time(-1),
-      handle(handle),
-      explicit_inventory(false),
-      explicit_inventory_compact(false),
-      parent_sim(boost::uuids::nil_uuid()),
-      parent_type("init"),
-      seed(kDefaultSeed),
-      stride(kDefaultStride) {}
-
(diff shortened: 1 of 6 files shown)


src/context.cc
```diff
@@ -315,6 +315,10 @@ int Context::time() {
   return ti_->time();
 }
 
+int Context::timeshift(int year, int month) {
```
dean-krueger, line 318: Looking at this while it's still a draft because I don't want to work on my real work at the moment, and the name sticks out to me. It suggests that you're doing something to shift time by the input (ie moving the simulation time instead of calculating the difference in time). Perhaps a different name such as `CalcElapsedTime()` or something similar would be better?

src/timer.cc
```diff
@@ -281,6 +281,22 @@ int Timer::dur() {
...
+  int timediff_ = 0;
+  if (year >= 0) {
+    timediff_ += (year - si_.y0) * kMonthsPerYear;
+  }
+  
+  if (month >= 0) {
+    timediff_ += month - si_.m0;
+  }
+
+  return timediff_;
```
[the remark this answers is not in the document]
dean-krueger: Completed with [56dd1f0](https://github.com/cyclus/cyclus/pull/1907/commits/56dd1f07f7538ab8c5f904ab7bc870391409fe0a). Commenting for clarity.

tests/timer_tests.cc
```diff
@@ -115,6 +115,70 @@ class TimerTestsFixture : public ::testing::TestWithParam<int> {
     #endif // CYCLUS_IS_PARALLEL
 };
 
+
+TEST_P(TimerTestsFixture, TimeShiftTest) {
+  cyclus::PyStart();
+  cyclus::Recorder rec;
+  cyclus::Timer ti;
+  cyclus::Context ctx(&ti, &rec);
+
```
dean-krueger, line 124: it MIGHT be nice to have a comment here explaining that the siminifo defaults to 2010, 1 (or whatever it is, I think it's that, though). Not super necessary, but would make reading the tests slightly easier.

dean-krueger: You can piece it out by reading, but still...

tests/timer_tests.cc
```diff
@@ -115,6 +115,70 @@ class TimerTestsFixture : public ::testing::TestWithParam<int> {
...
+  cyclus::SimInfo short_timestep(5);
+  short_timestep.dt = 1;
+
+  non_default_ti.Initialize(&ctx2, short_timestep);
+
+  obs = non_default_ti.CalcTimeDiff(2010, 8);
+  exp = 7 * cyclusMonth / ctx2.dt();
+  EXPECT_EQ(exp, obs);
+
+    // Create and initialize a new timer with a weird non-default timestep
```
dean-krueger: I think this is my fault, but the indentation on this comment appears to be off.

dean-krueger (changes requested):
Thanks @gonuke, I just had two really minor comments (one of them was my fault I think) and then I think this is good to go.

tests/timer_tests.cc
gonuke: I may have pieced that out myself :)

dean-krueger (approved):
Thanks for the quick changes @gonuke ! This looks good to me. I'm gonna set a timer on my phone for 30 minutes or so to let the tests run, and will merge after they all pass!

PR #1912, opened 2025-08-26
Add a Progress Bar to Cyclus Simulations

### LINKED ISSUE #1911: [Feature] Cyclus Simulations could use a Progress Bar

Long Cyclus simulations are rare, but occasionally are needed. It would be really nice if there was a little progress bar that showed you how far along they were

A little progress bar that displays in the terminal sort of like the ones you see when you download a package on conda or something like that. I'm imagining a little `(#######                         )` or something like that

Printing out the timestep you're on could be the same sort of thing?

By having some sort of indicator for how far along your simulation is

dean-krueger:
# Summary of Changes

This PR adds a progress bar to the Cyclus simulation loop to provide visual feedback on simulation progress. This feature is maybe not the most useful for short Cyclus simulations, which run VERY quickly, but for longer ones like those @trysarahtops is running it will be very nice to have.

## Features

- **Visual Progress**: Shows a progress bar with percentage and fraction (e.g., `[=========>     ] 45.2% (452/1000)`)
- **Smart Updates**: Automatically adjusts update frequency based on simulation duration
- **Configurable**: Can be enabled/disabled via environment variable
- **Non-intrusive**: Works alongside existing logging output

## Usage

The progress bar is enabled by default and will automatically appear during simulation runs.

### Environment Variables

- `CYCLUS_PROGRESS_BAR`: Control whether the progress bar is shown
  - Set to `0`, `false`, `no`, or `off` to disable
  - Set to `1`, `true`, `yes`, or `on` to enable (default)

### Automatic Disabling

The progress bar is automatically disabled when:
- Verbose logging is enabled (e.g., `-v 2` or higher verbosity levels)
- This prevents the progress bar from interfering with verbose debug output

### Example

```bash
# Disable progress bar
export CYCLUS_PROGRESS_BAR=0
cyclus input.xml

# Enable progress bar (default)
export CYCLUS_PROGRESS_BAR=1
cyclus input.xml

# Progress bar will be automatically disabled with verbose logging
cyclus input.xml -v 3 -o test.sqlite
```

## Update Frequency

The progress bar automatically adjusts its update frequency based on simulation duration:

- **Short simulations** (≤20 timesteps): Updates every timestep
- **Medium simulations** (21-100 timesteps): Updates every 5 timesteps  
- **Long simulations** (>100 timesteps): Updates every 10 timesteps

This helps reduce output noise while still providing useful progress information.

# Related CEPs and Issues

This PR is related to:

- Closes #1911 

# Associated Developers

Cursor AI: Primary code writing (~100%)

# Design Notes

I chose to try this with Cursor instead of using a prebuilt package for two main reasons:
1. It seemed easier to do with Cursor and I was curious about how well it would work
2. I wasn't sure about mixing things up with the license or creating new dependencies, which seemed like a mess

The progress bar is implemented in:
- `cyclus/src/progress_bar.h` - Header file
- `cyclus/src/progress_bar.cc` - Implementation
- `cyclus/src/timer.cc` - Integration with simulation loop

The progress bar uses ANSI escape codes to update in place and works with the existing logging system.

Some pictures of the bar working with and without verbose output can be found below:

Progress bar sits at bottom of screen
Screenshot 2025-08-26 at 12 59 46 PM

No progress bar when -v flag is set higher than 2 since after that point the timestep is printed anyway. 
Screenshot 2025-08-26 at 1 13 47 PM

# Testing and Validation

Built and tested Cyclus on my local machine.  All current unit tests pass. However, no new unit tests have been added. 

# Checklist

 - [x]  Add or update tests.
 - [x]  Document if needed.
 - [x]  Update the changelog.
 - [x]   Run clang-format


gonuke:
This appears to need a rebase for the CHANGELOG... and wonder if you've looked at this[ pre-existing progressbar full of options/features](https://github.com/p-ranav/indicators), and whether it makes sense to just use it?

We can just copy the files into our repo (per the license).  One benefit, in theory, is that if it is improved in the future, we can take advantage of it by just dropping in an updated copy.

dean-krueger:
No that's a good idea. I didn't know that you could just drop those files into the repo, and thought that we'd have to do some fancy dependency stuff. I'll take a look at whether that would be a good idea or not (the progress bar I had cursor makes is pretty specific to the way I wanted it to work and has some nice features/is sort of "exactly what I want" and not over-rich with features we won't use, so I may still argue that it's better later, but it's worth taking a look at this other one first).

abachma2 (commented):
Thanks for making this @dean-krueger. I have a few questions, some of which come from my lack of knowledge of the cyclus kernel/c++. If you decide to implement the other method you and Paul have talked about in the comments, please let me know and I can re-review.

dean-krueger:
> This appears to need a rebase for the CHANGELOG... and wonder if you've looked at this[ pre-existing progressbar full of options/features](https://github.com/p-ranav/indicators), and whether it makes sense to just use it?
> 
> We can just copy the files into our repo (per the license). One benefit, in theory, is that if it is improved in the future, we can take advantage of it by just dropping in an updated copy.

Can rebase for sure. I was, after coming back to this after a little while, preparing my best argument about why that other pre-built progress bar would be a little extra work to integrate (probably, at least by virtue of it not already being integrated), but it IS pretty cool and I feel like I've been convinced that we should use it instead of the one I had Cursor throw together. I'll go ahead and (maybe this morning before class) take a swing at dropping it in. Some of the features would be really nice (namely the elapsed time one).

abachma2:
Where are we at with this PR? Did you ever get the last bit done @dean-krueger?

dean-krueger:
No, sadly this one died on the vine a bit as more important things came up and then classes started again. If there's still interest I think it'd be a good thing to add, still, but haven't had the time to come back around to it. Perhaps once finals are over and my time becomes less crunched I can try to push this over the finish line a little bit here and there while Paul is reviewing something else (or if I start to go crosseyed working on the big ticket items).

abachma2:
I understand. Keeping this as a side project is fine. I would like to see this reach the finish line. This is a feature I would like to use.

dean-krueger:
Okay, I had a weird hour-ish today between finishing a bigger project and going home so I tried to tackle this. All this new version does is get rid of the custom progress bar cursor made and replaces it with the pre-packaged one @gonuke found. Otherwise it works more or less the same as before. It prints out the progress as a bar, a % completion, and the number of timesteps that have been run every 10 time steps, and if you do the -v flag above 2 it stops printing the progress bar because that flag makes CYCLUS print the time anyway. I've included a little gif of it working, for your viewing pleasure. Also worth noting: the include for that prepackaged progress bar is like 3000+ lines of code, which is why the diff on this PR exploded. Do we need all those features? Probably not, but I agree with Paul that it's nice to have the option and this means if there's ever a big update in progress bar technology we can presumably just drag and drop the new file in.

Screen Recording 2026-05-05 at 5 06 16 PM

src/timer.cc
```diff
@@ -53,13 +94,48 @@ void Timer::RunSim() {
...
       break;
     }
   }
 
+  // Finalize progress bar: leave the cursor on a new line. Only redraw if we
+  // did not already print 100% in the loop.
+  if (progress_bar_) {
+    const int completed_steps =
+        std::min(time_ - progress_origin_, progress_span_);
+    const int pc = std::max(completed_steps, 0);
```
gonuke: I notice this is slightly different from the math above for `capped` - is that intentional?

Can we streamline this into a function that returns the progress value?

src/timer.cc
```diff
@@ -16,6 +18,34 @@
 
 namespace cyclus {
 
+namespace {
+// Simulation time is int throughout Cyclus; indicators::ProgressBar uses size_t
+// for set_progress / max_progress only because that is the library API.
+inline size_t IndicatorProgress(int n) {
+  return static_cast<size_t>(std::max(n, 0));
```
gonuke: I notice that each time you use this, you've already tested it with `max`... maybe a member function of `Timer` could just return the current progress value having done all those tests?

src/timer.cc
```diff
@@ -16,6 +18,34 @@
...
+}
+
+bool ProgressBarEnabled() {
+  const char* env_var = std::getenv("CYCLUS_PROGRESS_BAR");
+  if (env_var) {
+    std::string val(env_var);
+    return !(val == "0" || val == "false" || val == "no" || val == "off");
```
gonuke: Does this mean that the environment variable overrides the logging check?

gonuke: How do we turn on the progress bar without the environment variable? I.e. how do we ever get to the test vs logging?

src/timer.cc
```diff
@@ -53,13 +94,48 @@ void Timer::RunSim() {
     EventLoop();
 #endif
 
+    // Progress reflects timesteps completed this iteration (loop uses
+    // time_ in [progress_origin_, si_.duration), so the last step reaches
+    // progress_span_ / progress_span_).
+    if (progress_bar_) {
+      const int completed_now = time_ - progress_origin_ + 1;
```
gonuke: If you move this after the next line that increments the timer, can you avoid the "+1" here?  And then it becomes the same calculation as below.

src/timer.cc
```diff
@@ -16,6 +18,34 @@
 
 namespace cyclus {
 
+namespace {
+// Simulation time is int throughout Cyclus; indicators::ProgressBar uses size_t
+// for set_progress / max_progress only because that is the library API.
+inline size_t IndicatorProgress(int n) {
```
gonuke: I think it makes sense for these functions to be member functions of `Timer`

gonuke (changes requested):
I think I see a few ways to streamline this...

src/timer.cc
dean-krueger: 1. How does the progress bar get turned on if you don't set the environment variable?

The progress bar defaults to "on" unless:
- You set an environment variable on your machine called `CYCLUS_PROGRESS_BAR` and set it to `off`, `no`, `0`, or `false`  (case sensitive, in this case, which probably isn't ideal but also this is pretty edge-case so maybe that's fine)
- You use the `-v` flag in cyclus with an argument greater than that assigned to `LEV_WARN` (I think this is `2`, but it may be `3`) because that's the point where cyclus just prints the timestep anyway.

2. Does setting the environment variable mean that ProgressBarEnabled() won't check the verbosity?

Short answer: yes. 

Long Answer: the first thing the `ProgressBarEnabled()` does is check if you have that specific environment variable (you likely don't), and then if you DO, AND it's set to one of those things that turns it off, it doesn't print the bar. If you DON'T, then it only cares about what verbosity Cyclus was told to use. If you set the environment variable to not print the bar, you probably don't want to then, conditionally, turn it on based on your `-v` level.

src/timer.cc
dean-krueger: You're talking about `IndicatorProgress`, `ProgressBarEnabled` and `ProgressUpdateFrequency`? If so, sure. They're isolated like this because Cursor figured they were more "helper functions the Timer would want to use" than "Things that define what a Timer is", but I don't feel too strongly about that.

src/timer.cc
dean-krueger: I think it may also make sense to just get rid of this function and do a static_cast each time, since that's sort of all it does. That way each place where we clamp the value we were passing in stays the same and the logic is right there for the reader. If you'd still prefer the other way let me know, but for now I'm going to just do that.

src/timer.cc
dean-krueger: Yes.

src/timer.cc
dean-krueger: okay, after looking at a different comment I see the use for a little function. Nevermind all that above.

dean-krueger:
Comments responded to! Should be good for another look. The tests all pass on my end, and the progress bar looks normal.

src/timer.cc
```diff
@@ -29,6 +31,17 @@ void Timer::RunSim() {
...
   ExchangeManager<Product> genrsrc_manager(ctx_);
+
+  if (!progress_bar_ && ProgressBarEnabled()) {
+    progress_span_ = std::max(si_.duration - time_, 1);
+    progress_bar_.reset(new indicators::ProgressBar{
+        indicators::option::BarWidth{50},
+        indicators::option::MaxProgress{ProgressValue(progress_span_)},
+        indicators::option::ShowPercentage{true},
+    });
+    progress_update_frequency_ = ProgressUpdateFrequency(si_.duration);
```
gonuke: This is happening nearly the same way in two places.  The actual lines of code are slightly different, but I think they are actually doing the same thing.  Can we make a function to `SetupProgressBar()`?

src/timer.cc
```diff
@@ -55,11 +68,43 @@ void Timer::RunSim() {
...
+      const int completed_now = time_ - progress_origin_;
+      if (completed_now % progress_update_frequency_ == 0 ||
+          completed_now == progress_span_) {
+        // Postfix must be set before set_progress: set_option does not redraw,
+        // but set_progress calls print_progress() which reads postfix_text.
+        progress_bar_->set_option(
+            indicators::option::PostfixText{
+                " (" + std::to_string(completed_now) + "/" +
+                std::to_string(progress_span_) + ")"});
+        progress_bar_->set_progress(ProgressValue(completed_now));
```
gonuke: This is happening nearly the same way in two places - the lines of code are a little different, but other than the conditional, they seem to be doing the same thing.  Can we write a function to `UpdateProgressBar(completed_now)`?

src/timer.cc
```diff
@@ -285,12 +332,60 @@ void Timer::Initialize(Context* ctx, SimInfo si) {
   if (si.branch_time > -1) {
     time_ = si.branch_time;
   }
+
+  progress_origin_ = time_;
+  progress_span_ = std::max(si.duration - progress_origin_, 1);
+
+  progress_bar_.reset();
```
gonuke, line 317: Do we need to reset here - we're about to reset again anyway.

src/timer.cc
```diff
@@ -55,11 +68,43 @@ void Timer::RunSim() {
...
+    }
+
+
     if (want_kill_) {
       break;
     }
   }
 
+  // Finalize progress bar: leave the cursor on a new line. Only redraw if we
+  // did not already print 100% in the loop.
```
gonuke: Is there a reason to not just redraw always? I assume it's not expensive?

src/timer.cc
gonuke, line 317: Do we need to wrap this in a conditional of `if (progress_bar_)`?

gonuke (changes requested):
As I review this again, I wonder if we can encapsulate our use of this progress bar into a few member functions?
```
void setupProgressBar();
// set the variables and reset the progress bar

bool updateProgressBar();
// determine whether the time is right to do an update

void redrawProgressBar();
// perform the steps to redraw the bar
```

Part of the motivation is code reuse and consistency, and the other is for cognitive load where it has been inserted in the previous code.

src/timer.cc
dean-krueger: I think this should be fixed with the refactor (coming shortly)

src/timer.cc
dean-krueger: I think this is also being addressed in the refactor (coming soon)

src/timer.cc
dean-krueger: I had to ask about this one, but it seems like the first one is in case the progress bar is disabled (catch any bars that might still be hanging around from earlier), and the second one is a different call that resets the bar "normally" when the bars are enabled. If the bar is enabled, then the first one is redundant, but if, say, you ran with a bar and then ran a verbose job (the bar "shouldn't" print) you'd still have the progress bar from earlier floating around.

src/timer.cc
dean-krueger: If I remember correctly there was something added to prevent the bar from drawing twice, which would happen at the very end. My guess is that this is that.

src/timer.cc
dean-krueger: No. This is in Initialize, so we want it to be able to set up the bar for the first time I'm pretty sure.

dean-krueger:
Since this started out as a Cursor job, I had Codex handle the refactor. I found it easier to read this way, though, for sure, so good suggestion. The functions are nice.

src/timer.cc
```diff
@@ -29,6 +31,11 @@ void Timer::RunSim() {
 
   ExchangeManager<Material> matl_manager(ctx_);
   ExchangeManager<Product> genrsrc_manager(ctx_);
+
+  if (!progress_bar_) {
```
gonuke: We're a little inconsistent with when we test this and/or when we test whether progress bar is enabled...

src/timer.cc
gonuke: It's still not clear to me what the (negative) consequence of drawing it twice is?  Doesn't it overwrite itself anyway?

src/timer.cc
```diff
@@ -270,6 +293,8 @@ void Timer::Reset() {
   build_queue_.clear();
   decom_queue_.clear();
   si_ = SimInfo(0);
+
```
gonuke, line 283: If we haven't initialized `progress_bar_` is it safe to try and reset it? Don't we need to make this conditional on whether progress bar is set?

src/timer.cc
gonuke: In our paradigm, is it possible to have an earlier progress bar floating around?

gonuke (changes requested):
I think your "collaborators" are making this more complicated than it needs to be... and it's not clear that we understand it all well enough to make better decisions, other than what it tells us. 

I see this as a pretty simple set of additions that we are not invoking consistently such that it will be confusing in the long term.

src/timer.cc
dean-krueger: Yep! 100% safe. Since progress_bar_ is an `std::unique_ptr` we're just calling `std::unique_ptr::reset()` here, which explicitly safely handles the case where it's null (apparently). If I'm being totally honest this is a little bit over my head, but [this](https://stackoverflow.com/questions/21606073/stdunique-ptrreset-checks-for-managed-pointer-nullity) stack overflow conversation plus the AI overview answer to googling "is calling reset() on a null std::unique_ptr safe?" has convinced me that this is fine.

dean-krueger:
> In our paradigm, is it possible to have an earlier progress bar floating around?

If you do a restart I think so, yeah. Like this:

```
cyclus --restart some_file.sqlite
```

I don't use this feature, and so I don't REALLY understand how it works, but it seems like this is a quick safety check for an edge case. I can get rid of it if you'd prefer, but it's possible that doing so could cause a bug in 10 years once someone else wants to restart a bunch of stuff (albeit a fairly minor terminal-aesthetic bug, I don't think it would affect results or anything).

If you're just running new simulations in the terminal with `cyclus input.xml` then this line is probably redundant.

src/timer.cc
dean-krueger: Okay, I took a step back from my "collaborators" and did some testing, specifically around whether the extra draw at the end of the simulation was necessary, since that seemed to be the part most "inconsistent". Commenting it out didn't change the display of the progress bar in my "normal" simulations, so I figured it was safe to delete it. Before I did, I looped Codex back in and had a conversation about what it was doing, and it appears that Cursor/Codex added/kept is because the Timer API has a function `KillSim()` which an agent can invoke to stop the simulation part-way through. If this happens, the progress bar might display weird, but thinking "I've never interacted with this..." I asked Codex to trace it and it basically came back and said "Yeah nothing in Cyclus or Cycamore actually calls this ever". It's POSSIBLE that some mysterious third party developer could call this, or that in the future we might want to call it for some reason, but if that happened the worst thing that would come of it would be that the progress bar might erroneously say " 1510/2000" when it was actually "1516/2000" and then the cursor would be on the same line as the bar. Because this seemed like a really minor issue for a feature that nobody I know uses, and the code that enabled it was legitimately a little confounding since it didn't seem necessary, I removed it. The flow of the progress bar in Timer is now:

```
if no progress bar:
   SetupProgressBar()

while (time_ < sim.dur):
   time_++
   //ShouldUpdate tests progress_bar_ != NULL
   if (ShouldUpdateProgressBar(timestep):
      RedrawProgressBar(timestep)
```
 
 which seems fairly straightforward to me. I also responded to your other comments. Hopefully this is sort of in-line with what you were looking for. It's a good reminder that we have some "extra" hooks in Cyclus that an AI could get hung up on aiming for "completeness" when they aren't REALLY necessary. Commit incoming.

src/timer.cc
```diff
@@ -285,12 +299,83 @@ void Timer::Initialize(Context* ctx, SimInfo si) {
...
+  return cyclus::Logger::ReportLevel() <= cyclus::LEV_WARN;
+}
+
+void Timer::SetupProgressBar() {
+  progress_origin_ = time_;
+  progress_span_ = std::max(si_.duration - progress_origin_, 1);
+
+  progress_bar_.reset();
+  progress_update_frequency_ = ProgressUpdateFrequency(si_.duration);
+  if (ProgressBarEnabled()) {
```
gonuke: Let's do this test outside the function and only call the function if true.

src/timer.cc
```diff
@@ -285,12 +299,83 @@ void Timer::Initialize(Context* ctx, SimInfo si) {
...
+void Timer::RedrawProgressBar(int completed_steps) {
+  const size_t progress = ProgressValue(completed_steps);
+  // Postfix must be set before set_progress: set_option does not redraw,
+  // but set_progress calls print_progress() which reads postfix_text.
+  progress_bar_->set_option(
+      indicators::option::PostfixText{
+          " (" + std::to_string(progress) + "/" +
+          std::to_string(progress_span_) + ")"});
+  progress_bar_->set_progress(progress);
+}
```
gonuke: Combine these two functions and access `time_` and `progress_origin_` directly without any parameters.

src/timer.cc
```diff
@@ -285,12 +299,83 @@ void Timer::Initialize(Context* ctx, SimInfo si) {
...
+size_t Timer::ProgressValue(int completed_steps) {
+  return static_cast<size_t>(
+      std::min(std::max(completed_steps, 0), progress_span_));
 }
 
 int Timer::dur() {
   return si_.duration;
 }
 
+Timer::Timer()
```
gonuke: replace this with default values in header.

src/timer.cc
```diff
@@ -285,12 +299,83 @@ void Timer::Initialize(Context* ctx, SimInfo si) {
...
+}
+
+int Timer::ProgressUpdateFrequency(int duration) {
+  if (duration > 100) {
+    return 10;
+  } else if (duration > 20) {
+    return 5;
+  }
+  return 1;
+}
```
gonuke, line 346: Make this update approximately every 1% or every timestep, which ever is longer.

gonuke (changes requested):
a little more consolidation

dean-krueger:
I got another round of (slightly quick and dirty) changes based on the conversation we had @gonuke. I THINK they correctly address your concerns, but if you see things that look wrong let me know and I will take more time with the next one. I was mostly hoping to get this into 1.6.1

src/timer.cc
```diff
@@ -287,6 +299,57 @@ void Timer::Initialize(Context* ctx, SimInfo si) {
...
+
+void Timer::SetupProgressBar() {
+  progress_origin_ = time_;
+  progress_span_ = std::max(si_.duration - progress_origin_, 1);
+
+  progress_bar_.reset();
+  progress_update_frequency_ = ProgressUpdateFrequency(si_.duration);
+  progress_bar_.reset(new indicators::ProgressBar{
+      indicators::option::BarWidth{50},
+      indicators::option::MaxProgress{ProgressValue(progress_span_)},
```
gonuke, line 321: This will always be just `progress_span_` so why not?


```suggestion
      indicators::option::MaxProgress{progress_span_},
```

src/timer.cc
```diff
@@ -287,6 +299,57 @@ void Timer::Initialize(Context* ctx, SimInfo si) {
...
+}
+
+void Timer::RedrawProgressBar() {
+  int completed_steps = time_ - progress_origin_;
+
+  if (progress_bar_ &&
+      (completed_steps % progress_update_frequency_ == 0 ||
+      completed_steps == progress_span_)) {
+
+    const size_t progress = ProgressValue(completed_steps);
```
gonuke, line 333: This may become the only place that it's being called, so you could just include the code for `ProgressValue` here

gonuke (changes requested):
More streamlining and reduction of functions

src/timer.cc
dean-krueger: because the function needs a `size_t` and `progress_span_` is normally an int (and we want it to be I think)

src/timer.cc
dean-krueger: I think we want it to be called other places because it also converts things to a `size_t` so that function from the progress bar include accepts it.

gonuke (approved):
LGTM - we'll go with this...

PR #1918, opened 2025-10-02
force both materials to decay on absorption

### LINKED ISSUE #1841: Bulk ResBufs have dubious decay behavior
A new "bulk" feature was added to `ResBuf` that forces all materials to be squashed into a single material object when they are pushed into the `ResBuf` (#1687).  However, because of the behavior when materials are squashed, this can cause problems when decay is important.

Specifically, in src/material.cc#L118 the `prev_decay_time_` is used from the larger of the two materials regardless of how large each material is or how different those decay times are.  This should be reconsidered for squashing materials in general, and for the bulk storage `ResBuf` in specific.

One option would be to force the `ResBuf` to decay both material objects upon the push.

gonuke:
# Summary of Changes
Force both materials involved in an absorption to decay before absorption.

## Design Notes
Previous behavior would result in dubious decay times for materials following absorption

**Importantly**, this PR introduces a new constraint on material decay:
* tracked materials can only decay forward and up to the current context time

- [X] API for Cyclus modules
- [X] Input files

# Related CEPs and Issues
Fixes #1841

# Associated Developers
@dean-krueger 

# Testing and Validation

- [ ] I have compiled and run the code locally
- [X] I have updated the changelog


dean-krueger:
The UnitValue test seems to be failing because it looks like you accidentally took out the unit-value averaging lines of the absorb function. Adding those back now and then it should (hopefully) work!

dean-krueger:
PR submitted to @gonuke's branch which, if accepted, should come here and hopefully tests pass after that.

dean-krueger:
@gonuke has brought up some good points about more careful testing that needs to happen now that this change has been implemented. They are:

> Regarding the context:
> 
> - In the changes that I proposed in my PR https://github.com/cyclus/cyclus/pull/1918, there is the possibility for absorb to happen for materials that are not governed by a context.
> 
> - the previous behavior of Material::Absorb() didn't care about a context because it only picked the decay time of one of the two materials.
> 
> - I added a check for the context so that we could advance the decay time to the current context time, rather than only the times in the materials.
> 
> - I can't remember all the ways that we might arrive at materials being outside of a context, but most of the tests for decay throughout the material tests do not assume a context.
> 
> - if we only want to perform Absorb() on materials in a context, then Material::Absorb() needs to be updated to check that it's true and fail/warn otherwise.
> 
> - if we want to allow Absorb() for materials not in a context, then we should include tests for that.

> Regarding backward decay:
> 
> - I don't think it's every been considered a normal behavior
> 
> - if we always want to force decay to move forward we need to not just use the context time, but take the max of the previous decay times and the context time to determine the new time
> 
> - we can also fail/warn if the previous decay times are ever beyond the current context time
> 
> - if we leave it as is, it's possible that we cause decay to go backwards
> 
> - whatever we choose, we need to test it appropriately

We (likely me) should look into adding these tests to improve our coverage.

dean-krueger:
Checks have been added to my PR into @gonuke's PR addressing the above. Hopefully those go through, and this can move forward, but obviously if I did something funky let me know and I can address it.

dean-krueger:
Now that gone/cyclus/#2 has been merged, I THINK the main thing on this is fixing the Cycamore tests. Let me know if you see other stuff that needs addressing.

Another thought I was having while making the tests work yesterday was that when we decay outside of a context we give it a number of months to decay (not time steps, since it's not in a context). It MIGHT make more sense to do it in seconds and then make people who want months use `N * cyclusMonth` as the argument so that you can do any timestep.

dean-krueger:
Okay, the giant wave of failing Cycamore tests I saw previously seems to have been related to the `!= !mat->HasContext` thing, but there was still one Cycamore test which failed. I've made a [new PR](https://github.com/gonuke/cyclus/pull/4/) into this PR that resolves it. Hopefully that ends up being enough to push this over the finish line! Give that a look whenever you get a chance @gonuke

gonuke:
Replaced by #1966

PR #1927, opened 2025-12-06
Quiet Tests

### LINKED ISSUE #1916: [Task] The `IntegTests/IntegTestsFixture.RunAllInfiles` test in the Cyclus Unit Tests is Incredibly Verbose
# What is this task about?

Something needs to be done to make this test only output when it fails, or something similar.

# Why is this task needed?

Every time I run the cyclus unit tests and one of them fails I have to scroll past a TON of output from one test in particular, which is super annoying. It's not HARD to get past it, but it's a lot of scrolling and I don't know that all that output every time is really necessary

This is maybe 1/8 of the output with the text set to the smallest possible setting in my terminal. You can see (if you look really hard) the test start at the top, and it doesn't finish by the bottom (it goes on for a long time after that, I wish I could capture my scroll bar in the screenshot to show it properly)
Image

By making that test not give all that output when it's passing.

dean-krueger:
# Summary of Changes

Frequent testers, Rejoice! This PR removes almost all of the overly verbose cout from the `cyclus_unit_tests`, both cleaning up the way it looks, and *dramatically* reducing the amount of scrolling up you need to do to find details about that one test you're failing.

## Design Notes

The primary two tests that were printing a lot to the terminal during the unit tests were:

- `IntegTests/IntegTestsFixture.RunAllInfiles/0`: responsible for maybe >95% of all system output during the entire test including all the GTEST output. This test was printing all the verbose coin output for every single input file it solved as a part of its test. This output, to my knowledge, served no purpose besides making me really sad when I had to scroll past it. Fixed by adding a bool `IsQuiet` to the context timer which can be seen by the `prog_solver` to make a new flag `actually_verbose` which if false (if TimerIsQuiet) will suppress the verbose output.
- `.../AgentTests.Schema/0`: a few smaller, but still quite large outputs. This test validates the Schemas used by agents, and it seems like there was a debug line (a `std::cout`) that printed the entire schema (without using that output) which never got removed. I removed this line.

Decided this made more sense as a V1.6.# change than a v2.0 change (mostly because I want to use it asap)

- [x] API for Cyclus modules
- [x] Input files

# Related CEPs and Issues

Fixes #1916 

# Associated Developers

🖥️ Cursor AI 

# Testing and Validation

This PR was all about tests! I basically just clean-built and ran the tests a bunch to make sure they all still behaved. No additional tests were added.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,7 +34,7 @@ Since last release
 * Allow multiple archetype blocks to facilitate includes (#1874)
 
 **Changed:**
-
+* Made the Unit Tests far less verbose by suppressing log output during RunSim (#1927)
 * Changed Dockerfile to use boost and boost-cpp instead of libboost-devel (#1906)
 * Changed TradeExecutor to use adjusted preferences from ExchangeContext (#1897)
 * Ran clang-format on src directory (#1881, #1893)


dean-krueger:
> I definitely like the idea of this, but this makes it hard to turn back on that output if tests are failing and a developer wants to monitor the output.
> 
> Is there a way to do this that involves passing in a command-line variable to tests?? I don't know much about how the google test framework would support that.

I agree. When we fail a test we do still get error message output:

Screenshot 2025-12-11 at 1 17 03 PM


but if something were to go wrong with coin I suppose it might be important to be able to see that output somehow? Perhaps an optional --verbose flag to the `cyclus_unit_tests` command to turn that back on would be the right move? I definitely think the tests should be quiet by default, though.

I'll toss this on my "Side Projects Saturday" docket and see where I get.

gonuke (approved):
This looks good to me.  Let's go ahead and make a new issue for a feature that would allow verbosity to be resumed with a command-line option of some kind.

Issue #1929, opened 2025-12-10
[Task] Rework `RIF_cost.cycpp.h` to enable marginal cost


dean-krueger:
# What is this task about?

We currently give LPC, but what I've decided we want ultimately is marginal cost (fixed cost plus materials) for a single unit. Mostly `facility_cost.cycpp.h` needs to be reworked to have a function that returns this. While in there I may also clean some stuff up. Not sure.

# Why is this task needed?

The surplus model I'm trying to add to Cyclus wants to $\underset{x}{min} \space (MC_a - MU_a) x_a$, and to get that we need to give it MC (and also MU but that's a different issue), not LPC (which is still useful for bookkeeping later)

# Relevant resources or references

[This document](https://www.overleaf.com/project/6880ee132697998dfdcbdbf8) on the old Cyclus Financial Model
[This link](https://github.com/cnerg/CFER/issues/9#issuecomment-3619365502) to a comment I left on my implementation meta-issue.

# How can this issue be closed?

Satisfactorily enabling MC as an output of agents via code injection.

dean-krueger:
I think we're taking a step away from welfare for now.

PR #1932, opened 2025-12-23
Marginal Utility Injection

### LINKED ISSUE #1924: [Feature] Add Code Injection for Marginal Utility

I want to add a welfare surplus maximization model to the DRE (as an optional mode, see #1923). This mode does something like `min (MC - MU) x` where `MC` is "marginal cost" (enabled by the cost code injection I did earlier) and `MU` is marginal utility.

I want to give facilities a marginal utility for commodities via code injection, similar to the cost code injection work I did earlier.

- Hard Coding: repeat code everywhere, code injection much better
- Different model: I've thought of different models extensively, this one seems like the "best" for [adding economic formality to cyclus](https://docs.google.com/presentation/d/1auJ_7Kzn-eHXis4gtILsne1m-sj_3qXWupuW_jNNevM/edit?usp=sharing)
- regular utility: We want MU directly, not u(x). Seems easier to ask for it than to ask for something else (which could break cyclus) and then do a transformation. 

# Additional context.

Asking for the functional form of MU and the relevant parameters. There has been some discussion [here](https://github.com/cnerg/CFER/issues/9) about specifics.

By making a code-injection marginal utility functional form input for facilities

dean-krueger:
# Summary of Changes

For surplus maximization we need marginal cost (MC) and marginal utility (MU). #1929 handled the MC part, and this PR should handle the MU half. 

## Design Notes

Similar to the MC code, this is done through code injection using the `.cycpp.h` format. It was decided early on that we should ask users for the functional form of MU directly, instead of asking for utility functions and then deriving the MU from there. It was also decided that we should limit the functional form of MU to a few specific functions, since these are the ones which will work best with the DRE. Finally, piecewise was originally a functional form we planned to implement, but it was uniquely challenging compared to the other options, and there wasn't a clear need for it at the moment, so I decided to kick that can down the road for when someone actually wants to use piecewise utility.

- [x] API for Cyclus modules
- [x] Input files

# Related CEPs and Issues

Closes #1924 

# Associated Developers

Cursor AI (Auto) 🖥️

# Testing and Validation

Cyclus was built and the unit tests were run. My plan is to get all the necessary pieces into the v2.0 branch and then do more thorough testing once they were all there. This applies to writing unit tests, too, since it is possible (even likely) that some of the actual implementation will change enough that writing and then rewriting tests would be a bit of a time waste.

### DIFF
--- a/src/toolkit/marginal_utility.cycpp.h
+++ b/src/toolkit/marginal_utility.cycpp.h
@@ -1,0 +1,52 @@
+/// This includes the required header to add marginal utility functional forms
+/// to archetypes.
+/// One should only need to:
+/// - '#include "toolkit/marginal_utility.cycpp.h"' in the header of the
+///    archetype class (as private)
+/// - The functional form and parameters will be available as state variables
+///    that can be set in the input file.
+
+/// How to add parameters to this file:
+/// 1. Add the pragma. A default value MUST be added to ensure backwards
+///    compatibility.
+/// 2. Add "std::vector<int> cycpp_shape_<param_name> = {0};" to the end of the
+///    file with the other ones, replacing <param_name> with the name you put
+///    in the pragma (must match exactly).
+
+// clang-format off
+#pragma cyclus var { \
+    "default": "Linear", \
+    "uilabel": "Marginal Utility Functional Form. Must be one of: Linear, Affine, Exponential, Logarithmic, Piecewise", \
+    "uitype": "combobox", \
+    "categorical": ["Linear", "Affine", "Exponential", "Logarithmic", "Piecewise"], \
+    "doc": "The functional form to use for calculating marginal utility. Must be one of: Linear, Affine, Exponential, Logarithmic, Piecewise" \
+    }
+std::string mu_functional_form;
+
+#pragma cyclus var { \
+    "default": [1.0], \
+    "uilabel": "Marginal Utility Parameters", \
+    "doc": "Parameters for the marginal utility functional form. The number and meaning of parameters depends on the selected functional form." \
+    }
+std::vector<double> mu_parameters;
+// clang-format on
+
+/// @brief Calculates the marginal utility based on the selected functional form
+/// and parameters.
+///
+/// This function will be implemented in Step 3.
+///
+/// @param commods Vector of commodity names
+/// @param prefs Vector of preference values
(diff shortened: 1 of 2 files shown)


gonuke:
This has a bunch of the "quieting" stuff in it and appears to need a rebase

dean-krueger:
> This has a bunch of the "quieting" stuff in it and appears to need a rebase

Oh snap, yeah it does. I'll do that asap.

dean-krueger:
Done! @gonuke . Should be ready for another look. Sorry about that!

src/toolkit/marginal_utility.cycpp.h
```diff
@@ -0,0 +1,138 @@
...
+std::map<std::string, double> mu_cache_;
+
+/// @brief Calculates the marginal utility based on the selected functional form
+/// and parameters.
+///
+/// Functional forms:
+/// - Linear: MU = k * P (requires 1 parameter: k)
+/// - Affine: MU = b + k * P (requires 2 parameters: b, k)
+/// - Exponential: MU = B * exp(k * P) (requires 2 parameters: B, k)
+/// - Logarithmic: MU = B * ln(1 + k * P) (requires 2 parameters: B, k)
```
gonuke, line 48: So I think I understand the intent here... the independent variable in these functional forms is the user supplied preference. So a user could just calculate the results of these functional modifications and provide them with a linear of k=1 and get the same result.  Thus, these functions are just providing shortcuts to users, right?

dean-krueger: Yeah, I suppose that's true. Especially in a world where facilities are only realistically choosing between 1-3 goods that would be a pretty manageable way to get MU. I expect that ~90% of the time the functional form of choice will be linear, too, so it's really more about encouraging users to think of a consistent way to translate the abstract notion of their "preference" for something to a utility in the correct units (like we talked about way back when in that meeting where we first brought this up).

gonuke:
Did you look at the [symbolic function factories](https://github.com/cyclus/cyclus/blob/aa9f257b96ecb921a31fcb9cabfeb6b1df1cd9ec/src/toolkit/symbolic_function_factories.h#L67)

dean-krueger:
I believe we're moving away from this. Closing this PR.

PR #1937, opened 2026-01-13
Cyclus v2.0 DRE Overhaul

dean-krueger:
# Summary of Changes

This PR replaces the Dynamic Resource Exchange’s preference-based objective with a direct cost-based objective.

- Replaces per-arc `pref` values with:
  - `unit_cost` on bids,
  - `unit_cost_mod` on requests, and
  - an `arc_cost` used directly by solvers.
- Defines the default arc cost as `unit_cost + unit_cost_mod`, with defaults of `1.0` and `0.0`, respectively.
- Renames the adjustment hooks from `Adjust*Prefs` to `Adjust*Params`, and replaces `PrefMap` with `RequestBidMap`. Adjustment hooks now modify the per-request/bid arc-cost map; erasing an entry removes that arc from the exchange graph.
- Updates the exchange translator, graph, solvers, debug output, transaction records, and Python bindings to use the new terminology and data flow.
- Updates `MatlSellPolicy` so its bid cost is based on a material’s `unit_value` plus an explicitly configured policy unit cost. It now validates unset and non-finite costs.
- Reworks solver behavior for direct costs:
  - optimization solvers minimize `arc_cost`;
  - faux unmet-demand arcs receive a penalty strictly worse than real arcs, including zero- and negative-cost cases;
  - the greedy solver and preconditioner prioritize lower-cost arcs and requests, including negative costs.
- Replaces the old marginal-cost code-injection header with `unit_cost.cycpp.h`.
- Expands and updates C++ tests for arc removal, cost translation/recording, zero- and negative-cost solver behavior, greedy ordering, and sell-policy cost validation.

## Design Notes

The central design change is that an arc no longer represents an inverted preference. Instead, it represents a cost that can be minimized directly.

A bid supplies its base `unit_cost`, while a request supplies a `unit_cost_mod`. The exchange context initializes each request/bid pair with their sum, and the adjustment phase can then override that pair’s `arc_cost` or remove the pair entirely. The translated graph retains the bid and request components for traceability, while solvers use the adjusted `arc_cost`.

This avoids the old preference inversion and shift-based handling, makes zero and negative costs meaningful, and keeps the optimization and greedy solvers aligned around the same “lower cost is better” interpretation. The faux-arc penalty is computed from real effective arc costs so feasible real trades remain preferable to unmet demand.

The greedy preconditioner still combines commodity importance with exchange information, but now uses a finite monotonic transform of average arc cost. This preserves the intended ordering for zero, `-1`, and costs below `-1`, where the old preference-era transform was not valid.

This is an API-breaking terminology change: archetypes that previously overrode `AdjustMatlPrefs` or `AdjustProductPrefs` must use `AdjustMatlParams` or `AdjustProductParams` and operate on the new request/bid cost map.

# Check the box if your change does not break any of the following:
- [ ] API for Cyclus modules
- [ ] Input files

We break all three here, I'm pretty sure.

# Associated Developers

🖥️ Cursor AI (auto), Codex 5.6 Terra, Codex 5.6 Sol, Claude Code Sonnet/Opus

# Testing and Validation

Cyclus Unit Tests were updated and all pass. Autonuke run verified test accuracy and basic coverage of new features. Many new tests were added. Note: this still breaks Cycamore, and so that will need to get fixed in its own v2.0-dev branch.


src/trade_executor.h
```diff
@@ -105,43 +105,51 @@ template <class T> class TradeExecutor {
...
-
-          // If the bid has NaN preference, use the request preference
-          if (std::isnan(original_preference)) {
-            original_preference = trade.request->preference();
+          // Get original MC and MU
+          double original_mc = trade.bid->preference();
+          if (std::isnan(original_mc)) {
+            original_mc = 0.0;  // NaN means no cost
           }
+          double original_mu = trade.request->preference();
```
gonuke, line 113: Is there a reason that we don't do the same check for `nan` on the MU?  Or is this just backwards compatibility on the MC side?

src/prog_translator.cc
```diff
@@ -155,8 +155,14 @@ void ProgTranslator::XlateGrp_(ExchangeNodeGroup* grp, bool request) {
       }
 
       if (request) {
-        CheckPref(a.pref());
-        ctx_.obj_coeffs[arc_id] = ExchangeSolver::Cost(a, excl_);
+        // MC must be >= 0 (can be 0 for free resources)
```
gonuke: Is there a reason we can't retool `CheckPref` for this?  If not, can we delete it?

src/greedy_solver.h
```diff
@@ -20,31 +20,43 @@ void Capacity(boost::shared_ptr<cyclus::ExchangeNode>, cyclus::Arc const&,
...
-  return (lpref != rpref) ? (lpref > rpref) : (lid > rid);
-}
+/// @brief A comparison functor for sorting a container of Arcs by arc weight,
+/// in ascending order (i.e., lowest cost Arc first). In the case of a tie, a
+/// lexicalgraphic ordering of node ids is used.
+/// Note: Lower arc weight (MC - MU + shift) is better, so we sort ascending.
+struct ReqPrefComp {
+  // Note: shift_ is no longer needed since arc weight is stored in pref()
+  // Keeping for potential future use, but not using it
+  ReqPrefComp(double shift) {}
```
gonuke: I see you are keeping this below, but is there a need to keep it here right now?

src/greedy_solver.cc
```diff
@@ -152,7 +152,8 @@ void GreedySolver::GreedilySatisfySet(RequestGroup::Ptr prs) {
     if (graph_->node_arc_map().count(*req_it) > 0) {
       const std::vector<Arc>& arcs = graph_->node_arc_map().at(*req_it);
       sorted = std::vector<Arc>(arcs);  // make a copy for now
-      std::stable_sort(sorted.begin(), sorted.end(), ReqPrefComp);
+      // ReqPrefComp no longer needs shift since it uses arc.pref() directly
+      std::stable_sort(sorted.begin(), sorted.end(), ReqPrefComp(0.0));
```
gonuke: If you keep `shift` in the argument list for this functor, maybe you should pass in `shift_` here, still?

src/greedy_preconditioner.cc
gonuke: Do you still need this?  Or could you retool this to use it in the same way the previous code did, with `std::accummulate`?

src/exchange_graph.h
```diff
@@ -81,6 +81,9 @@ class Arc {
     vnode_ = other.vnode();
     exclusive_ = other.exclusive();
     excl_val_ = other.excl_val();
+    mc_ = other.mc();
+    mu_ = other.mu();
+    pref_ = other.pref();
```
gonuke, line 86: For reasons that are not immediately obvious to me, we did not previously copy the `pref_` with this assignment operator.  However, I notice that we did copy `pref_` with the copy constructor.  I only mention this to be sure that there is not somewhere that expects assignment to NOT copy these??

src/exchange_graph.h
```diff
@@ -97,14 +100,31 @@ class Arc {
...
+  
+  /// @brief marginal utility (from request node)
+  inline double mu() const { return mu_; }
+  inline void mu(double mu) { mu_ = mu; }
+  
+  /// @brief returns the arc weight (MC - MU + shift) used in the objective function
+  /// This is computed and stored after all arcs are created and shift is known.
+  /// For the individual components, use mc() and mu() instead.
   inline double pref() const { return pref_; }
+  /// @brief sets the arc weight (MC - MU + shift)
```
gonuke: But this is not what it does... it seems to allow it to be set arbitrarily.

src/exchange_graph.h
gonuke, line 46: Does an `ExchangeNode` still have `prefs` in this new formulation?

src/exchange_context.h
```diff
@@ -76,19 +87,31 @@ template <class T> struct ExchangeContext {
...
     bidders.insert(pb->bidder());
 
     bids_by_request[pb->request()].push_back(pb);
 
+    // MC comes from bid preference (or 0.0 if NaN)
     double bid_pref = pb->preference();
+    double mc = std::isnan(bid_pref) ? 0.0 : bid_pref;
+    
+    // MU comes from request preference
+    double mu = pb->request()->preference();
```
gonuke: Do we worry about `nan` on request preferences?

src/exchange_context.h
```diff
@@ -26,6 +26,17 @@ template <class T> struct PrefMap {
   typedef Bid<T>* bid_ptr;
 };
 
+/// @brief Maps for storing marginal cost (MC) and marginal utility (MU) adjustments
+/// These use the same structure as PrefMap but are semantically different:
+/// - MCMap stores marginal cost adjustments (from bids)
+/// - MUMap stores marginal utility adjustments (from requests)
+template <class T> struct MCMap {
+  typedef std::map<Request<T>*, std::map<Bid<T>*, double>> type;
+};
```
gonuke, line 35: Does it make sense to include the `typedef`s for `request_ptr` and `bid_ptr` here?  I note that they are defined in the other templated typedefs near here, but haven't studied where/how they are used.

src/greedy_preconditioner.cc
```diff
@@ -18,12 +18,28 @@ inline double SumPref(double total, std::pair<Arc, double> pref) {
...
+  if (graph == NULL) {
+    return 0.0;
+  }
+  
+  const std::map<ExchangeNode::Ptr, std::vector<Arc>>& node_arc_map = graph->node_arc_map();
+  
+  auto it = node_arc_map.find(n);
+  if (it == node_arc_map.end() || it->second.size() == 0) {
+    return 0.0;
+  }
```
gonuke: This makes me wonder if we should have a method 
`std::vector<Arc>& ExchangeGraph::GetArcsFromNode(ExchangeNode::Ptr n)`.  This may be the only place it's used right now, but probably a useful functionality.

src/exchange_solver.cc
```diff
@@ -2,15 +2,34 @@
...
 #include "context.h"
 #include "exchange_graph.h"
+#include "error.h"
 
 namespace cyclus {
 
-double ExchangeSolver::Cost(const Arc& a, bool exclusive_orders) {
-  return (exclusive_orders && a.exclusive()) ? a.excl_val() / a.pref()
-                                             : 1.0 / a.pref();
+double ExchangeSolver::Cost(const Arc& a, ExchangeGraph* graph, bool exclusive_orders) {
```
gonuke: This doesn't use the graph, and is therefore identical to `ArcWeight` (which doesn't use the `shift`!?

src/exchange_solver.h
```diff
@@ -55,8 +69,13 @@ class ExchangeSolver {
   double PseudoCostByPref(double cost_factor);
   /// @}
 
-  /// return the cost of an arc
-  inline double ArcCost(const Arc& a) { return Cost(a, exclusive_orders_); }
+  /// return the cost of an arc (instance method that uses the solver's graph)
+  inline double ArcCost(const Arc& a) {
+    if (graph_ == NULL) {
+      throw ValueError("ExchangeSolver::ArcCost requires graph to be set");
+    }
```
gonuke: Leaving a note here because `Cost` doesn't use `graph_` so maybe it doesn't matter anymore?

src/exchange_translator.h
```diff
@@ -69,36 +70,98 @@ template <class T> class ExchangeTranslator {
...
+      double arc_weight = it->mc() - it->mu() + ex_ctx_->shift_;
+      
+      // Reject arcs with negative weights - these trades should not be considered
+      if (arc_weight < 0.0) {
+        CLOG(LEV_DEBUG1) << "Removing arc with negative weight (" << arc_weight 
+                        << "). MC=" << it->mc() << ", MU=" << it->mu() 
+                        << ", shift=" << ex_ctx_->shift_ << ". This trade will be rejected.";
+        // Remove from node_arc_map first
+        auto& unode_arcs = node_arc_map[it->unode()];
+        unode_arcs.erase(std::remove(unode_arcs.begin(), unode_arcs.end(), *it), unode_arcs.end());
```
gonuke: I'm confused by this construct that combines `std::map::erase()` and `std::remove()`?  It seems as though `std::remove()` should do the work here and that there should be nothing for `std::map::erase()` to do...??

src/exchange_translator.h
```diff
@@ -69,36 +70,98 @@ template <class T> class ExchangeTranslator {
...
+        auto& vnode_arcs = node_arc_map[it->vnode()];
+        vnode_arcs.erase(std::remove(vnode_arcs.begin(), vnode_arcs.end(), *it), vnode_arcs.end());
+        // Remove from arcs vector
```
gonuke:
```suggestion
        // Remove from arcs vector and advances iterator
```

src/exchange_translator.h
```diff
@@ -69,36 +70,98 @@ template <class T> class ExchangeTranslator {
...
+      // Warn if arc weight is zero
+      if (arc_weight == 0.0) {
+        CLOG(LEV_WARN) << "Arc weight is zero. MC=" << it->mc() << ", MU=" << it->mu() 
+                       << ", shift=" << ex_ctx_->shift_;
+      }
+      
+      ++it;
+    }
+
+    // Also update arc weights in node_arc_map_ to keep it in sync
```
gonuke, line 85: You inherited this, but it seems odd to me that we are keeping a list of `Arc`s that are separate objects from the `Arc`s in the node-arc-map.  It may be its own refactor, but I wonder if `node_arc_map_` could have pointers to the arcs in the vector of arcs so that they are always consistent.  Maybe there is a reason for this I'm forgetting or didn't ever know?

src/exchange_translator.h
```diff
@@ -69,36 +70,98 @@ template <class T> class ExchangeTranslator {
...
+    auto& mc_map = ex_ctx_->trader_mc[req->requester()][req];
+    auto& mu_map = ex_ctx_->trader_mu[req->requester()][req];
+    
+    auto mc_it = mc_map.find(bid);
+    auto mu_it = mu_map.find(bid);
+    if (mc_it == mc_map.end() || mu_it == mu_map.end()) {
+      std::stringstream ss;
+      ss << "Bid not found in exchange context for arc addition.";
+      throw ValueError(ss.str());
+    }
```
gonuke, line 116: Is this robustness that you think should have existed in the original? or is this version more prone to something like this?

src/exchange_translator.h
```diff
@@ -69,36 +70,98 @@ template <class T> class ExchangeTranslator {
...
+    if (mc < 0) {
+      CLOG(LEV_DEBUG1) << "Skipping arc with negative MC=" << mc;
       return;
-    } else if (pref == 0) {
+    }
+    if (mu < 0) {
+      CLOG(LEV_DEBUG1) << "Skipping arc with negative MU=" << mu;
+      return;
+    }
+    if (mu <= 0) {
```
gonuke: You can only get here if `mu == 0` since you've already screened out `mu < 0` - is that intentional?

src/exchange_translator.h
```diff
@@ -219,18 +282,17 @@ ExchangeNodeGroup::Ptr TranslateBidPortfolio(
 /// updates the unit capacities for the associated nodes on the arc
 template <class T>
 Arc TranslateArc(const ExchangeTranslationContext<T>& translation_ctx,
-                 Bid<T>* bid) {
```
gonuke, line 222: Is there no longer a notion of defaults where you don't have to provide `mc` or `mu`?

gonuke (changes requested):
I did a first pass.  I don't see any major concerns, but the presence of this notion of `shift` seems to be unnecessary for the moment and clutters things a little.

I have only reviewed the degree to which the new code appears to perform analogous things to the old code without (re)learning the whole code and being comfortable that the whole code (still) works.  The changes seem well-contained such that this may not be necessary, other than the value of a periodic refamiliarization (for me) of all of this.

dean-krueger:
Thanks for the review @gonuke. I'm still fighting a few failing tests (should be done with that soon), after which I'll go through and take a look at your comments. As for the shift, there's this (perhaps purely historical) idea that negative total arc weights should be discarded entirely by the DRE (Enrichment does this in its AdjustMatlPrefs function to get rid of trades without any U235 in them), so the shift is there to make sure that the minimum "real" bid is 0 (and since the optimum of a LP is invariant to affine transformations in the objective we're safe to do this and then either subtract it out later or not since it's also not necessarily meaningful anyway). Happy to discuss further! As much as getting rid of "legacy" mode helped simplify all this a ton, there are still a few bumps and uneven joints between the new code and the old code that could probably be cleaned up/hammered out with a little more time/effort/thought.

src/exchange_context.h
dean-krueger: Short answer: likely not right now, but maybe later?

Long answer: I think those are only used in the `cpp_cyclus.pxd` file. My understanding is that this file is for the python implementation of Cyclus, which I didn't touch as a part of any of this. I think it should still work like it used to (there was a bit of stuff that was done to keep some backwards compatibility, for better or worse), but almost certainly will not have any of this MC MU business that was added as a part of this. Whether that rabbit hole needs to be dove into (and more importantly, whether it needs to be done right now while some of the dust is still settling on final implementation of all this) is maybe its own conversation, though I would argue this would be best as an `issue-->come-back-later` if anything at all.

src/exchange_context.h
dean-krueger: My original thinking was no, since I THINK by default it's 1.0 for everyone, but it's so cheap to add a check that it's probably worth doing.

src/exchange_graph.h
dean-krueger: I don't think this gets used, no. I'm honestly not sure it ever got used? I looked in the main branch before these edits and couldn't find anything, but that's a different story. It's one of those things that may be there just because "one day it may be helpful!", so I could see leaving it or removing it since it's vestigial at best. Happy to do either.

src/exchange_graph.h
dean-krueger: Just to be safe with this one, given the nature of your question, I had cursor take a look through the code base (it can be much more thorough much more quickly than I can be) and it concluded that this operation is currently unused in Cyclus, meaning this change should be okay. 

Side Note: It may be an interesting exercise at some point to have cursor or some similar tool go through and map out all the unused content we have floating around (it seems like there's at least a little bit of it) and see if we can't decide what's worth keeping vs what should be gotten rid of? Certainly not something to do now, though.

src/exchange_graph.h
dean-krueger: I updated the comments to be more accurate. This structure is likely something I'll want to circle back to on another pass after a little more design consideration.

src/exchange_translator.h
dean-krueger: [This page](https://www.geeksforgeeks.org/cpp/std-remove-algorithm-in-cpp/) explains it well:

> As you may have noticed, we still had to use the vector::erase() function to actually delete the elements. This is due to the fact that std::remove() function moves all the elements to remove at the end of the container and returns an iterator to the new end of the range that only contains the elements not to remove. It doesn't actually delete these elements from the container.

you need both, though.

src/exchange_translator.h
dean-krueger: I did this manually because it's easier that way, but this was added.

src/exchange_translator.h
dean-krueger: I'm not totally sure I understand the question, but I'm going to take a stab based on my best guess:

Previously what we did was we said "Let's find a preference to assign to this arc. Did the supplier give us one? if yes --> arc.pref = supplier.pref, if no --> arc.pref = requester.pref", what we do now is say "we need a MC from the supplier and an MU from the requester. Does the supplier give us one? if yes --> arc.mc = supplier.MC, if no--> supplier.MC = nan --> set arc.MC == 0. Does the requester give us one? (same structure)". you may be asking, and I believe already asked earlier about the requester.MU being nan but there's a default, so: "why did we only check for nan with the supplier.MC?", and the answer to that is "because the DRE functions previously had built in constructors for missing a single pref (which was, if I remember correctly, always attributed to the requester). Now, we need two pieces of information, though, and so if the requester always has a default built in, and archetype developers were taking advantage of that, AND they were unwittingly using the new Cyclus which expects two inputs (MC and MU) the one they'd be missing would almost certainly be MC, which would be nan, and MU would be set to whatever the default was". 

The above is all off the top of my head, so there may be some minor inaccuracies, but I believe that's what's going on. Hopefully that answers your question, but if not, feel free to reword and ask again!

src/exchange_solver.cc
dean-krueger: okay, so a few things are going into this:

1. when we do the SolveGraph step, which is what calls this Cost function, we've already translated the graph. During translation, we take the graph which has a bunch of arcs with MC and MU but dummy/default prefs (arc weights), and does the step where we calculate the actual pref (`arc_weight = it->mc - it->mu - ex_ctx->shift_` and then `a->pref(arc_weight)` or something like that), and so when we do `double arc_weight = a.pref()` we already have the shift accounted for and stored in the `pref` that `pref()` returns. **This means we can probably delete the `shift` argument**
2. `ArcWeight` doesn't appear to be called anywhere and could probably be deleted (NOTE: with all things like this I like to be sure, so I asked Cursor to double check this for me and it couldn't find any either).
3. The inclusion of `ExchangeGraph* graph` in the `ExchangeSolver::Cost()` function is PROBABLY from an earlier iteration of Cursor trying to make the shift work, and it just forgot/neglected to clean it up after the fact. Since it's unused, I can work to remove it from the call. The `graph` part was not there in the current live version of Cyclus, so this should be fine.


To address these, I will:
1. Delete the `ArcWeight()` function and all its infrastructure. This will also get rid of the `shift` issue.
2. Remove the `ExchangeGraph* graph` argument from `ExchangeSolver::Cost()`

src/exchange_translator.h
dean-krueger: Yeah, I was also thinking this was odd when I was mapping out how all this worked. I don't truly know what the point of the node_arc_map is, but today I learned about Chesterton’s Fence and think that your suggestion that this be its own refactor is a good one. I will make an issue now to reexamine this in the future.

dean-krueger: Done.

src/exchange_translator.h
dean-krueger: This appears to be Cursor wanting to throw a special error message. You could probably use `.at()` to find the keys in those maps instead (like the old code did), but if you do then you're stuck with the error message that `.at()` gives you if it doesn't find the key. If you use `.find()` and it doesn't find what it's looking for it just returns an iterator pointing to map.end(), so you can check it after and throw your own error. This could be changed back if desired.

src/exchange_translator.h
dean-krueger: This Cursorism is a bit weird, and I think was done so that there could be different error messages between thrown-out arcs and non-positive MU (which I vaguely remember being problematic, but am not totally sure at the moment about why)? In the universe where we remove the shift and allow negative arcs all of this goes away, though, so I'm tempted not to worry about it too much.

src/trade_executor.h
dean-krueger: I think this is related to the response I made earlier about checking MC nan and backwards compatibility. Short answer, if I remember correctly this is a backwards compatibility thing, since it was previously standard to only provide a "pref" once via the requester, but now we expect two values (MC and MU) and they need to come from the bidder and the requester. Older code will only provide from the requester (mapped to MU) so it's not unlikely that MC will be unset --> nan --> need a guard.

src/greedy_solver.cc
dean-krueger: This ended up not being used at all, and we're probably removing shift soon anyway, so I'm just going to take this out and if that causes a problem down the road it can be added back...

src/greedy_preconditioner.cc
dean-krueger: Should be easy enough.

dean-krueger: Time spent writing code: 5 minutes
Time spent figuring out why all my tests were failing: 1h 55m

But alas, we got there in the end.

src/prog_translator.cc
dean-krueger: oh we totally can. Good call. I'll do that.

dean-krueger: Done. I will say, in the universe where  the shift goes away and we allow negative arc weight this function probably goes away too, so I probably shouldn't have spent as much time on it as I did, but it was a good little exercise. If you're reading though again and see issues with this function please feel free to point them out, but I'll probably get rid of it entirely soon, so I may not implement suggestions. We'll see.

dean-krueger:
Alright, that took a while (oops), but finally finished going through your comments @gonuke. Thanks a ton for that first round of review. Before you do another, I think it may make sense for me to take a swing at refactoring this code a little bit to do the following:

1. Remove the shift entirely. It's nonsense and actively works against what I think we ultimately want.
2. Rename some things. Specifically MC and MU are probably bad names for what they are, and should be changed.
3. Look through all the files I touched for weird hanging functions like SumPref

I have three big project deliverables for classes next week, so I PROBABLY won't be able to get to it before spring break, but I'll try to hit that hard during that week where I am less distracted by class. I will, however, try to think about things a little bit and maybe put together some slides for next week's FCS meeting to get feedback on design choices I'm thinking about.

dean-krueger:
Status Update: I have one more PR coming from dean-krueger/cyclus into this PR and then I think it's ready to not be a draft anymore. When that happens, I'll probably give the title and description a bit of an update to better reflect what-all's actually in here.

dean-krueger:
Now that everything that should be in here is in here, I'll do two things:

1. Update the PR title and description to be more accurate to what this ended up being
2. Provide a diff between this PR branch and the `rename_dre_vars` branch so that it's easy for @gonuke (and me, since it's been a while) to know what's been looked at as a part of `rename_dre_vars` and what actually needs more looking.

Hopefully (and I think probably), that diff shows that there aren't many other things outside of rename_dre_vars, which ended up ballooning a bit outside of a simple variable name change and this is EZPZ (or at least as ezpz as possible). I will hold on re-requesting review until those two items are complete.

dean-krueger:
Autonuke seems happy with this, let's see if the real @gonuke agrees!

src/toolkit/matl_sell_policy.cc
```diff
@@ -94,6 +96,7 @@ MatlSellPolicy& MatlSellPolicy::Init(Agent* manager, ResBuf<Material>* buf,
   Trader::manager_ = manager;
   buf_ = buf;
   name_ = name;
+  unit_cost_ = 0.0;
```
gonuke: Isn't this already initialized to 0 in the constructor?  I think we are only setting values that are changing here from the parameter list.

src/toolkit/matl_sell_policy.cc
```diff
@@ -103,6 +106,7 @@ MatlSellPolicy& MatlSellPolicy::Init(Agent* manager, ResBuf<Material>* buf,
   buf_ = buf;
   name_ = name;
   set_throughput(throughput);
+  unit_cost_ = 0.0;
```
gonuke: ibid

src/toolkit/matl_sell_policy.cc
```diff
@@ -112,6 +116,7 @@ MatlSellPolicy& MatlSellPolicy::Init(Agent* manager, ResBuf<Material>* buf,
   buf_ = buf;
   name_ = name;
   set_ignore_comp(ignore_comp);
+  unit_cost_ = 0.0;
```
gonuke: ibid

src/toolkit/matl_sell_policy.cc
```diff
@@ -123,6 +128,7 @@ MatlSellPolicy& MatlSellPolicy::Init(Agent* manager, ResBuf<Material>* buf,
   name_ = name;
   set_throughput(throughput);
   set_ignore_comp(ignore_comp);
+  unit_cost_ = 0.0;
```
gonuke: ibid

src/toolkit/matl_sell_policy.cc
```diff
@@ -139,6 +145,7 @@ MatlSellPolicy& MatlSellPolicy::Init(Agent* manager, ResBuf<Material>* buf,
   set_ignore_comp(ignore_comp);
   set_package(package_name);
   set_transport_unit(transport_unit_name);
+  unit_cost_ = 0.0;
```
gonuke: ibid

src/toolkit/matl_sell_policy.cc
gonuke: Related - do we need/want a version of `Init` that allows you to set the `unit_cost`?  I know you already have a method to set this, but it looks like we generally set things through initialization.

src/bid.h
```diff
@@ -3,13 +3,15 @@
...
 #include <boost/weak_ptr.hpp>
-#include <limits>
 
 #include "request.h"
 #include "package.h"
 
 namespace cyclus {
 
+/// Default unit_cost values are one
+static const double kDefaultUnitCost = 1.0;
```
gonuke, line 13: Is this the most convenient place to define this? i.e. do we only use it in places that otherwise already include `bid.h`?

src/bid.h
```diff
@@ -24,54 +26,51 @@ template <class T> class Bid {
...
-                               boost::shared_ptr<T>
-                                   offer,
+                               boost::shared_ptr<T> offer,
                                Trader* bidder,
                                typename BidPortfolio<T>::Ptr portfolio,
                                bool exclusive,
-                               double preference) {
-    return new Bid<T>(request, offer, bidder, portfolio, exclusive, preference);
+                               double unit_cost,
+                               Package::Ptr package = Package::unpackaged()) {
```
gonuke: I'm not sure why we added packaging to this function signature, but then I'm a little unclear about the different variations that exist here across all these different factories.  Unless you are sure we need it, maybe we take it out and in a separate issue/PR we reorder these methods with better docstrings so that it's clear which ones are referencing which other ones and what the differences are.

src/material.cc
```diff
@@ -279,7 +279,7 @@ Material::Material(Context* ctx, double quantity, Composition::Ptr c,
       ctx_(ctx),
       prev_decay_time_(0),
       package_name_(package_name) {
-  SetUnitValue(unit_value);
+  this->unit_value(unit_value);
```
gonuke, line 282: In general, it should not be necessary to reference `this` since this is a member function
```suggestion
  unit_value(unit_value);
```

src/product.cc
```diff
@@ -135,7 +135,7 @@ Product::Product(Context* ctx, double quantity, std::string quality,
       tracker_(ctx, this),
       ctx_(ctx),
       package_name_(package_name) {
-  SetUnitValue(unit_value);
+        this->unit_value(unit_value);
```
gonuke, line 138: ibid

```suggestion
        unit_value(unit_value);
```

CHANGELOG.rst
gonuke, line 1: Is it necessary to add multiple entries to the changelog here?  There is no prohibition, but can we just assert that all of these changes are part of a single feature addition?

src/prog_translator.h
```diff
@@ -13,6 +13,7 @@ namespace cyclus {
 
 class ExchangeGraph;
 class ExchangeNodeGroup;
+class Arc;
```
gonuke: Do we need to forward declare this?  I don't see any change in this file that requires it.  We use it in the C++ but does it get included there in a header?

src/exchange_graph.cc
```diff
@@ -128,4 +130,19 @@ void ExchangeGraph::AddMatch(const Arc& a, double qty) {
   matches_.push_back(std::make_pair(a, qty));
 }
 
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+const std::vector<Arc>& ExchangeGraph::GetArcsFromNode(
+    ExchangeNode::Ptr node) const {
+  const std::map<ExchangeNode::Ptr, std::vector<Arc>>& arcs_by_node =
+      this->node_arc_map();
```
gonuke: This is a member variable- you can just access it directly & don't need to call this getter

src/exchange_translator.h
gonuke: We might still want a version of `TranslateArc` that takes no value for `unit_cost` and/or `unit_cost_mod` and calls the more complete one with default values?

src/exchange_solver.cc
```diff
@@ -2,32 +2,39 @@
...
 
 double ExchangeSolver::Cost(const Arc& a, bool exclusive_orders) {
-  return (exclusive_orders && a.exclusive()) ? a.excl_val() / a.pref()
-                                             : 1.0 / a.pref();
+  // Use stored arc cost which is set during translation.
+  double arc_cost = a.arc_cost();
+
+  if (exclusive_orders && a.exclusive()) {
+    // For exclusive arcs, scale by excl_val if needed
+    return arc_cost * (a.excl_val() > 0 ? 1.0 / a.excl_val() : 1.0);
```
gonuke: Why do we divide by this now?

src/exchange_solver.cc
```diff
@@ -80,18 +86,28 @@ double ExchangeSolver::PseudoCostByCap(double cost_factor) {
...
-  for (int i = 0; i != arcs.size(); i++) {
-    const Arc& a = arcs[i];
-    // remove exclusive value factor from costs for preferences that are less
-    // than unity. otherwise they can artificially raise the maximum cost.
-    double factor =
-        (a.exclusive() && a.excl_val() < 1) ? 1 / a.excl_val() : 1.0;
-    max_cost = std::max(max_cost, ArcCost(a) * factor);
+double ExchangeSolver::PseudoCostByArcCost(double cost_factor) {
+  double max_cost = -std::numeric_limits<double>::infinity();
+  bool found_arc = false;
```
gonuke: I'm not sure I understand the need for the `found_arc` logic here? it seems redundant once you are initializing the  max to negative infinity, and throwing exceptions on non-finite arc costs. It seems only relevant is there are no arcs, which you can presumably test directly.

src/exchange_solver.cc
```diff
@@ -80,18 +86,28 @@ double ExchangeSolver::PseudoCostByCap(double cost_factor) {
...
+      found_arc = true;
+    }
   }
-  return max_cost * (1 + cost_factor);
+
+  if (!found_arc) {
+    return 0.0;
+  }
+
+  double margin = cost_factor * std::max(1.0, std::abs(max_cost));
```
gonuke: Why abs value of `max_cost`? And why not `(1+cost_factor)`?

src/exchange_solver.h
```diff
@@ -42,21 +52,52 @@ class ExchangeSolver {
...
   }
 
-  /// @brief Calculates the ratio of the maximum objective coefficient to
-  /// minimum unit capacity plus an added cost. This is guaranteed to be larger
-  /// than any other arc cost measure and can be used as a cost for unmet
-  /// demand.
-  /// @param cost_factor the additional cost for false arc costs, i.e., max_cost
-  /// * (1 + cost_factor)
-  /// @{
+  /// @brief Calculates the default faux-arc penalty for unmet demand.
```
gonuke, line 55: Let's plan to discuss these `Pseudo` cost functions to  see sure we're all on the same page with their role in the DRE

tests/exchange_solver_tests.cc
```diff
@@ -1,19 +1,127 @@
...
+using cyclus::ExchangeNode;
+using cyclus::ValueError;
 
 // - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-class MockSolver: public ExchangeSolver {
+class MockSolver : public ExchangeSolver {
  public:
-  explicit MockSolver() : i(0) {}
+  explicit MockSolver(bool exclusive_orders = kDefaultExclusive)
+      : ExchangeSolver(exclusive_orders), i(0) {}
```
gonuke: Is this change necessary? The parent class' default constructor already sets the default value, and you never instantiate a `MockSolver` with a non-default value.

tests/exchange_test_cases.cc
```diff
@@ -242,8 +242,8 @@ void Case3a::Construct(ExchangeGraph* g, bool exclusive_orders) {
   q = 5;
   c1 = 5;
   c2 = 10;
-  p1 = 1;
-  p2 = 0;
+  p1 = 0;
+  p2 = 1;
```
gonuke, line 246: These have switched in relative magnitude.  Are we sure they are testing the same thing? Are they consistent with the comments in the header that define each case?  I think we need to check them all to be sure.

tests/exchange_translator_tests.cc
```diff
@@ -307,7 +423,9 @@ TEST(ExXlateTests, XlateArc) {
   ExchangeNodeGroup::Ptr bset =
       TranslateBidPortfolio(xlator.translation_ctx(), bport);
 
-  Arc a = TranslateArc(xlator.translation_ctx(), bid);
+  double unit_cost = bid->unit_cost();
+  double unit_cost_mod = req->unit_cost_mod();
+  Arc a = TranslateArc(xlator.translation_ctx(), bid, unit_cost, unit_cost_mod);
```
gonuke, line 428: This relates to my comment in the change to the default options for this method - let's discuss

tests/exchange_translator_tests.cc
```diff
@@ -325,96 +443,113 @@ TEST(ExXlateTests, XlateArc) {
...
+  Bid<Material>* bid8 = bport->AddBid(req2, get_mat(u235, qty + 1), supplier,
                                           exclusive);
 
   ExchangeContext<Material> ctx;
   ExchangeTranslator<Material> xlator(&ctx);
   TranslateRequestPortfolio(xlator.translation_ctx(), rport);
   TranslateBidPortfolio(xlator.translation_ctx(), bport);
 
+  // Helper to get unit_cost and unit_cost_mod for TranslateArc
+  auto get_cost_value = [](Bid<Material>* b) -> std::pair<double, double> {
```
gonuke: This also seems related to default behavior of `TranslateArc`

tests/exchange_translator_tests.cc
```diff
@@ -325,96 +443,113 @@ TEST(ExXlateTests, XlateArc) {
...
+    double unit_cost = b->unit_cost();
+    double unit_cost_mod = b->request()->unit_cost_mod();
+    return std::make_pair(unit_cost, unit_cost_mod);
+  };
+
   // bid > request && req exclusive && bid !exclusive,
   // so excl_val set to request qty
-  Arc a1 = TranslateArc(xlator.translation_ctx(), bid1);
+  auto cost_value1 = get_cost_value(bid1);
+  Arc a1 = TranslateArc(xlator.translation_ctx(), bid1, cost_value1.first, cost_value1.second);
```
gonuke: Can't `TranslateArc` extract this info from the `bid`?

tests/exchange_translator_tests.cc
```diff
@@ -430,7 +565,13 @@ TEST(ExXlateTests, SimpleXlate) {
   EXPECT_EQ(1, graph->arcs().size());
   EXPECT_EQ(0, graph->matches().size());
   const Arc& a = *graph->arcs().begin();
-  EXPECT_EQ(pref, a.unode()->prefs[a]);
+  // After Translate(), arc.arc_cost() contains unit_cost + unit_cost_mod
+  EXPECT_EQ(unit_cost_mod, a.unit_cost_mod());
+  // Bid has no explicit unit_cost, defaults to 1
+  EXPECT_EQ(1.0, a.unit_cost());
```
gonuke: Replace with the variable for default

tests/toolkit/matl_sell_policy_tests.cc
```diff
@@ -39,10 +41,10 @@ class MatlSellPolicyTests: public ::testing::Test {
     CompMap v;
     v[pyne::nucname::id("H1")] = 1;
     comp = Composition::CreateFromAtom(v);
-    mat = Material::CreateUntracked(qty, comp);
+    mat = Material::CreateUntracked(qty, comp, 0.0);
```
gonuke: Did the interface change so that we now need to provide this value? Or the semantics?

tests/toolkit/matl_sell_policy_tests.cc
```diff
@@ -172,7 +236,7 @@ TEST_F(MatlSellPolicyTests, Package) {
   cm[922350000] = 0.05;
   cm[922380000] = 0.95;
   Composition::Ptr comp = Composition::CreateFromMass(cm);
-  mat = Material::Create(a, qty, comp, Package::unpackaged_name());
+  mat = Material::Create(a, qty, comp, Package::unpackaged_name(), 0.0);
```
gonuke: Is this change necessary?

tests/toolkit/matl_sell_policy_tests.cc
```diff
@@ -236,7 +300,7 @@ TEST_F(MatlSellPolicyTests, TransportUnit) {
...
```
gonuke: Necessary change?

tests/toolkit/matl_sell_policy_tests.cc
```diff
@@ -330,7 +394,7 @@ TEST_F(MatlSellPolicyTests, PackageFailedTrade) {
   CompMap cm;
   cm[922380000] = 1;
   Composition::Ptr comp = Composition::CreateFromMass(cm);
-  mat = Material::Create(a, qty, comp, Package::unpackaged_name());
+  mat = Material::Create(a, qty, comp, Package::unpackaged_name(), 0.0);
```
gonuke: Necessary change?

tests/resource_exchange_tests.cc
```diff
@@ -278,19 +275,21 @@ TEST_F(ResourceExchangeTests, PrefValues) {
...
 
   Bidder* bidr = new Bidder(tc.get(), commod);
 
   BidPortfolio<Material>::Ptr bp(new BidPortfolio<Material>());
-  Bid<Material>* pbid = bp->AddBid(preq, mat, bidr);
-  Bid<Material>* cbid = bp->AddBid(creq, mat, bidr);
+
+  // Bids without a unit_cost default to unit_cost = 1
+  Bid<Material>* pbid = bp->AddBid(preq, mat, bidr, false);
+  Bid<Material>* cbid = bp->AddBid(creq, mat, bidr, false);
```
gonuke: I don't think we need to add the explicit `false` here since that is the default value in one of the interfaces

tests/resource_exchange_tests.cc
```diff
@@ -303,21 +302,24 @@ TEST_F(ResourceExchangeTests, PrefValues) {
...
+  EXPECT_EQ(context.trader_arc_costs[child], cexp);
 
   EXPECT_NO_THROW(exchng->AdjustAll());
 
-  pobs[preq].begin()->second = std::pow(preq->preference(), 2);
-  cobs[creq].begin()->second = std::pow(std::pow(creq->preference(), 2), 2);
-  EXPECT_EQ(context.trader_prefs[parent], pobs);
-  EXPECT_EQ(context.trader_prefs[child], cobs);
+  pexp[preq].begin()->second = p_arc_cost * p_arc_cost;
+  cexp[creq].begin()->second = std::pow(c_arc_cost, 4);
```
gonuke, line 320: Suggestion for new issue:
The fact that the adjusted values are powers of the original values is not intuitive, but determined by the fact that this mock facility chooses that operation for adjustment.

It might be better to define some adjustment amount and multiply by that, so that the test for success can refer to those adjustment factors and it will be more transparent.

(Not this PR, since this was inherited from previous code)

tests/greedy_preconditioner_tests.cc
```diff
@@ -1,27 +1,115 @@
...
+  // Test single arc: u2 should have average of 1.5
+  EXPECT_DOUBLE_EQ(AvgCost(u2, &g), 1.5);
+
+  // Test node with no arcs returns 0.0
+  EXPECT_DOUBLE_EQ(AvgCost(u3, &g), 0.0);
+  EXPECT_EQ(0u, g.node_arc_map().count(u3));
+
+  // Test ordering: u1 (2.0) > u2 (1.5) > u3 (0.0)
+  EXPECT_TRUE(AvgCost(u1, &g) > AvgCost(u2, &g));
+  EXPECT_TRUE(AvgCost(u2, &g) > AvgCost(u3, &g));
```
gonuke: This seems redundant to above tests

tests/greedy_preconditioner_tests.cc
```diff
@@ -34,8 +122,8 @@ TEST(ConditionerTests, Conditioning) {
   n12->commod = "spam";
   ExchangeNode::Ptr n13(new ExchangeNode());
   n13->commod = "eggs";
-  double n1epref = 1/4;
-  double n1spref = 3/4;
+  double n1ecost = 1.0/4.0;
+  double n1scost = 3.0/4.0;
```
gonuke, line 122: Is this change necessary? C++ always does floating point math

tests/greedy_preconditioner_tests.cc
```diff
@@ -68,18 +156,19 @@ TEST(ConditionerTests, Conditioning) {
   Arc n21e(n21, eggs);
   Arc n22s(n22, spam);
 
+  // Set arc costs - this is what AvgCost reads
+  n11e.arc_cost(n1ecost);
+  n12s.arc_cost(n1scost);
+  n13s.arc_cost(n1scost);
+  n21e.arc_cost(n2ecost);
+  n22s.arc_cost(n2scost);
```
gonuke, line 160: In previous version this was passed `n2ecost`- is the change intentional/meaningful? (Maybe it was wrong before?)

tests/greedy_preconditioner_tests.cc
```diff
@@ -94,44 +183,44 @@ TEST(ConditionerTests, Conditioning) {
...
 
   // final state
-  EXPECT_EQ(g.request_groups().at(0), g2);
-  EXPECT_EQ(g.request_groups().at(0)->nodes().at(0), n22);
-  EXPECT_EQ(g.request_groups().at(0)->nodes().at(1), n21);
-  EXPECT_EQ(g.request_groups().at(1), g1);
-  EXPECT_EQ(g.request_groups().at(1)->nodes().at(0), n12);
-  EXPECT_EQ(g.request_groups().at(1)->nodes().at(1), n11);
-  EXPECT_EQ(g.request_groups().at(1)->nodes().at(2), n13);
+  EXPECT_EQ(g.request_groups().at(0), g1);
```
gonuke, line 213: It looks like the ordering changed. Since order is something we are testing... is this intentional?

tests/greedy_solver_tests.cc
```diff
@@ -15,26 +15,34 @@ using cyclus::GreedySolver;
...
-  nodes.push_back(u1);
   nodes.push_back(u2);
+  nodes.push_back(u1);
 
-  EXPECT_EQ(nodes[0], u1);
-  EXPECT_EQ(nodes[1], u2);
-  std::sort(nodes.begin(), nodes.end(), AvgPrefComp);
   EXPECT_EQ(nodes[0], u2);
   EXPECT_EQ(nodes[1], u1);
+  std::sort(nodes.begin(), nodes.end(), AvgArcCostComp(&g));
```
gonuke, line 43: Note to @gonuke to understand why we need an `ExchangeGraph` now to perform this operation?

tests/prog_translator_tests.cc
```diff
@@ -56,15 +67,23 @@ TEST(ProgTranslatorTests, translation) {
...
+  // For exclusive arcs with excl_val > 0: obj_coeff = arc_cost * (1.0 / excl_val)
   std::vector<double> obj_coeffs;
+  double excl_val = 2.0;  // excl_val for exclusive arcs (set by excl_flow[1])
   for (int i = 0; i != narcs; i++) {
-    obj_coeffs.push_back((excl_flow[i] != 0) ?
-                         excl_flow[i] / prefs[i] : 1 / prefs[i]);
+    double coeff = arc_costs[i];
+    if (excl_flow[i] != 0) {
+      // Exclusive arcs: ExchangeSolver::Cost returns arc_cost * (1.0 / excl_val)
+      coeff = arc_costs[i] * (1.0 / excl_val);
```
gonuke: There is a previous comment to confirm this division??

gonuke: And why doesn't this use the value of `excl_flow[i]`?

gonuke (changes requested):
This all looks pretty good.  Most comments are cosmetic - maybe things introduced by AI? 

There are a few conceptual things for us to discuss in a meeting, perhaps with some slides to help frame the conversation.

src/toolkit/matl_sell_policy.cc
dean-krueger: this was the route I went initially, but it was a nightmare trying to get all the init functions overloaded with a double that had a default value. Two or three of them had conflicting signatures or were ambiguous if you had a default.

src/bid.h
dean-krueger: That's a good question! I can take another look. My gut feeling is that since the unit cost is logically tied to the bid it's at least "fine" to do this here, but it's possible there's a "better" spot?

src/bid.h
dean-krueger: This may have been a thing that was added to disambiguate function overloads or something, I'll take a closer look, though.

src/material.cc
dean-krueger: I think here it is for some reason. The reason it's there in the first place is because it conflicts with something (I think the Resource unit_value) and so if I remember correctly you get an error if there isn't `this->unit_value()` and I think it's because it doesn't know which unit value it is. I can remove it and, assuming I get an error like I think I will, paste the error just for peace of mind.

CHANGELOG.rst
dean-krueger: No, it just seemed like a lot of stuff. Happy to make it all one line.

src/toolkit/matl_sell_policy.cc
dean-krueger: Good catch. No need to do it again!

src/bid.h
dean-krueger: Okay, I looked at the blame and saw that I did this three months ago (time flies, wow). That was when I was initially adding things to the end of these that needed `= ...` (I forget what that's called, but in python it's keyword vs positional arguments I think and maybe that's what it is in C++ too?). However, to have those, I remember needing to have the packaging stuff otherwise it would yell at me. I can remove these for now.

src/material.cc
dean-krueger: Yeah, as I remembered, this is the error you get when you don't do `this->unit_value`:

image

 It IS a member function, but it's a member function of resource, so I guess the compiler gets confused if you just do `unit_value(unit_value)`.

src/product.cc
dean-krueger: see above

src/prog_translator.h
dean-krueger: Nope! You're right. This looks like it was a part of an earlier version of this PR. Autonuke probably just didn't catch it as a part of its cleanup pass. I'll remove it.

src/exchange_graph.cc
dean-krueger: Would you look at that! You're right.

src/exchange_solver.cc
dean-krueger: Hard to say. I suspect the thought was "flip from 1/pref to arc cost so we should flip the other one to keep it the same!" but that's not right... We need to multiply by it. We have

$$
arc \space cost \times flow
$$

where flow is

$$
flow = y \times excl \space val
$$

and

$$
y \in \set{0,1}
$$

represents the exclusivity thing from the MILP (the binary that tells whether flow happens on that arc at all) and excl_val is the total flow across the arc if that trade happens at all. Since we're calculating cost here, it should be the unit_cost times the total flow, not the unit_cost divided by the total flow. That would be silly. Really good catch!

dean-krueger: flipped this around and fixed the corresponding tests.

src/exchange_solver.cc
dean-krueger: We need to ensure that the pseudo cost is less attractive than the worst possible real bid. Before, we forbid 0 and negative bids, so doing `max_cost * 1 + cost_factor` always gave a pseudo cost worse than the worst max_cost, which was always positive. However, since we now allow negative and zero bids we need to protect against all the costs being zero (`std::max(1.0, ...)` makes sure that our pseudo cost isn't also 0) and also against all the costs being negative (consider the ordering of the costs with real bids at -1, -2, and -3: without the abs there the pseudocost would come out to be -1.1 which is BETTER than -1!)

src/exchange_solver.h
dean-krueger: Sure! I'll try to remember to throw some slides together for the FCS meeting that week we get back from GLOBAL walking through how I see them and hopefully that'll be a good starting point for a conversation about this.

tests/exchange_solver_tests.cc
dean-krueger: Probably not. There may have been some forward-thinking intention of adding a test that used something like this, but for the sake of not overpopulating this PR I think it makes sense to remove it.

tests/exchange_test_cases.cc
dean-krueger: I believe they switched because previously higher pref was better and now lower arc_cost is better, but just to be sure I had Codex do a full audit of the file against upstream/v2.0-dev and check that the tests are spiritually the same, and also that the header documentation was consistent. It found:

1. The switched order changes "more preferred" into "less costly" correctly.
2. There is a documentation issue, which I will fix.

Good catch!

dean-krueger: I guess it also found a copy-paste error in another of the tests, so I'll fix that too. Just in case you're looking at this again and notice I changed case 2b: i did so to make the test match the header (the old version was a copy of case 2f)

tests/exchange_translator_tests.cc
dean-krueger: Happy to. I added that function you were talking about such that it used `kDefaultUnitCost` and `kDefaultUnitCostMod` but the other semi-obvious way to have done that is to have it default to `bid->unit_cost()` and `bid->req->unit_cost_mod()` (or however that works in reality). Feel free to either indicate preference for one vs the other here, or we can talk about it more in the FCS meeting.

dean-krueger: okay, reading your next comment I changed it from `kDefault...` to the `bid->unit_cost()` thing so I could get rid of that helper and then all those tests simplify. That seems like probably the right way to go. Still happy to talk about it. Maybe there's a use case for the other one, too, but idk.

tests/exchange_translator_tests.cc
dean-krueger: Yep! I axed this after changing the default `TranslateArc` function to use the values from the bid.

tests/exchange_translator_tests.cc
dean-krueger: it sure can!

tests/exchange_translator_tests.cc
dean-krueger: I'm surprised Autonuke didn't catch this magic number! Good catch. Fixed.

tests/toolkit/matl_sell_policy_tests.cc
dean-krueger: Materials (actually resources) have a sentinel "unset unit value" value of nan:

resource.h L118-120
```resource.h
protected:
constexpr static double kUnsetUnitValue =
    std::numeric_limits<double>::quiet_NaN();
```

we went with nan so that if you truly didn't set a value it wasn't accidentally mistaken for a free material with unit_value = 0.0. Later, however, some guards were introduced to the matl_sell_policy to prevent materials with unset unit_value from being traded:

matl_sell_policy.cc L239-244
```matl_sell_policy.cc
double unit_value = m->unit_value();
if (!std::isfinite(unit_value)) {
  throw ValueError(
    "MatlSellPolicy cannot bid material with non-finite unit_value"
  );
}
```

so I had to go and set these materials to have a unit_value of 0.0.

HOWEVER

now that I type all this out I'm starting to think that it's probably fine for materials to be free by default, since the new DRE philosophy has so far been "if you don't do anything different you get old Cyclus behavior"... I think I'll change `kUnsetUnitValue` to 0.0 and remove these bits.

tests/toolkit/matl_sell_policy_tests.cc
dean-krueger: Yes but I'm undoing it to do it different. See above.

tests/toolkit/matl_sell_policy_tests.cc
dean-krueger: See above

tests/toolkit/matl_sell_policy_tests.cc
dean-krueger: See above

tests/resource_exchange_tests.cc
dean-krueger: Yep, fair enough! Removed.

tests/resource_exchange_tests.cc
dean-krueger: Made

tests/greedy_preconditioner_tests.cc
dean-krueger: agreed. Removed.

tests/greedy_preconditioner_tests.cc
dean-krueger: I do not think this is true. 

```g.cc
double n1ecost = 1/4;
double n1scost = 3/4;

std::cout << "n1ecost: " << n1ecost std::endl;
```

gives:

image

which is what I'd expect from integer division.

tests/greedy_preconditioner_tests.cc
dean-krueger: I'm afraid I don't quite see what you're getting at (specifically I don't see this previous version you're talking about so it's hard to tell what you mean exactly). Happy to look again, but could you clarify?

dean-krueger: That being said, I suspect it was a mistake. There are a few confusing naming conventions being used here. I'll draw a picture and maybe it will make more sense to both of us.

dean-krueger: oh wait, I see where you're talking about. 

`n22->prefs[n22s] = n2epref` 

n22 is asking for spam and is connected to the spam supplier, so I THINK it makes sense to set the arc cost with node 2's spam cost (`n2scost`). That being said, earlier in the test we also connect n13 to the spam supplier even though it wants eggs. It is not totally clear to me whether that was a mistake or if it's testing something specific, though I suspect it is a mistake as well, and will look more closely as I draw this picture.

dean-krueger: I THINK n13 gets connected to spam to see what happens if you cross them? I can't think of any other reason to have it (if we connect it to eggs I think it's identical to n11).

tests/greedy_preconditioner_tests.cc
dean-krueger: Yes. `GreedyPreconditioner::Condition` used to sort descending because we did 1/pref (higher pref is better), but we flipped it so now we want to sort ascending (lower cost is better). This made the final ordering of those nodes opposite from what it used to be.

That being said: this test kind of sucks as it is right now. It's confusing and does a whole bunch of stuff all at once. I don't REALLY want to fix it, but I'm on the fence enough about it that if you ALSO think it's not very clear what's happening here that'd feel like enough for me to go and make it more clear overall.

tests/greedy_solver_tests.cc
dean-krueger: So this is sorting the nodes by their average arc cost. We pass in nodes to sort, and then use the AvgArcCostComp function to compare two nodes. It goes in and uses the Greedy Preconditioner function `AvgCost(node, graph)` to loop over all the arcs connected to the `node` on the `graph`. However, we need the graph to look up what the average arc cost of arcs connected to that node is since ExchangeNodes don't record this value or have any sense of which other nodes they're connected to (that's the graph's job).

dean-krueger: Before, we did this:

```gs.h
inline bool AvgPrefComp(ExchangeNode::Ptr l, ExchangeNode::Ptr r) {
  int lid = l->agent_id;
  int rid = r->agent_id;
  double lpref = AvgPref(l);
  double rpref = AvgPref(r);
  return (lpref != rpref) ? (lpref > rpref) : (lid > rid);
```

which used AvgPref:

```file.cc
double AvgPref(ExchangeNode::Ptr n) {
  std::map<Arc, double>& prefs = n->prefs;
  return prefs.size() > 0
             ? std::accumulate(prefs.begin(), prefs.end(), 0.0, SumPref) /
                   prefs.size()
             : 0;
```

we USED to store the prefs in each exchange node:

```en.h
struct ExchangeNode {
 public:
  typedef boost::shared_ptr<ExchangeNode> Ptr;

  ExchangeNode();
  explicit ExchangeNode(double qty);
  ExchangeNode(double qty, bool exclusive);
  ExchangeNode(double qty, bool exclusive, std::string commod);
  ExchangeNode(double qty, bool exclusive, std::string commod, int agent_id);

  /// @brief the parent ExchangeNodeGroup to which this ExchangeNode belongs
  ExchangeNodeGroup* group;

  /// @brief unit values associated with this ExchangeNode corresponding to
  /// capacties of its parent ExchangeNodeGroup. This information corresponds to
  /// the resource object from which this ExchangeNode was translated.
  std::map<Arc, std::vector<double>> unit_capacities;

  /// @brief preference values for arcs
  std::map<Arc, double> prefs;

  /// @brief whether this node represents an exclusive request or offer
  bool exclusive;

  ...
```

and I honestly don't remember why this got axed. Hopefully that makes sense, though!

tests/prog_translator_tests.cc
dean-krueger: Division corrected. The excl_flow was just all 2's, so using excl_val was the same, but definitely less clear/correct. Changed it to correctly use the excl_flow[i] instead.

dean-krueger:
Thanks a ton for the timely review @gonuke! It took me some time to get around to all of your comments, but I think I've finally addressed them all. The ones which were basically "yep, you're right good catch!" I "resolved" so they weren't cluttering things as much, and the ones which maybe warrant another look I left unresolved but tried to make sure I responded appropriately to each. This SHOULD be ready for another round of review.

src/toolkit/matl_sell_policy.cc
gonuke: Let's make an issue to review these.  I think they've growth with additions in ways that need to be more carefully considered, but outside this PR.

src/bid.h
gonuke: This may be another place that needs a new issue for a careful review of all the default arguments to make sure we have the most logical/consistent set of options.

src/exchange_solver.cc
gonuke: If you take `std::max(1.0, max_cost)` you'll take care of cases where `max_cost < 0`. Right now, if `max_cost` is -100, this will give `cost_factor * 100` which seems higher than necessary (maybe it doesn't matter??).

Separately, this could also be `std::max(0.1, max_cost)`, right?  It just needs to be anything larger than 0.

And also - Still not sure why we don't multiply by `(1 + cost_factor)`?

tests/greedy_preconditioner_tests.cc
gonuke: huh... I guess that makes sense.  It also means that it probably wasn't working before? Or is one of them 0 and the other 1 by integer division and thus still different in the logical way.  Technically, I think only one of the constants needs to be a double to ensure floating point division

src/exchange_solver.cc
gonuke:
```suggestion
  double margin = std::max(1.0, cost_factor * max_cost);
```

tests/exchange_test_cases.cc
```diff
@@ -98,8 +98,8 @@ void Case2b::Construct(ExchangeGraph* g, bool exclusive_orders) {
   qty = 10;
   unit_cap_req = 1;
   capacity = 10;
-  unit_cap_sup = 2;
-  flow = capacity / unit_cap_sup;
+  unit_cap_sup = 1;
+  flow = capacity;
```
gonuke, line 102: Is there a reason for this change?  Does it change the purpose of the test? Does it change the behavior of the test?

tests/resource_tests.cc
```diff
@@ -90,15 +90,15 @@ TEST_F(ResourceTest, MaterialExtractGraphid) {
...
 TEST_F(ResourceTest, MaterialUnitValue) {
-  EXPECT_EQ(m1->UnitValue(), 10);
-  EXPECT_EQ(m2->UnitValue(), 20);
+  EXPECT_EQ(m1->unit_value(), 10);
+  EXPECT_EQ(m2->unit_value(), 20);
   m1->Absorb(m2);
-  EXPECT_EQ(m1->UnitValue(), 17);
-  EXPECT_TRUE(std::isnan(m3->UnitValue()));
+  EXPECT_EQ(m1->unit_value(), 17);
+  EXPECT_EQ(0.0, m3->unit_value());
```
gonuke:
```suggestion
  EXPECT_EQ(kUnsetUnitValue, m3->unit_value());
```

tests/resource_tests.cc
```diff
@@ -129,15 +129,15 @@ TEST_F(ResourceTest, ProductExtractGraphid) {
...
 TEST_F(ResourceTest, ProductUnitValue) {
-  EXPECT_EQ(p1->UnitValue(), 10);
-  EXPECT_EQ(p2->UnitValue(), 20);
+  EXPECT_EQ(p1->unit_value(), 10);
+  EXPECT_EQ(p2->unit_value(), 20);
   p1->Absorb(p2);
-  EXPECT_EQ(p1->UnitValue(), 17);
-  EXPECT_TRUE(std::isnan(p3->UnitValue()));
+  EXPECT_EQ(p1->unit_value(), 17);
+  EXPECT_EQ(0.0, p3->unit_value());
```
gonuke:
```suggestion
  EXPECT_EQ(kUnsetUnitValue, p3->unit_value());
```

gonuke (changes requested):
A few final comments here (until we talk about the greedy preconditioner tests)

tests/exchange_test_cases.cc
dean-krueger: Yes! This was an autonuke catch. Previously, this test as written was a clone of case 2f, but the comment in the header file said it should be testing something different. Not sure how it got put in as a clone, but it didn't match the comment, so I changed it to match since that seemed like intended behavior. I have autonuke read through all the tests of all the files that get touched in a PR and look for correctness and completeness, so this was flagged.

tests/resource_tests.cc
dean-krueger: Note, changing this to match what you have to avoid the magic numbers, but doing so requires moving kUnsetUnitValue into resource's public zone (api? I've never understood how to use this term...)

gonuke: Or make the tests a friend of the class so it can access private info?

dean-krueger:
Less the test I'm making slides for, this should I think be good for another round!

PR #1940, opened 2026-03-03
Make Recipes Required by Schema Validation

### LINKED ISSUE #1939: [Bug] Simulations (maybe only sometimes?) require a recipe block even if unused
# Describe the bug.

In simulations which do not otherwise need recipes, and therefore have no recipe block (I checked the validation, `recipe` is `zeroOrMore`) cyclus throws:

```
terminate called after throwing an instance of 'cyclus::ValueError'
  what():  Invalid table name Compositions
```

Adding a recipe block (even if it's unused) causes this to stop.

# Steps to reproduce the bug.

1. Basic cyclus input file with source and storage
2. no recipe block
3. `cyclus input_file.xml -o test.sqlite` <-- (test.sqlite does not exist, that throws a different error if it does)
4. error thrown
5. add recipe block 
6. re-run
7. no error

## What inputs and outputs are involved?

[throws_error.xml](https://github.com/user-attachments/files/24650466/throws_error.xml)

[works.xml](https://github.com/user-attachments/files/24650467/works.xml)

 - Operating System: Linux
 - Cyclus Version: v2.0-dev (fairly sure this is unrelated)
 - Python Version: 3.13 (pretty sure)

Recipes should be optional

# Additional context.

Not sure! I haven't entirely figured this bug out, but I wanted to report it.

Fixing the bug.

dean-krueger:
# Summary of Changes
It was previously identified that the lack of a recipe block in simulations which use Cycamore agents caused the simulation to fail, due to the Compositions table requiring at least one recipe (regardless of whether it got used or not), even though the schema lists recipes as `zeroOrMore`. In a conversation I had with @gonuke, it was decided that it probably doesn't make sense to run Cyclus simulations without any recipes, and so to more easily help users identify that recipes are required I have changed the schema from `zeroOrMore` to `oneOrMore`.

## Design Notes

I changed the regular and the "flat" schemas, but I wasn't really sure what the difference was. I figured this was probably the right thing to do, though. 

- [x] API for Cyclus modules
- [ ] Input files (if the input files only use the agents from Cyclus (prey, predator, etc) this may break them. 

# Related CEPs and Issues

Fixes #1939 

# Associated Developers

None

# Testing and Validation

When I made this change the input files in IntegTests which didn't have recipes (because they only used Cyclus agents and not Cycamore agents and so did not build the Compositions table) failed as expected, and I added a test to make sure that if a file doesn't have a recipe it should fail schema validation. Otherwise all other unit tests passed.


gonuke (approved):
Thanks for tackling this @dean-krueger .

Seeing the updates to the predator/prey inputs indicates that there could be a model without recipes that does make sense, in general. 

I think that as soon as Cycamore agents are used, it's probably necessary.

I think that the most typical use case will require recipes, and - at least for now - a problem that only uses agents that don't require recipes will be exceedingly rare.

gonuke (changes requested):
Apparently there are python tests that need some attention , too.

dean-krueger:
Oh interesting! I've never had to mess with the python tests before. I'm guessing there are a few input files in there which don't include the recipe. It also occurred to me that the Recipe page on the website lists them as optional. I'll make an issue to fix that.

gonuke (approved):
Thanks @dean-krueger

dean-krueger:
Did this fail tests still, though? I saw that there were three that weren't passing earlier. It looked like they were failing because of GitHub issues, but I also see that Cymetric looks broken, too? @gonuke

gonuke:
We've been failing cymetric for a while because of changed in the transaction table

PR #1943, opened 2026-05-06
Agent Registered Discrete Event Timing with Default Behavior

meg-krieg:
This very preliminary implementation is intended to see how discrete event timing would be incorporated into Cyclus. 
The current changed files feel very (and suspiciously) light compared to the involved functionality they hope to achieve. I hope that this PR will undergo a lot of iterations to improve. Given the discussions in CEP31 so far, it was decided that discrete-event cyclus implementation should 

1.  give flexibility to cycamore archetype developers.
- Hence, this PR introduces a default trade-event scheduling behavior that replicates cyclus's fixed increment timestep. It is up to the archetype developer to create their own event request function/trade scheduling function ```EventRequest``` that overrides the default. This request/*consumption* based scheduling function should always use ```context()--> RegisterRequester(----)``` and ```context()-->DeregisterRequester(---)``` to add/remove a facility to/frome trading timeline map
- if cyclus 1.6v fixed increment is not to be the default, what should be?
2. any agent can add itself to the execution queue for any timestamp in the future
- ```EventRequest()``` to schedule trade events is added into the ```Tock```. Alternatively ,build is preconfigured and a subset of decommissions are scheduled at ```facility::build```(explained in files). 
3. whenever an agent acts, it must Tick/Trade/Tock this is presumably consumer-driven; there is no reason for a supplier to trigger action
- Refer to 1. 
- If we want to ensure archetype developers are creating a scheduling function based on consumption... should we split the trader class into two entities (consumer/producer?)
4. all suppliers must be available any time that a consumer is acting they must also perform a Tick/Trade/Tock
- no changes needed
5. all consumers of the same commodity should always act at the same time
- not yet implemented. Should this link between commodities and a consumer occur in Cycamore (for archetype developers to forward to context into a simulation wide map) or cyclus..?


meg-krieg:
@dean-krueger 
I was wondering (because you are working a lot with mapping  preferences/economics features in cyclus and the toolkit) whether you know/have recommendations of already pre-existing ability in the toolkit that maps commodities to all consumers of that commodity (or some variation of that)? I noticed some interesting stuff with ```CommodityProducer``` and ```CommodityProducerManager```. But, I have been mainly deep diving into time-simulation dynamics, and I haven't looked into DRE/material specifics as much as you have, so I thought I'd ask!

dean-krueger:
> I was wondering... whether you know/have recommendations of already pre-existing ability in the toolkit that maps commodities to all consumers of that commodity (or some variation of that)?

Check `exchange_context.h` L106 - L107. It has:

```
  /// @brief maps commodity name to requests for that commodity
  typename CommodMap<T>::type commod_requests;
```
  
this gets populated every time the DRE runs and AddRequest gets called, but is only available during the DRE (the exchange context gets deleted after the DRE runs I believe), but you could fairly easily copy it and then just every time the DRE runs check if there are any new facilities asking for that resource? This is all sort of the "5 minutes of digging" response, but if you're interested in a more thorough design consideration let me know and I could sketch something out if I knew more about what you were trying to do with it!

PR #1946, opened 2026-05-22
Decay small amounts

### LINKED ISSUE #1917: [Feature] Allow Cyclus to Decay Small Amounts of Materials

In Tricycle, the Fusion Energy System module for Cyclus, we are very often dealing with very small amounts of tritium which has a short half life, and our simulations take place over the span of days. Currently, Cyclus does not perform decay on these small amounts of tritium, and so it becomes difficult to make realistic simulations in Tricycle.

Allow the decay of very small amounts of materials in Cyclus, either by default, or with a decay-mode flag (such as `lazy` , `never`, or `manual`, ie `precise`).

Simply lowering the value of `eps` in the `material.cc` file technically works, but then we lose the functionality we wanted initially, which is that we'd get VERY minute changes in compositions of materials with really long half-lives which would be expensive/annoying/.

# Additional context.

It is possible that there is a bug somewhere in decay/ResBuf interactions, since what I'd expect is that the part where the material holds on to its previous decay time  doesn't SEEM to be working. This may require some sleuthing I probably don't have time for at this exact moment, so a bandaid solution may be good in the meantime.

dean-krueger:
# Summary of Changes

Changes `eps = 1e-3` in `Material::Decay()` to `eps_decay = 1e-4`, pegging the value to allowing tritium to decay on a one day time step, and adds a test to make sure this works.

## Design Notes

I started by writing the test, which failed until I lowered `eps_decay`. I also decided to rename `eps` --> `eps_decay` to bring it more in line with other `eps` in cyclus (notably `eps_rsrc`).

- [ x API for Cyclus modules
- [x] Input files

# Related CEPs and Issues

Closes #1917

# Associated Developers

None.

# Testing and Validation

Added a test to make sure tritium could decay in a simulation with a 1 day time step.

### DIFF
--- a/tests/material_tests.cc
+++ b/tests/material_tests.cc
@@ -338,7 +338,7 @@ TEST_F(MaterialTest, DecayShortcut) {
 
   double sec_per_month = 2629152;
   double u235_lambda = pyne::decay_const(u235) * sec_per_month;  // per month
-  double eps = 1e-3;
+  double eps = 1e-4;
   double threshold = -1 * std::log(1-eps) / u235_lambda;
 
   // If delta t is small w.r.t. composition's decay constants, no decay is
--- a/tests/material_tests.cc
+++ b/tests/material_tests.cc
@@ -440,6 +440,31 @@ TEST_F(MaterialTest, DecayHeatTest) {
   ASSERT_NEAR(3.614E-14 , dec_heat, 0.0005);
 }
 
+TEST_F(MaterialTest, DecaySmallAmount) {
+  // eps_decay is defined such that tritium can decay on a 1 day time step
+  const int tritium_id = 10030000;
+  const double qty = 1; //kg, NOTE: fractional amounts all that matter 
+
+  CompMap v;
+  v[tritium_id] = 1.0;
+  Composition::Ptr tritium_comp = Composition::CreateFromMass(v);
+
+  Material::Ptr tritium = Material::Create(fac_day_timestep, qty, tritium_comp);
+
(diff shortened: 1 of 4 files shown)


tests/material_tests.cc
```diff
@@ -338,7 +338,7 @@ TEST_F(MaterialTest, DecayShortcut) {
 
   double sec_per_month = 2629152;
   double u235_lambda = pyne::decay_const(u235) * sec_per_month;  // per month
-  double eps = 1e-3;
+  double eps = 1e-4;
```
connor-welch, line 341: Should this be `eps_decay` instead of `eps`?

connor-welch (commented):
This looks good, @dean-krueger. I only had one comment about the naming of `eps` in the tests.

tests/material_tests.cc
dean-krueger: That's a good question. Since this isn't literally `eps_decay` (it's more like `expected_eps` maybe) I didn't change the name to `eps_decay`, but maybe calling it something like `expected_eps` isn't a bad idea. Thanks for the suggestion!

connor-welch: That makes sense. I think `eps` or `expected_eps` both work.

connor-welch (approved):
I think this is good to be merged, but we can leave it open until after the long weekend so others can look at it.

meg-krieg (approved):
I read through the fix + test. Seems pretty straightforward, I have no other change comments!!

gonuke (approved):
Looks good to me - thanks @dean-krueger

Issue #1832, opened 2024-11-07
Random seed for simulation RNG


nuclearkatie:
Every simulation now has a pseudo-random number generator (#1599) initialized using a seed that can be set in the control block. However, the only current options are to use the default seed, or specify another seed. This makes identical input files repeatable.

But it would also be nice to have a way to run the same simulation with random seeds, i.e. run 10 simulations and have the RNG automatically seeded 10 different ways. Any individual simulation would be still repeatable because the seed is recorded in the input file, which is saved in the output. Reusing that seed would produce the same results.

@gonuke pointed out that setting the seed to -1 might be a good way to do this without introducing another parameter

gonuke:
Some thoughts regarding using a seed value of -1:
Pro: no need for another variable that may be inconsistent with someone also setting a seed
Con: overloading a single variable with slightly different purposes can lead to some user confusion

PR #1950, opened 2026-05-27
Random Seed Based on Date-Time

meg-krieg:
# Summary of Changes

This is intended to resolve issue #1832. The issue described having the user input (-1) as a parameter to indicate to the simulate to generate a random seed for random numbers in the controls block. 
 
## Design Notes

Currently, the user input would actually be 0 instead of -1. I am not super familiar with the schema validation but maybe there is a workaround for this next part -

- I don't know if the xml validation schema can specify that the user can *only* enter + integers and -1 (not 0 or any other negative values). Hence, I just settled on switching the seed input to non-negative integers (include 0 as the option for "random" seed).
 I thought maybe using the date-time cast as an integer to be the random seed when the user wants a random seed. I don't know if it would make sense to use the date time to seed a generator to generate a random seed to generate random numbers for the simulation run... if that makes any sense. So, I just removed the middle part and opted to use date-time directly. 

I would love some comments/critique! 

- [ ] API for Cyclus modules
- [ ] Input files

# Testing and Validation

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -16,7 +16,6 @@
 #include "random_number_generator.h"
 #include "version.h"
 #include <chrono>
-#include <thread>
 
 namespace cyclus {
 
@@ -335,7 +334,6 @@ int Context::random_uniform_int(int low, int high) {
 
 int Context::date_time_int(){
   auto time = std::chrono::steady_clock::now();
-  std::this_thread::sleep_for(std::chrono::milliseconds(5));
   auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(time.time_since_epoch());
   return duration.count();
 }
--- a/src/context.cc
+++ b/src/context.cc
@@ -16,6 +16,7 @@
 #include "random_number_generator.h"
 #include "version.h"
 #include <chrono>
+#include <thread>
 
 namespace cyclus {
 
@@ -333,7 +334,8 @@ int Context::random_uniform_int(int low, int high) {
 }
 
 int Context::date_time_int(){
-  auto time = std::chrono::system_clock::now();
+  auto time = std::chrono::steady_clock::now();
+  std::this_thread::sleep_for(std::chrono::milliseconds(5));
   auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(time.time_since_epoch());
   return duration.count();
 }
--- a/src/context.cc
+++ b/src/context.cc
@@ -15,6 +15,7 @@
 #include "timer.h"
 #include "random_number_generator.h"
 #include "version.h"
+#include <chrono>
 
 namespace cyclus {
 
@@ -329,6 +330,12 @@ double Context::random_01() {
 
 int Context::random_uniform_int(int low, int high) {
   return rng_->random_uniform_int(low, high);
+}
+
+int Context::date_time_int(){
+  auto time = std::chrono::system_clock::now();
+  auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(time.time_since_epoch());
+  return duration.count();
 }
 
 double Context::random_uniform_real(double low, double high) {
--- a/src/context.h
+++ b/src/context.h
@@ -291,6 +291,8 @@ class Context {
   /// Returns a random number from a uniform integer distribution.
   int random_uniform_int(int low, int high);
 
+  int date_time_int();
+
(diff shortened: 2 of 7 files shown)


src/context.h
```diff
@@ -284,6 +284,8 @@ class Context {
   /// Returns a random number from a uniform integer distribution.
   int random_uniform_int(int low, int high);
 
+  int date_time_int();
```
gonuke, line 295: Can you add one or two comment lines to document the purpose of this function?

gonuke (changes requested):
Thanks @meg-krieg - this looks like a good approach.

Just one request for more comments.

I also wonder if there is a good way to test this?

meg-krieg:
I've added a test, but it required an extra line (super mini time lag) in the function to ensure that multiple cyclus simulation runs (or more-so the date-time grabbing) do not happen at the same time.

src/context.cc
```diff
@@ -327,6 +329,13 @@ int Context::random_uniform_int(int low, int high) {
   return rng_->random_uniform_int(low, high);
 }
 
+int Context::date_time_int(){
+  auto time = std::chrono::steady_clock::now();
+  std::this_thread::sleep_for(std::chrono::milliseconds(5));
```
gonuke: If this is just needed for the test to pass, can't we introduce this lag in the test instead?

gonuke (changes requested):
Wondering if we want the lag in the main code or just in the tests?

src/context.cc
```diff
@@ -327,6 +328,12 @@ int Context::random_uniform_int(int low, int high) {
   return rng_->random_uniform_int(low, high);
 }
 
+int Context::date_time_int(){
+  auto time = std::chrono::steady_clock::now();
+  auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(time.time_since_epoch());
+  return duration.count();
+}
```
dean-krueger, line 339: I know the functions below don't do this, but I couldn't help but point out that the Google C++ style reference says:

"Ordinarily, functions follow [PascalCase](https://en.wiktionary.org/wiki/Pascal_case): start with a capital letter and have a capital letter for each new word.

```
AddTableEntry()
DeleteUrl()
OpenFileOrDie()
```

[...]

Accessors and mutators (get and set functions) may be named like variables, in snake_case. These often correspond to actual member variables, but this is not required. For example, int count() and void set_count(int count)."

I don't think this is an accessor or mutator (though I could for sure be wrong). This probably warrants a separate, "beyond v1.7" issue, because I don't think we do a great job of following the style guide (I certainly have a hard time with it a lot of the time), but just in case you cared about this sort of thing, I wanted to bring it up!

src/xml_file_loader.cc
```diff
@@ -512,7 +512,8 @@ void XMLFileLoader::LoadControlParams() {
   cy_eps_rsrc = si.eps_rsrc = eps_rsrc_;
 
   // get seed
-  si.seed = OptionalQuery<int>(qe, "seed", kDefaultSeed);
+  int seed_inp = OptionalQuery<int>(qe, "seed", kDefaultSeed);
```
dean-krueger, line 515: Genuine question, not trying to "point something out", what does "inp" mean?

dean-krueger (commented):
Thanks @meg-krieg! Seems fairly straightforward and good to me. I noticed one thing, and had a question. I'm leaving this as a "comment" so it doesn't block.

src/xml_file_loader.cc
gonuke: I think it means "input" - this is getting the value that is optionally provided in the input file

src/context.cc
gonuke: I think it's effectively an accessor

src/xml_file_loader.cc
meg-krieg: I can change it to seed_input so it's more explicit

dean-krueger: oh gotcha! and not necessary. If you were going to make further changes anyway maybe, but I don't think it's that important.

gonuke (approved):
LGTM - thanks @meg-krieg

gonuke:
This appears to have conflicts from other recent PR's that need to be resolved

meg-krieg:
merge conflict should be fixed now!

Issue #1920, opened 2025-11-14
[Feature] Method to find a specific agent prototype in a context by name


gonuke:
# What is the motivation for your feature request? Please describe what you want to use it for.
There may be times when it is useful to get a pointer to a specific agent based on it's name.  Currently, this capability needs to be implemented in an archetype that wants it, but it would be best to add it to the `context` and be available for all.

# Describe the feature you'd like.
Add a method `cyclus::Agent* GetAgentByName(std::string)` that searches the context's agent list to find an agent with the matching name.

## Describe alternatives you've considered.
This could be implemented by an archetype developer (see `cycamore::TariffRegion`)

# Additional context.
It might also be useful to have `GetRegionByName()`, `GetInstByName()`, and `GetFacilityByName()`

# How can this issue be closed?
Adding methods as suggested above to the `Context` class

nsryan2:
The examples we discussed in the Cyclus community call were:
1. [how it has been done] With "state of the world"-type simulations wherein users have created different prototypes for each reactor in the system they were studying. It was noted that this makes sense as many reactors are bespoke and have different parameters.
2. [where this could be used] At sites where multiple identical units were/would be deployed, or with some advanced reactor designs that will not be bespoke to the deployment location.

gonuke:
> The examples we discussed in the Cyclus community call were:
> 
> 1. [how it has been done] With "state of the world"-type simulations wherein users have created different prototypes for each reactor in the system they were studying. It was noted that this makes sense as many reactors are bespoke and have different parameters.
> 2. [where this could be used] At sites where multiple identical units were/would be deployed, or with some advanced reactor designs that will not be bespoke to the deployment location.

I don't think this is the issue we were referring to on the call, after all (#1899).  

This issue is a simpler follow-on from additions made in #1914 to address #1913.  Those additions were made to support the `TariffRegion` development in Cycamore.  In that development, there are local methods that will use the capability of these methods to find a specific ancestor of the current agent based on it's name.  This issue recommends generating such methods in the Cyclus kernel.

gonuke:
Upon seeing #1951 by @meg-krieg, I wonder if this is a poorly posed issue - and maybe there are other flaws in #1914?

We don't ever actually name individual agents, and only name their prototypes.  In a typical simulation, there may be many agents with the same prototype name, so there may not be a unique answer to "finding an agent by name" if we are actually relying on the prototype name.

meg-krieg:
All I can think of is to 
[...]

PR #1951, opened 2026-05-28
Find Agent by Name 

meg-krieg:
# Summary of Changes

Resolves issue #1920 to add look up capability in context for finding agent by prototype name. 

# Testing and Validation

### DIFF
--- a/src/context.cc
+++ b/src/context.cc
@@ -116,6 +116,16 @@ void Context::DelAgent(Agent* m) {
   }
 }
 
+Agent* Context::GetAgentByName(std::string name){
+  std::set<Agent*>::iterator it;
+  for (it = agent_list_.begin(); it != agent_list_.end(); ++it){
+    if((*it)->prototype() == name){
+      return (*it);
+    }
+  }
+  cyclus::Warn<KEY_WARNING>("Agent set does not contain " + name + " name");
+}
+
 void Context::SchedBuild(Agent* parent, std::string proto_name, int t) {
 #pragma omp critical
   {
--- a/tests/context_tests.cc
+++ b/tests/context_tests.cc
@@ -178,3 +178,26 @@ TEST_F(ContextTests, GetAgentList) {
   delete ctx;
   EXPECT_EQ(3, DonutShop::destruct_count);
 }
+
+TEST_F(ContextTests, GetAgentByName) {
+  Timer ti;
+  Recorder rec;
+  Context* ctx = new Context(&ti, &rec);
+
+  Agent* agent1 = new DonutShop(ctx, "chocolate");
+  Agent* agent2 = new DonutShop(ctx, "chocolate");
+  Agent* agent3 = new DonutShop(ctx, "chocolate");
+
+  agent1->prototype("dunkin donuts");
+  agent2->prototype("krispy kreme");
+  agent3->prototype("voodoo donut");
+
+  ASSERT_NO_THROW(ctx->GetAgentByName("dunkin donuts"));
+
+  cyclus::warn_as_error = true;
+  EXPECT_THROW(ctx->GetAgentByName("donut shop"), cyclus::KeyError);
+
+  cyclus::warn_as_error = false;
+               
+  delete ctx;
+}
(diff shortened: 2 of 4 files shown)


src/context.cc
```diff
@@ -116,6 +116,16 @@ void Context::DelAgent(Agent* m) {
...
 }
 
+Agent* Context::GetAgentByName(std::string name){
+  std::set<Agent*>::iterator it;
+  for (it = agent_list_.begin(); it != agent_list_.end(); ++it){
+    if((*it)->prototype() == name){
+      return (*it);
+    }
+  }
+  cyclus::Warn<KEY_WARNING>("Agent set does not contain " + name + " name");
```
gonuke, line 126:
```suggestion
  cyclus::Warn<KEY_WARNING>("Context set does not contain agent with name: " + name);
```

tests/context_tests.cc
```diff
@@ -177,4 +177,27 @@ TEST_F(ContextTests, GetAgentList) {
...
+  Recorder rec;
+  Context* ctx = new Context(&ti, &rec);
+
+  Agent* agent1 = new DonutShop(ctx, "chocolate");
+  Agent* agent2 = new DonutShop(ctx, "chocolate");
+  Agent* agent3 = new DonutShop(ctx, "chocolate");
+
+  agent1->prototype("dunkin donuts");
+  agent2->prototype("krispy kreme");
+  agent3->prototype("voodoo donut");
```
gonuke, line 193: I recommend setting this test up in the same way as `CreateAgents`.

src/context.cc
```diff
@@ -116,6 +116,16 @@ void Context::DelAgent(Agent* m) {
   }
 }
 
+Agent* Context::GetAgentByName(std::string name){
```
gonuke, line 119: I realize that we don't ever actually name agents, but only name their prototypes,so maybe this should be:
```suggestion
Agent* Context::GetAgentByPrototypeName(std::string name){
```
Or
```suggestion
Agent* Context::GetAgentPrototypeByName(std::string name){
```
?

tests/context_tests.cc
```diff
@@ -177,4 +177,27 @@ TEST_F(ContextTests, GetAgentList) {
...
+
+  ASSERT_NO_THROW(ctx->GetAgentByName("dunkin donuts"));
+
+  cyclus::warn_as_error = true;
+  EXPECT_THROW(ctx->GetAgentByName("donut shop"), cyclus::KeyError);
+
+  cyclus::warn_as_error = false;
+               
+  delete ctx;
```
dean-krueger, line 202: Two thoughts here:

1. Is there a way to check that the function actually gives you the agent (It looks like right now you only check that it doesn't throw when you ask for it)
2. I always ALWAYS forget how this works, but in simulations sometimes we make a facility and call it something like "UraniumMine" and then deploy 50 of them. If I do GetAgentByName(UraniumMine)` this gives the like UraniumMine that's being copied 50 times and not a list of all the copies? If so, is that what we want? If not, what happens if it sees more than one `UraniumMine`? Can you make two facilities with the same name but different implementations? There may be a little bit of edge-case testing that is needed.

I hope that second bullet point makes sense?

dean-krueger (commented):
Thanks @meg-krieg ! This looks really good to me (though many times I have thought that and then a more experienced reviewer brings up some good points). I left a comment about some testing. I have no idea how possible some of that is, but I think it's worthwhile to think a bit about edge cases and maybe try to do deeper testing than just not throwing. I'm going to hit the "Comment" button so that if what I was saying about the tests was nonsense you don't get blocked.

gonuke (changes requested):
Thanks @meg-krieg - I started this review and then went to leave comments in the issue as well on the high level goal here...  There are a few suggestions here, assuming we move forward in some way.

tests/context_tests.cc
gonuke: @dean-krueger this second point occurred to me also and I went to the issue (#1920)  to discuss it further since I think the issue (by me) was not well-posed.

PR #1954, opened 2026-05-29
Changes to cycpp.py to make it less finicky

### LINKED ISSUE #1898: [Bug] Default values unassigned for toolkit Position lat/long causing extra warnings
# Describe the bug.
When the `cyclus::toolkit::Position` injected code is included in an archetype, it will call `InitializePosition()` using injected member variables `latitude` and `longitude`.  Those variables are declared in the cyclus pragma as being default = 0, but there is currently no code that enforces this.  

They are optional, so when an input file does not specify them, their values may be garbage from those memory locations and throw warnings about being out of range.

# Steps to reproduce the bug.
Cycamore unit tests are throwing these errors sometimes/always?

Ensure reasonable default values are specified for `latitude` and `longitude` when injected.

dean-krueger:
# Important Note from Dean
I have, recently, been trying to rely less on using AI to write code, for a variety of reasons. However, the cycpp.py file is so arcane and long that I felt it was appropriate to sick OpenAI's Codex (not Cursor, trying it out) on this particular problem. Not only do I think it fixed the issue I set out to have it fix, I think it ALSO fixed two other bugs with the cycpp.py file, including the one that caused me to have to go in and do `// clang-format off/on` 1,000,000 times in Cycamore (haven't tested this yet to be certain, but I have a hunch). I will still be doing my best to flex my own coding muscles going forward, but this one felt like such a massive time/effort save that it was hard not to just let the superhuman AI do it.

Also, I had Codex write the PR report, since it did 100% of the code writing/exploring/debugging/test writing, etc.

Reviewers: I think this should be fairly straightforward to review, I took a look at what Codex did before making this PR, but please feel free to leave general comments about Codex and the work it did. That will be good feedback.

# Summary of Changes
This PR updates `cycpp.py` so annotated Cyclus state variables can use C++ default member initializers, e.g. `double latitude = 0.0;`. This allows injected cycpp snippets such as `position.cycpp.h` to provide real construction-time defaults instead of requiring every consuming archetype to remember constructor initializer-list entries.

The PR also makes cycpp parsing a little more tolerant of existing pragma forms encountered during downstream Cycamore builds, including explicit code-generation pragmas such as `#pragma cyclus clone` during pass 2 and `#pragma cyclus var{...}` without a space before the annotation dictionary.

# Related CEPs and Issues

This PR is related to:

- Closes #1898 

# Design Notes
The main parser change is intentionally narrow: `VarDeclarationFilter` now only attempts to parse a declaration when a preceding `#pragma cyclus var` annotation is pending. This prevents it from accidentally consuming unrelated pragmas or comments while still allowing initialized declarations like `double latitude = 0.0`.

The generic pragma error filter was updated to recognize cycpp’s explicit code-generation pragmas and no-space `var{...}` syntax as valid forms. This keeps existing Cycamore/cycpp usage working without requiring formatting-only edits in downstream projects.

- [x] API for Cyclus modules
- [x] Input files

# Associated Developers
@dean-krueger

# Testing and Validation
This change was tested with the cycpp unit tests, excluding the existing integration test that assumes a Unix-style `cycpp.py` executable on `PATH` and use of `which` from a Windows shell. The relevant parser tests pass locally, including added coverage for initialized state variable declarations, explicit code-generation pragmas, comments/pragmas near state variable parsing, and `#pragma cyclus var{...}` syntax.

Additional validation was performed by rebuilding Cyclus and running its unit tests locally. The installed cycpp was then used while building Cycamore, which exercised the downstream pragma forms that motivated the compatibility fixes.

# Checklist

### DIFF
--- a/cli/cycpp.py
+++ b/cli/cycpp.py
@@ -397,7 +397,10 @@ class PragmaCyclusErrorFilter(Filter):
     """Filter for handling invalid #pragma cyclus. This should be the last filter."""
     regex = re.compile(r'\s*#\s*pragma\s+cyclus(.*)')
 
-    directives = frozenset(['var', 'note', 'exec', 'decl', 'def', 'impl'])
+    directives = frozenset(['var', 'note', 'exec', 'decl', 'def', 'impl',
+                            'clone', 'initfromcopy', 'initfromdb',
+                            'infiletodb', 'schema', 'annotations',
+                            'initinv', 'snapshotinv', 'snapshot'])
 
     def isvalid(self, statement):
         """Checks if a statement is valid for this fliter."""
@@ -407,7 +410,8 @@ class PragmaCyclusErrorFilter(Filter):
         g1 = m.group(1).strip()
         if len(g1) == 0:
             return False
-        s0 = g1.split(None, 1)[0]
+        m0 = re.match(r'(\w+)', g1)
+        s0 = m0.group(1) if m0 is not None else ''
         return s0 not in self.directives
 
     def transform(self, statement, sep):
@@ -472,7 +476,7 @@ class VarDecorationFilter(DecorationFilter):
     This evals the contents of dict and puts them in state.var_annotations, to be
     consumed by the next match with VarDeclarationFilter.
     """
-    regex = re.compile(r"\s*#\s*pragma\s+cyclus\s+var\s+(.*)")
+    regex = re.compile(r"\s*#\s*pragma\s+cyclus\s+var\s*(.+)")
 
     def transform(self, statement, sep):
         state = self.machine
@@ -485,7 +489,19 @@ class VarDeclarationFilter(Filter):
     """State varible declaration.  Only operates if state.var_annotations is
     not None. Access for member variable must be public.
     """
-    regex = re.compile(r"(.*\w+.*?)\s+(\w+)")
+    regex = re.compile(r"\s*(.+?)\s+(\w+)\s*(?:=.*)?$")
+
+    def isvalid(self, statement):
+        state = self.machine
+        if state.var_annotations is None:
+            self.match = None
+            return False
+        stripped = statement.lstrip()
+        if (stripped.startswith('#') or stripped.startswith('//') or
+                stripped.startswith('/*')):
+            self.match = None
+            return False
+        return super(VarDeclarationFilter, self).isvalid(statement)
 
     def transform_pass2(self, statement, sep):
         state = self.machine
--- a/cli/cycpp.py
+++ b/cli/cycpp.py
@@ -203,7 +203,7 @@ class Filter(object):
         self.match = None
 
     def isvalid(self, statement):
-        """Checks if a statement is valid for this fliter."""
+        """Checks if a statement is valid for this filter."""
         self.match = m = self.regex.match(statement)
         return m is not None
 
@@ -288,7 +288,7 @@ class TypedefFilter(AliasFilter):
         state.aliases |= {(depth, typ, a) for a in g0[-1:] + g[1:]}
 
 class UsingFilter(AliasFilter):
-    """Filter for accumumating using aliases."""
+    """Filter for accumulating using aliases."""
     regex = re.compile(r"\s*using\s+(?!namespace\s+)([\w:]+)\s*")
 
     def transform(self, statement, sep):
@@ -298,7 +298,7 @@ class UsingFilter(AliasFilter):
             state.aliases.add((state.depth, name, name.rsplit('::', 1)[1]))
 
 class NamespaceFilter(Filter):
-    """Filter for accumumating namespace encapsulations."""
+    """Filter for accumulating namespace encapsulations."""
     # handles anonymous namespaces as group(1) is None
     regex = re.compile(r"\s*namespace(\s+\w*)?\s*[^=]*", re.DOTALL)
 
@@ -318,7 +318,7 @@ class NamespaceFilter(Filter):
             del state.namespaces[-1]
 
 class UsingNamespaceFilter(Filter):
-    """Filter for accumumating using namespace statement."""
+    """Filter for accumulating using namespace statement."""
     regex = re.compile(r"\s*using\s+namespace\s+([\w:]*)\s*")
 
     def transform(self, statement, sep):
@@ -337,7 +337,7 @@ class UsingNamespaceFilter(Filter):
                                    if d_ns[0] > depth}
 
 class NamespaceAliasFilter(AliasFilter):
-    """Filter for accumumating namespace renames."""
+    """Filter for accumulating namespace renames."""
     regex = re.compile(r"\s*namespace\s+(\w+)\s*=\s*([\w:]+)\s*")
 
     def transform(self, statement, sep):
@@ -380,7 +380,7 @@ class ClassAndSuperclassFilter(ClassFilter):
             for sup in superclasses:
                 trysup = state.canonize_class(sup)
                 if trysup is None:
-                    # We cannot raise an error here becuase there are too many
+                    # We cannot raise an error here because there are too many
                     # corner cases we do not and should not support in C++
                     continue
                 sc.add(trysup)
@@ -403,7 +403,7 @@ class PragmaCyclusErrorFilter(Filter):
                             'initinv', 'snapshotinv', 'snapshot'])
 
     def isvalid(self, statement):
-        """Checks if a statement is valid for this fliter."""
+        """Checks if a statement is valid for this filter."""
         self.match = m = self.regex.match(statement)
         if m is None:
             return False
@@ -486,7 +486,7 @@ class VarDecorationFilter(DecorationFilter):
         state.var_annotations = self._eval()
 
 class VarDeclarationFilter(Filter):
-    """State varible declaration.  Only operates if state.var_annotations is
+    """State variable declaration.  Only operates if state.var_annotations is
     not None. Access for member variable must be public.
     """
     regex = re.compile(r"\s*(.+?)\s+(\w+)\s*(?:=.*)?$")
@@ -727,7 +727,7 @@ class StateAccumulator(object):
             self.context[classname]['vars'] = OrderedDict()
 
     def accumulate(self, statement, sep):
-        """Modify the existing state by incoprorating the statement, which is
+        """Modify the existing state by incorporating the statement, which is
         partitioned from the next statement by sep.
         """
         self.nlines_since_linemarker += statement.count('\n') + sep.count('\n')
@@ -792,7 +792,7 @@ class StateAccumulator(object):
                 return self._canonize_targs(talias, targs)
             for d, nsa in sorted(self.using_namespaces, reverse=True):
                 if len(tname.split(scopz)) > len(nsa.split(scopz)):
-                    # fixed point of reccursion when type would be more scoped than
+                    # fixed point of recursion when type would be more scoped than
                     # the alias - which is impossible.
                     continue
                 try:
@@ -823,7 +823,7 @@ class StateAccumulator(object):
                 return self.canonize_type(talias, name)
             for d, nsa in sorted(self.using_namespaces, reverse=True):
                 if len(t.split(scopz)) > len(nsa.split(scopz)):
-                    # fixed point of reccursion when type would be more scoped than
+                    # fixed point of recursion when type would be more scoped than
                     # the alias - which is impossible.
                     continue
                 try:
@@ -963,7 +963,7 @@ class CodeGeneratorFilter(Filter):
         classname = groups[1] if len(groups) > 1 else None
         if classname is None:
             if len(cg.classes) == 0:
-                TypeError("{0}Classname could not determined".format(
+                TypeError("{0}Classname could not be determined".format(
                         cg.includeloc()))
             classname = cg.classname()
         classname = classname.strip().replace('.', '::')
@@ -1414,7 +1414,7 @@ class InfileToDbFilter(CodeGeneratorFilter):
         else:
             val = alias[1][0]
 
-        # the extra assignment (bub, sub) is because we want the intial sub
+        # the extra assignment (bub, sub) is because we want the initial sub
         # rhs to be from outer scope - otherwise the newly defined sub will be
         # in scope causing segfaults
         tree_idx = idx or '0'
@@ -1448,7 +1448,7 @@ class InfileToDbFilter(CodeGeneratorFilter):
             val = alias[1]
         else:
             val = alias[1][0]
-        # the extra assignment (bub, sub) is because we want the intial sub
+        # the extra assignment (bub, sub) is because we want the initial sub
         # rhs to be from outer scope - otherwise the newly defined sub will be
         # in scope causing segfaults
         tree_idx = idx or '0'
@@ -1480,7 +1480,7 @@ class InfileToDbFilter(CodeGeneratorFilter):
             val = alias[1]
         else:
             val = alias[1][0]
-        # the extra assignment (bub, sub) is because we want the intial sub
+        # the extra assignment (bub, sub) is because we want the initial sub
         # rhs to be from outer scope - otherwise the newly defined sub will be
         # in scope causing segfaults
         tree_idx = idx or '0'
@@ -1509,7 +1509,7 @@ class InfileToDbFilter(CodeGeneratorFilter):
             alias[1] = 'first'
         if alias[2] is None:
             alias[2] = 'second'
-        # the extra assignment (bub, sub) is because we want the intial sub
+        # the extra assignment (bub, sub) is because we want the initial sub
         # rhs to be from outer scope - otherwise the newly defined sub will be
         # in scope causing segfaults
         tree_idx = idx or '0'
@@ -1536,7 +1536,7 @@ class InfileToDbFilter(CodeGeneratorFilter):
             alias[1] = 'key'
         if alias[2] is None:
             alias[2] = 'val'
-        # the extra assignment (bub, sub) is because we want the intial sub
+        # the extra assignment (bub, sub) is because we want the initial sub
         # rhs to be from outer scope - otherwise the newly defined sub will be
         # in scope causing segfaults
         # subtree must be specified if in recursive level
@@ -2131,7 +2131,7 @@ class CodeGenerator(object):
         return s + "\n"
 
     def generate(self, statement, sep):
-        """Modify the existing statements list by incoprorating, modifying, or
+        """Modify the existing statements list by incorporating, modifying, or
         ignoring this statement, which is partitioned from the next statement by
         sep.
         """
@@ -2226,7 +2226,7 @@ class Proxy(MutableMapping):
         return key in self.__dict__['_d']
 
 def outter_split(s, open_brace='(', close_brace=')', separator=','):
-    """Takes a string and only split the outter most level."""
+    """Takes a string and only split the outermost level."""
     outter = []
     ns = s.split(separator)
     count = 0
--- a/tests/cycpp_test.py
+++ b/tests/cycpp_test.py
@@ -168,10 +168,13 @@ def test_synerror():
     m = MockMachine()
     f = PragmaCyclusErrorFilter(m)
     assert not f.isvalid("#pragma cyclus var {}")
+    assert not f.isvalid("#pragma cyclus var{}")
     assert not f.isvalid("#pragma cyclus")
+    assert not f.isvalid("#pragma cyclus clone")
+    assert not f.isvalid("#pragma cyclus annotations")
+    assert not f.isvalid("#pragma cyclus snapshotinv")
 
-    assert  f.isvalid('#pragma cyclus nooooo')
-    statement, sep = "#pragma cyclus var{}", "\n"
+    statement, sep = "#pragma cyclus nooooo", "\n"
     assert  f.isvalid(statement)
     with pytest.raises(SyntaxError):
         f.transform(statement, sep)
@@ -190,6 +193,11 @@ def test_vdecorfilter():
     f.transform(statement, sep)
     assert m.var_annotations == {'name': 'James Bond'}
 
+    statement, sep = "#pragma cyclus var{'name': 'Moneypenny'} ", "\n"
+    assert  f.isvalid(statement)
+    f.transform(statement, sep)
+    assert m.var_annotations == {'name': 'Moneypenny'}
+
 def test_vdeclarfilter():
     """Test VarDeclarationFilter"""
     m = MockMachine()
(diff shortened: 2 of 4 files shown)


cli/cycpp.py
```diff
@@ -397,7 +397,10 @@ class PragmaCyclusErrorFilter(Filter):
...
     regex = re.compile(r'\s*#\s*pragma\s+cyclus(.*)')
 
-    directives = frozenset(['var', 'note', 'exec', 'decl', 'def', 'impl'])
+    directives = frozenset(['var', 'note', 'exec', 'decl', 'def', 'impl',
+                            'clone', 'initfromcopy', 'initfromdb',
+                            'infiletodb', 'schema', 'annotations',
+                            'initinv', 'snapshotinv', 'snapshot'])
 
     def isvalid(self, statement):
         """Checks if a statement is valid for this fliter."""
```
gonuke: Just noticed this, so while we're here...?


```suggestion
        """Checks if a statement is valid for this filter."""
```

cli/cycpp.py
```diff
@@ -397,7 +397,10 @@ class PragmaCyclusErrorFilter(Filter):
     """Filter for handling invalid #pragma cyclus. This should be the last filter."""
     regex = re.compile(r'\s*#\s*pragma\s+cyclus(.*)')
 
-    directives = frozenset(['var', 'note', 'exec', 'decl', 'def', 'impl'])
+    directives = frozenset(['var', 'note', 'exec', 'decl', 'def', 'impl',
+                            'clone', 'initfromcopy', 'initfromdb',
+                            'infiletodb', 'schema', 'annotations',
+                            'initinv', 'snapshotinv', 'snapshot'])
```
gonuke, line 403: Can we determine whether the lack of these in this list was an oversight or had some intention?  It's suspicious to me that there have been so many missing entries.

cli/cycpp.py
gonuke, line 512: The regex that establishes this match should see up to 3 groups now.

cli/cycpp.py
```diff
@@ -407,7 +410,8 @@ def isvalid(self, statement):
         g1 = m.group(1).strip()
         if len(g1) == 0:
             return False
-        s0 = g1.split(None, 1)[0]
+        m0 = re.match(r'(\w+)', g1)
+        s0 = m0.group(1) if m0 is not None else ''
         return s0 not in self.directives
```
gonuke, line 415: I'll confess that I'm a little baffled by the logic here, but it always worked....

cli/cycpp.py
```diff
@@ -407,7 +410,8 @@ def isvalid(self, statement):
         g1 = m.group(1).strip()
         if len(g1) == 0:
             return False
-        s0 = g1.split(None, 1)[0]
+        m0 = re.match(r'(\w+)', g1)
+        s0 = m0.group(1) if m0 is not None else ''
```
gonuke, line 414: I think this is part of the change that provides flexibility regarding whitespace after a directive.  In the old version, it did a `string.split()` which split on whitespace.  In this version is does a regex match on `\w+` which matches any contiguous set of "word" characters.

cli/cycpp.py
```diff
@@ -472,7 +476,7 @@ class VarDecorationFilter(DecorationFilter):
     This evals the contents of dict and puts them in state.var_annotations, to be
     consumed by the next match with VarDeclarationFilter.
     """
-    regex = re.compile(r"\s*#\s*pragma\s+cyclus\s+var\s+(.*)")
+    regex = re.compile(r"\s*#\s*pragma\s+cyclus\s+var\s*(.+)")
```
gonuke, line 479: This is a second change to be flexible with whitespace after the 'var' directive.  This is pretty straightforward, changing a minimum of 1 whitespace character to a minimum of 0 whitespace characters.

cli/cycpp.py
```diff
@@ -485,7 +489,19 @@ class VarDeclarationFilter(Filter):
     """State varible declaration.  Only operates if state.var_annotations is
     not None. Access for member variable must be public.
     """
-    regex = re.compile(r"(.*\w+.*?)\s+(\w+)")
+    regex = re.compile(r"\s*(.+?)\s+(\w+)\s*(?:=.*)?$")
```
gonuke, line 492: I think this is to allow the declaration of default values on cyclus variables, but I'm not sure I understand the regex's well enough to know why it makes a difference.  

It is clear that the new regex explicitly matches the assignment ("=") at the end, but doesn't appear to use it. Maybe that doesn't matter, and just consuming it in the regex is all that it needs to do, to leave the other matches "clean".

This now appears to match:
* 0 or more whitespace, followed by
* 1 or more other characters (optionally??), followed by
* 1 or more whitespace, followed by
* a single contiguous word, followed by
* 0 or more whitespace, followed by
* optional "=" attached to anything up to the end of the line

The original regex appears to have matched:
* 0 or more of non-word characters, followed by
* a single contigous word, followed by
* 0 or more non-word, non-space characters, followed by
* 1 or more whitespace, followed by
* a single contiguous word

It seems to me that both should match fine on: "int foo = 12;"

cli/cycpp.py
```diff
@@ -485,7 +489,19 @@ class VarDeclarationFilter(Filter):
...
+        state = self.machine
+        if state.var_annotations is None:
+            self.match = None
+            return False
+        stripped = statement.lstrip()
+        if (stripped.startswith('#') or stripped.startswith('//') or
+                stripped.startswith('/*')):
+            self.match = None
+            return False
+        return super(VarDeclarationFilter, self).isvalid(statement)
```
gonuke, line 504: Do we understand why we need this new validation test?

tests/cycpp_test.py
```diff
@@ -168,10 +168,13 @@ def test_synerror():
...
     f = PragmaCyclusErrorFilter(m)
     assert not f.isvalid("#pragma cyclus var {}")
+    assert not f.isvalid("#pragma cyclus var{}")
     assert not f.isvalid("#pragma cyclus")
+    assert not f.isvalid("#pragma cyclus clone")
+    assert not f.isvalid("#pragma cyclus annotations")
+    assert not f.isvalid("#pragma cyclus snapshotinv")
 
-    assert  f.isvalid('#pragma cyclus nooooo')
-    statement, sep = "#pragma cyclus var{}", "\n"
```
gonuke, line 174: Does this suggest that this version with no space was always meant to be valid??

tests/cycpp_test.py
```diff
@@ -168,10 +168,13 @@ def test_synerror():
...
+    statement, sep = "#pragma cyclus nooooo", "\n"
```
gonuke, line 177: I'm not sure I understand why this is valid, but it was always meant to be valid.

tests/cycpp_test.py
```diff
@@ -190,20 +193,51 @@ def test_vdecorfilter():
...
+    assert m.var_annotations == {'name': 'Moneypenny'}
+
 def test_vdeclarfilter():
     """Test VarDeclarationFilter"""
     m = MockMachine()
     f = VarDeclarationFilter(m)
     assert not f.isvalid("one ")
 
     statement, sep = "one two", "\n"
+    m.var_annotations = {}
```
gonuke, line 209: Do we know why this is necessary? I'm pretty sure it never gets checked.

tests/cycpp_test.py
```diff
@@ -190,20 +193,51 @@ def test_vdecorfilter():
...
     assert  f.isvalid(statement)
     m.classes = [(0, "trader")]
     m.access = {"trader": "public"}
     # m.var_annotations = {'name': 'James Bond'}
+    m.var_annotations = None
```
gonuke: Is this necessary? correct? Is there any chance that this test wants to ensure that this gets set to `None` by the transform rather than being initialized as `None` and still being `None`?

Related: I'm suspicious of the commented out non-None initialization of this...

tests/cycpp_test.py
```diff
@@ -190,20 +193,51 @@ def test_vdecorfilter():
...
+
+def test_vdeclarfilter_skips_pragmas_and_comments():
+    """Test VarDeclarationFilter leaves non-declarations for later filters"""
+    m = MockMachine()
+    m.var_annotations = {}
+    f = VarDeclarationFilter(m)
+
+    assert not f.isvalid("#pragma cyclus")
+    assert not f.isvalid("// a comment")
+    assert not f.isvalid("/* a comment */")
```
gonuke, line 251: Does it matter where these comments are?  Presumably we were always skipping comments in some locations, and this must only apply to comments that are somehow in the vicinity of other things we do?

Related: could we have a comment at the end of a variable declaration?
```
#pragma cyclus var {}
int foo = 1; // this is a comment about this variable
```

gonuke (changes requested):
This is awesome @dean-krueger - thanks for diving in to this.  I've left many comments that are just for discussion and building our collective intiution.  Feel free to respond with just a thumbs up if you agree with my assessment or any other repsonse.

cli/cycpp.py
dean-krueger: For a second I thought that Codex made this mistake and I couldn't believe my eyes. I can fix that for sure, though! As per usual, I'll probably do it manually just because it hurts my brain to do it by hitting the "commit suggestion" button on github for some reason.

cli/cycpp.py
dean-krueger: I'm not sure we can, but I can ask codex if it has any ideas. My suspicion is that, since the preprocessor does a bunch of passes to try to narrow down what the `#pragma cyclus whatever` is, this was just a case of "the old one worked coincidentally and the file is really complicated" and so nobody ever caught it.

cli/cycpp.py
dean-krueger: This file has the unfortunate combination of being really long, and having a bunch of variables that are named stuff like `m0` and `g1` and it's always been really hard for me to parse. I can have Codex do a little write-up explaining what's going on here, though.

cli/cycpp.py
dean-krueger: # Codex's Thoughts

I did the archaeology. My read is: this was almost certainly an oversight / drift, not an intentional exclusion.

Evidence:

- PragmaCyclusErrorFilter was introduced in commit 69746b15 with message: “Added syntax error for malformed #pragma cyclus directives.”
- In that original commit, the allowed list was exactly:
```
['var', 'note', 'exec', 'decl', 'def', 'impl']
```
- But before that commit, cycpp already had codegen filters with pragmanames like clone, initfromcopy, initfromdb, infiletodb, schema, annotations, snapshot, snapshotinv, and initinv.
- Also before that commit, tests/cycpp_tests.h already contained explicit pragmas like:
```
#pragma cyclus clone
#pragma cyclus initfromcopy
#pragma cyclus schema
#pragma cyclus snapshot
```
So those entries were not absent because they didn’t exist yet.

My best explanation: the original list was written as a quick malformed-pragma guard around the “front door” cycpp decorators and mode keywords: var, note, exec, plus decl/def/impl. But cycpp has two overlapping concepts:

- directive/mode words: var, note, exec, decl, def, impl
- codegen decorator names: clone, schema, snapshot, etc.
The error filter only included the first group. Since it is installed in both pass 2 and pass 3, explicit codegen pragmas visible during pass 2 can fall through to the error filter unless they’re included in its valid-directive list.

I don’t see evidence of an intended policy like “explicit codegen pragmas should be invalid here.” The codegen filter docstrings explicitly document forms like #pragma cyclus [def|decl|impl] clone [classname], and existing tests/headers used them. So I’d answer your advisor: “I think this was a historical completeness bug in the error filter, likely masked by limited coverage/path differences, rather than an intentional exclusion.”

# Dean's Thoughts
Seems sort of like what I was thinking.

cli/cycpp.py
dean-krueger: This filter is the “last chance” cycpp pragma validator. It only fires when a statement looks like `#pragma cyclus ...`, but none of the earlier, more specific filters handled it. Its job is to distinguish “valid cycpp pragma that another pass/filter should handle” from “probably malformed cycpp pragma.”

Line by line:

```python
self.match = m = self.regex.match(statement)
if m is None:
    return False
```

If the statement is not a `#pragma cyclus...` statement at all, this filter is irrelevant.

```python
g1 = m.group(1).strip()
if len(g1) == 0:
    return False
```

If the statement is just bare `#pragma cyclus`, that is valid shorthand for the default codegen pragma, so the error filter should not claim it.

The changed bit:

```python
m0 = re.match(r'(\w+)', g1)
s0 = m0.group(1) if m0 is not None else ''
return s0 not in self.directives
```

Previously this used:

```python
s0 = g1.split(None, 1)[0]
```

That works only when the directive is separated from its arguments by whitespace, e.g.:

```cpp
#pragma cyclus var {"default": 0}
```

But it misclassifies this valid existing style:

```cpp
#pragma cyclus var{"default": 0}
```

For that string, `g1` is:

```python
'var{"default": 0}'
```

The old `split()` sees the whole thing as the first token:

```python
'var{"default":'
```

or effectively a non-`var` token, so the error filter thinks it is an unknown directive.

The new regex extracts only the leading word characters from `g1`. So:

```python
re.match(r'(\w+)', 'var{"default": 0}').group(1)
```

returns:

```python
'var'
```

That lets the filter correctly recognize `var{...}` as a valid `var` directive. The fallback `''` is just defensive: if `g1` starts with punctuation or some other odd malformed text, it will not match a known directive and the filter will report it as invalid.

This part:

```python
re.match(r'(\w+)', g1)
```

does not know about `"default"` or `{...}` at all. It only says: “starting at the beginning of `g1`, capture one or more word characters.”

So for examples:

```python
'var{"default": 0}'        -> 'var'
'var {"default": 0}'       -> 'var'
'clone'                   -> 'clone'
'snapshotinv cycamore::X' -> 'snapshotinv'
'nooooo whatever'         -> 'nooooo'
'{not_a_word}'            -> no match
```

The point is to extract the directive name from the front of the pragma, regardless of whether the next character is whitespace, `{`, or something else. Then cycpp checks whether that extracted directive is in `self.directives`.

cli/cycpp.py
dean-krueger: Yep! It would seem so.

cli/cycpp.py
dean-krueger: Yes with the nuance that Codex also changed (.*) to (.+), so the annotation body must contain at least one character. That keeps bare `#pragma cyclus var` from being treated as a valid var annotation (or so says Codex)

cli/cycpp.py
dean-krueger: From Codex (all of these are from codex but I want to make it super clear):

Your intuition is right on the important bit: the assignment part is not used directly. It is consumed so that the captured groups stay clean.

The old regex was:

```python
r"(.*\w+.*?)\s+(\w+)"
```

It does match `int foo = 12`, but not in the way we want. Because it is not anchored and group 1 can expand oddly, it captures:

```python
'int foo =' , '12'
```

So cycpp thinks the variable type is `int foo =` and the variable name is `12`.

The new regex is:

```python
r"\s*(.+?)\s+(\w+)\s*(?:=.*)?$"
```

For `int foo = 12`, it captures:

```python
'int', 'foo'
```

Breakdown:

```python
\s*
```

Ignore leading whitespace.

```python
(.+?)
```

Capture the type. The `?` here makes `+` non-greedy, meaning “take as little as possible while still allowing the rest of the regex to match.” This helps stop at the variable name instead of swallowing too much.

```python
\s+
```

Require whitespace between type and variable name.

```python
(\w+)
```

Capture the variable name.

```python
\s*(?:=.*)?$
```

Optionally consume an initializer from `=` to the end of the statement, but do not capture it.

So the difference is not “can it match?” but “what does it capture?” cycpp only uses:

```python
vtype, vname = self.match.groups()
```

so the captures must be exactly type and variable name.

One subtlety: the old regex works for plain declarations like:

```cpp
double latitude;
```

but for initialized declarations like:

```cpp
double latitude = 0.0;
int foo = 12;
```

it can treat the initializer value as the variable name. The new regex prevents that by making the initializer an explicit trailing, non-captured part of the declaration match.

cli/cycpp.py
dean-krueger: Again, from Codex:
We need this because `VarDeclarationFilter` became broad enough to recognize more valid C++ declarations, but that also made it capable of accidentally recognizing non-declarations.

The important context is that cycpp processes each statement by trying filters in order and stops at the first filter whose `isvalid()` returns true.

Before this change, `VarDeclarationFilter` could match some things it should not, but its narrower regex happened not to catch bare pragmas. After updating the regex to handle initialized declarations, it could match:

```cpp
#pragma cyclus
```

as if it were:

```python
type = "#pragma"
name = "cyclus"
```

That was bad because `VarDeclarationFilter` appears before `DefaultPragmaFilter` in pass 3. So it swallowed bare `#pragma cyclus` before `DefaultPragmaFilter` could expand it into generated methods like `Clone`, `InitInv`, and `SnapshotInv`. The symptom was that agents became abstract classes.

This guard fixes that by saying: only try to parse a variable declaration if cycpp is actually expecting one, meaning a previous `#pragma cyclus var ...` has set:

```python
state.var_annotations
```

So:

```python
if state.var_annotations is None:
    self.match = None
    return False
```

means “there is no pending state-variable annotation, so this cannot be a cyclus state variable declaration.”

The comment/pragma checks are the second safety layer:

```python
stripped = statement.lstrip()
if (stripped.startswith('#') or stripped.startswith('//') or
        stripped.startswith('/*')):
```

Those prevent cycpp from consuming pragmas or comments as declarations even when a `#pragma cyclus var` is pending. That matters for cases like:

```cpp
#pragma cyclus var {...}
// comment about the variable
double x;
```

or another preprocessor line/comment appearing between annotation and declaration. We want comments/pragmas to pass through to other filters or remain inert, not become the annotated variable declaration.

So the short version is: this validation is needed because `VarDeclarationFilter` should be stateful. It should only activate in the narrow context where cycpp is waiting for the declaration following `#pragma cyclus var`. Without that, the broader initializer-aware regex can steal statements from later filters.

tests/cycpp_test.py
dean-krueger: No. In fact, that test suggests the opposite.

The confusing part is the method name:

```python
PragmaCyclusErrorFilter.isvalid(...)
```

Here, “valid” means “valid for this error filter to handle,” not “valid cycpp syntax.”

So the old test:

```python
assert f.isvalid('#pragma cyclus nooooo')
statement, sep = "#pragma cyclus var{}", "\n"
assert f.isvalid(statement)
with pytest.raises(SyntaxError):
    f.transform(statement, sep)
```

meant:

```cpp
#pragma cyclus var{}
```

was expected to be caught by the error filter and raise `SyntaxError`.

That lines up with the old `VarDecorationFilter` regex:

```python
r"\s*#\s*pragma\s+cyclus\s+var\s+(.*)"
```

which required whitespace after `var`. So `var{}` did not match the real var filter and fell through to the error filter.

Our change intentionally changes that behavior. We found real downstream cycamore code using:

```cpp
#pragma cyclus var{"default": [], "internal": True}
```

so we made cycpp accept both forms. I would not describe that as “always meant to be valid”; I’d describe it as “previously considered malformed by the unit test, but now accepted for compatibility with existing downstream usage and formatter behavior.”

# Comment from Dean
This also has downstream effects for our clang-format PR, which was only necessary BECAUSE clang-format wants there to be no whitespace there AND cycpp.py needed to see whitespace, further justifying changing this.

tests/cycpp_test.py
dean-krueger: # Dean's thoughts
I think, if I'm reading Codex's assessment on your previous question correctly, `#pragma cyclus nooooo` was supposed to be "valid" as something to filter out, aka not valid syntax?

tests/cycpp_test.py
dean-krueger: It's necessary now because `VarDeclarationFilter.isvalid()` checks whether cycpp is currently expecting a variable declaration.

In real cycpp flow, this happens in two steps:

```cpp
#pragma cyclus var {...}
double x;
```

The first line is handled by `VarDecorationFilter`, which sets:

```python
state.var_annotations = {...}
```

Then `VarDeclarationFilter` is allowed to treat the next declaration as the annotated state variable.

In the unit test, we are testing `VarDeclarationFilter` directly, without first running `VarDecorationFilter`. So this line:

```python
m.var_annotations = {}
```

mocks the “a `#pragma cyclus var` was just seen” state.

It does get checked here:

```python
if state.var_annotations is None:
    self.match = None
    return False
```

An empty dict is enough because the check is only “is it `None`?” not “does it contain data?” So:

```python
None
```

means “not waiting for a var declaration,” while:

```python
{}
```

means “waiting for a var declaration, with empty annotations.”

That matches valid cycpp syntax too:

```cpp
#pragma cyclus var {}
int x;
```

So the short answer is: it is test setup for the new stateful `isvalid()` behavior. Without it, `f.isvalid("one two")` correctly returns false.

tests/cycpp_test.py
dean-krueger: Your suspicion is fair.

In the current test, this line:

```python
m.var_annotations = None
```

is necessary only because earlier in the same test we now do:

```python
m.var_annotations = {}
assert f.isvalid(statement)
```

That setup is needed because `VarDeclarationFilter.isvalid()` now only returns true when a var annotation is pending.

But then, if we left `m.var_annotations = {}` and called:

```python
f.transform(statement, sep)
```

the test would no longer be exercising the old “no-op when no annotation is pending” behavior. It would try to process `"one two"` as a real state variable declaration. With the current mock setup, that path is probably not even fully valid because `m.access` is keyed oddly:

```python
m.access = {"trader": "public"}
```

where the production code expects:

```python
m.access[tuple(state.classes)]
```

So the `m.var_annotations = None` line is basically restoring the old test condition before calling `transform()`.

That said: I agree this is a confusing test now. It does **not** prove that `transform()` clears `var_annotations` from non-`None` to `None`. It proves that calling `transform()` while no annotation is pending remains a no-op.

The commented-out line:

```python
# m.var_annotations = {'name': 'James Bond'}
```

does look like someone once considered testing the real transform path, but left it disabled. A clearer future cleanup would probably split this into two tests:

1. `isvalid()` requires pending annotations.
2. `transform()` consumes pending annotations and sets `var_annotations` back to `None`, with a properly configured mock.

So: necessary for the current test shape, but not ideal. Your concern is valid.

# Dean's thoughts
This seems like it'd probably be good to fix/clarify now while we're tinkering with this. I'll go ahead and have Codex do this now and fix that typo you saw earlier and commit a change in a bit.

tests/cycpp_test.py
dean-krueger: Yes, location matters.

Those test lines are meant to cover comments that appear while cycpp is “waiting” for the declaration after a `#pragma cyclus var`, for example:

```cpp
#pragma cyclus var {}
// comment about foo
int foo;
```

or:

```cpp
#pragma cyclus var {}
/* comment about foo */
int foo;
```

Because `VarDeclarationFilter` is now active only while `var_annotations` is pending, a comment between the pragma and the declaration is exactly the dangerous case. Without the comment guard, the broader declaration regex could misread:

```cpp
// a comment
```

as something like:

```python
type = "// a"
name = "comment"
```

and consume the pending annotation before it reaches the real variable. So those tests are specifically about comments in the immediate post-pragma vicinity.

For an end-of-line comment like:

```cpp
#pragma cyclus var {}
int foo = 1; // this is a comment about this variable
```

I believe cycpp’s statement splitting should see the declaration first, ending at the semicolon:

```cpp
int foo = 1
```

Then the `// ...` comment is a separate later statement. By then `VarDeclarationFilter.transform()` should have consumed the annotation and reset `var_annotations` to `None`, so the comment should not matter.

The case I would be less confident about is an inline block comment inside the declaration itself, e.g.:

```cpp
#pragma cyclus var {}
int /* comment */ foo = 1;
```

or:

```cpp
#pragma cyclus var {}
int foo /* comment */ = 1;
```

Those are valid C++, but cycpp’s regex-based parsing may or may not handle them cleanly. I would not assume this PR supports that unless tested. The current comment checks are mainly protecting comments as standalone statements between the annotation and the declaration.

# Dean's Thoughts
If people are leaving comments like

```
int foo /*I have something REALLY important to say about foo*/ = 1;
```

I personally want something to break. That's a crazy way to leave a comment in my opinion.

dean-krueger:
Went through and had Codex (I may have accidentally called it cursor at some point, but it was always Codex) respond to your comments, since I genuinely don't know very much about this at all and it clearly does. I occasionally left my own thoughts on the comments it had (clearly marked, hopefully). I know you said just thumbs up, but I wanted to provide as much clarity as I could since I still had the context window up (and it may forget its thinking when I close it later). It's a lot to read, but I think provides a lot of clarity on the stuff you were wondering about.

gonuke:
> Went through and had Codex (I may have accidentally called it cursor at some point, but it was always Codex) respond to your comments, since I genuinely don't know very much about this at all and it clearly does. I occasionally left my own thoughts on the comments it had (clearly marked, hopefully). I know you said just thumbs up, but I wanted to provide as much clarity as I could since I still had the context window up (and it may forget its thinking when I close it later). It's a lot to read, but I think provides a lot of clarity on the stuff you were wondering about.

The discussion here is a useful way to build some capacity and understanding about what's going on.  Having Codex chime in is useful too.

cli/cycpp.py
gonuke: Not true - the 3rd group is non-capturing `(?:)`

gonuke (approved):
Thanks for your work recruiting Codex to this task @dean-krueger - and for making sure we all (both?) understand what's happening.

PR #1956, opened 2026-06-01
Element Specified in Recipes

### LINKED ISSUE #1818: Docs seem to imply that recipes can contain element-only IDs when they must be nuclides
Documentation for the use of [nucname](https://github.com/cyclus/cyan/blob/0dccb3c4e1cec35b03271320325c11a7092cc253/nuc/nucname.h#L374C1-L374C42) in Cyclus implies that use of a Z number only (i.e., just the element) is valid input for a recipe

```
For integer input the form resolution order is:
  ///   - id
  ///   - zz (elemental z-num only given)
...
```

That implies that this could be a valid definition for U-10Zr fuel:
```
<recipe>
    <name>fuel</name>
    <basis>mass</basis>
    <nuclide> <id>U235</id> <comp>0.002250</comp> </nuclide>
    <nuclide> <id>U238</id> <comp>0.897631</comp> </nuclide>
    <nuclide> <id>40</id> <comp>0.1</comp> </nuclide>
</recipe>
```

This does pass nucname checks. Using the z-number 40 is valid nucname id for elemental Zr with natural abundance, but using this in a recipe fails CompMap checks because 40 is not a valid _nuclide_. 

It's fair that only nuclides can be used for recipes, although I wish I could declare an element and have the composition automatically recalculated per nuclide using natural abundances. Zr is a great example of this. It seems like we could do this by using `natural_abund()` from PyNE, but we don't currently. 

But as things stand now it should be clearer in the docs that while nucname itself is fine with an element as id, that's not valid for defining a recipe, which requires definition at the nuclide level

meg-krieg:
# Summary of Changes

Added two functions for accepting elements as a material specification in a recipe. These functions based on whether recipe is mass or atom fraction-based will split elements into its isotopes based on natural abundances. Recipes that are atom-frac based I believe have to undergo 2 conversions (to and from mass fraction) such that this split can occur. 

## Design Notes

I noticed there was some issues within the ``pyne::Material::expand_elements()`` that I had to adjust. I think these need to be more properly vetted...When these changes are not included, for example, carbon gets split into C-13 and N-14 (changes zz number). 

- [ ] API for Cyclus modules
- [ ] Input files

# Related CEPs and Issues

Fixes #1818 

- [] I have added documentation for new or changed features
- [ ] This code follows the style guide

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,6 +34,7 @@ Since last release
 * Allow multiple archetype blocks to facilitate includes (#1874)
 * New composition specification based on single nuclide (#1949) 
 * Added ability to calcuate a time shift between the start time and some other time stamp (#1907)
+* Recipes can accept id entries as entries (#1956)
 
 **Changed:**
 * Modified cycpp.py to fix a few whitespace-related bugs, and allow cyclus vars to be initialized (#1954)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,7 +34,7 @@ Since last release
 * Allow multiple archetype blocks to facilitate includes (#1874)
 * New composition specification based on single nuclide (#1949) 
 * Added ability to calcuate a time shift between the start time and some other time stamp (#1907)
-* Recipes can accept id entries as entries (#1956)
+* Recipes can accept element entries as well as nuclides (#1956)
 
 **Changed:**
 * Modified cycpp.py to fix a few whitespace-related bugs, and allow cyclus vars to be initialized (#1954)
--- a/src/comp_math.cc
+++ b/src/comp_math.cc
@@ -84,7 +84,7 @@ CompMap MaterialMF(const CompMap& v) {
   return zz_expand.mult_by_mass();
   }
 
-bool ValidSpecies(const CompMap& v) {
+bool ValidNucs(const CompMap& v) {
   CompMap::const_iterator it;
   for (it = v.begin(); it != v.end(); ++it) {
     if (!pyne::nucname::isnuclide(it->first) && !pyne::nucname::iselement(it->first)) {
--- a/src/comp_math.cc
+++ b/src/comp_math.cc
@@ -70,7 +70,7 @@ void Normalize(CompMap* v, double val) {
   }
 }
 
-CompMap MaterialAF(const CompMap& v) {
+CompMap ExpandAtomComp(const CompMap& v) {
   pyne::Material zz;
   zz.from_atom_frac(v); //pyne materials assume mass-frac for expand_elements 
   pyne::Material zz_expand = zz.expand_elements();
@@ -78,7 +78,7 @@ CompMap MaterialAF(const CompMap& v) {
   return nuclides;
 }
 
-CompMap MaterialMF(const CompMap& v) {
+CompMap ExpandMassComp(const CompMap& v) {
   pyne::Material zz(v);
   pyne::Material zz_expand = zz.expand_elements();
   return zz_expand.mult_by_mass();
--- a/src/comp_math.cc
+++ b/src/comp_math.cc
@@ -72,7 +72,7 @@ void Normalize(CompMap* v, double val) {
 
 CompMap MaterialAF(const CompMap& v) {
   pyne::Material zz;
-  zz.from_atom_frac(v); //pyne materials assume mass-frac compmap
+  zz.from_atom_frac(v); //pyne materials assume mass-frac for expand_elements 
   pyne::Material zz_expand = zz.expand_elements();
   CompMap nuclides = zz_expand.to_atom_frac();
   return nuclides;
--- a/src/comp_math.cc
+++ b/src/comp_math.cc
@@ -74,8 +74,7 @@ CompMap ExpandAtomComp(const CompMap& v) {
   pyne::Material zz;
   zz.from_atom_frac(v); //pyne materials assume mass-frac for expand_elements 
   pyne::Material zz_expand = zz.expand_elements();
-  CompMap nuclides = zz_expand.to_atom_frac();
-  return nuclides;
+  return zz_expand.to_atom_frac();
 }
 
 CompMap ExpandMassComp(const CompMap& v) {
--- a/src/comp_math.cc
+++ b/src/comp_math.cc
@@ -70,10 +70,24 @@ void Normalize(CompMap* v, double val) {
   }
 }
 
-bool ValidNucs(const CompMap& v) {
+CompMap MaterialAF(const CompMap& v) {
+  pyne::Material zz;
+  zz.from_atom_frac(v); //pyne materials assume mass-frac compmap
+  pyne::Material zz_expand = zz.expand_elements();
+  CompMap nuclides = zz_expand.to_atom_frac();
+  return nuclides;
+}
+
+CompMap MaterialMF(const CompMap& v) {
+  pyne::Material zz(v);
+  pyne::Material zz_expand = zz.expand_elements();
+  return zz_expand.mult_by_mass();
+  }
+
+bool ValidSpecies(const CompMap& v) {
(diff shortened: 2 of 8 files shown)


src/pyne.cc
gonuke, line 1: The PyNE code is amalgamated from the PyNE repo and we should not change it locally.  I'm not sure why it needs to be changed, so I'll look into that separately.

src/comp_math.h
```diff
@@ -38,8 +38,16 @@ void ApplyThreshold(CompMap* v, double threshold);
 /// The sum of quantities of all nuclides of v is normalized to val.
 void Normalize(CompMap* v, double val = 1.0);
 
-/// Returns true if all nuclide keys in v are valid.
-bool ValidNucs(const CompMap& v);
```
gonuke, line 42: We may want to keep this function since it's part of the public interface.  Removing it may make various archetypes suddenly break.

gonuke (changes requested):
Thanks @meg-krieg...

I've made a few comments already, but before I continue, I wonder if there is functionality built into PyNE to help us with this?

src/comp_math.h
meg-krieg: This function was just renamed as ``ValidSpecies`` to be explicit about the fact it was checking whether entries were also elements. I could just rename it to its original form but keep the check for elements in it.

gonuke: Yes, but some functions may be calling `ValidNucs` somewhere still that we don't know about.

src/pyne.cc
meg-krieg: Ok! I agree it didn't seem like the right move but some odd effects were happening, and I noticed (maybe incorrectly) that I don't think ```expand_elements()``` was converting natural abundances from percentages

meg-krieg:
@gonuke is this a better fix? There are no edits to pyne now

src/comp_math.h
```diff
@@ -38,7 +38,15 @@ void ApplyThreshold(CompMap* v, double threshold);
...
 void Normalize(CompMap* v, double val = 1.0);
 
-/// Returns true if all nuclide keys in v are valid.
+/// creates pyne material and returns nuclide-mass frac map
+//in case an element needs to be expanded
+CompMap MaterialMF(const CompMap& v);
+
+/// creates pyne material and returns nuclide-atom frac map
+//in case an element needs to be expanded
+CompMap MaterialAF(const CompMap& v);
```
dean-krueger: I think these are MaterialMF (Mass Fraction) and AF (Atom Fraction), but slightly more descriptive function names might be nice for future readers. It may also be slightly confusing since we have Cyclus "materials" which are distinct from CompMaps which is what this returns. I don't have a great suggestion for what you could call this, but I felt confused when I read this the first time because of that.

dean-krueger (commented):
Thanks @meg-krieg! I had one comment about function names (unfortunately it's a bit of a "this might need another look" without much helpful direction).

src/comp_math.h
meg-krieg: yeah... that's a valid point. I will fix!

src/comp_math.cc
```diff
@@ -70,10 +70,24 @@ void Normalize(CompMap* v, double val) {
...
+  zz.from_atom_frac(v); //pyne materials assume mass-frac for expand_elements 
+  pyne::Material zz_expand = zz.expand_elements();
+  CompMap nuclides = zz_expand.to_atom_frac();
+  return nuclides;
+}
+
+CompMap ExpandMassComp(const CompMap& v) {
+  pyne::Material zz(v);
+  pyne::Material zz_expand = zz.expand_elements();
+  return zz_expand.mult_by_mass();
```
gonuke, line 83: Why do you `mult_by_mass`? Cyclus uses normalized compositions anyway....

Maybe you just wanted a method that returned a comp_map?

Perhaps `get_dens_frac(frac_type="mass")` will get the comp map you want with a clearer purpose

src/comp_math.cc
```diff
@@ -70,10 +70,24 @@ void Normalize(CompMap* v, double val) {
   }
 }
 
+CompMap ExpandAtomComp(const CompMap& v) {
+  pyne::Material zz;
+  zz.from_atom_frac(v); //pyne materials assume mass-frac for expand_elements 
+  pyne::Material zz_expand = zz.expand_elements();
+  CompMap nuclides = zz_expand.to_atom_frac();
+  return nuclides;
```
gonuke: Looking to make this more aligned with the other method, so maybe you can use `get_dens_frac(frac_type="atom")` here instead, and skip the intermediate variable

gonuke (changes requested):
A couple of nuanced points for long term clarity

src/comp_math.cc
meg-krieg: No, ```pyne::Material``` ```.comp``` can pretty easily return a CompMap but I did ``.mult_by_mass()`` because there is a unit test outside of CompMath tests that fail due the output from ``expand_elements`` being normalized. I wasn't sure where else the impact trickled -- similar to changing the name of ``ValidNucs``. But I can just update that unit test (``CapacityConstraintTests.MaterialQuality``)

src/comp_math.cc
gonuke:
```suggestion
  return zz_expand.to_atom_frac();
```

gonuke (changes requested):
Let's just streamline a single line here and go with it like this.

gonuke:
We also need a CHANGELOG entry @meg-krieg

CHANGELOG.rst
```diff
@@ -34,6 +34,7 @@ Since last release
 * Allow multiple archetype blocks to facilitate includes (#1874)
 * New composition specification based on single nuclide (#1949) 
 * Added ability to calcuate a time shift between the start time and some other time stamp (#1907)
+* Recipes can accept id entries as entries (#1956)
```
gonuke:
```suggestion
* Recipes can accept element entries as well as nuclides (#1956)
```

gonuke (changes requested):
Small change log clarification (& maybe it will run tests this time?)

src/composition.cc
```diff
@@ -15,24 +15,26 @@ namespace cyclus {
 int Composition::next_id_ = 1;
 
 Composition::Ptr Composition::CreateFromAtom(CompMap v) {
-  if (!compmath::ValidNucs(v)) throw ValueError("invalid nuclide in CompMap");
+   if (!compmath::ValidNucs(v)) throw ValueError("invalid nuclide or element in CompMap");
```
gonuke, line 18: Indentation typo
```suggestion
  if (!compmath::ValidNucs(v)) throw ValueError("invalid nuclide or element in CompMap");
```

tests/comp_math_tests.cc
```diff
@@ -4,11 +4,50 @@
...
+  //will not be normalized here 
+
+  CompMap expand_set = cm::ExpandMassComp(e);
+
+  float c12_atom_abund = 0.98938;
+  float c13_atom_abund = 0.01078;
+  float c12_wo_abund = pyne::atomic_mass(60120000)*c12_atom_abund/pyne::atomic_mass(60000000);
+  float c13_wo_abund = pyne::atomic_mass(60130000)*c13_atom_abund/pyne::atomic_mass(60000000);
+
+  CompMap::iterator it;
```
gonuke, line 44: This is unused
```suggestion
```

tests/comp_math_tests.cc
```diff
@@ -4,11 +4,50 @@
 #include "composition.h"
 #include "cyc_limits.h"
 #include "error.h"
+#include "pyne.h"
+#include "env.h"
```
gonuke, line 8: Not sure if/why we need this header?

gonuke (changes requested):
A few minor comments - and then with a rebase for CHANGELOG - it'll be ready to go.

PR #1959, opened 2026-06-11
add more documentation to the schema

gonuke:
# Summary of Changes
Ask `cycpp` to add variable documentation from the metadata in more locations, esp. complex data types.  Previously, this documentation was only available when the type was primitive. Not it will be added for lists, vectors, maps, pairs, etc.

## Design Notes
This required the addition of a method to properly "escape" text for XML safety.


abachma2 (approved):
Approving, and merging with the tests now passing

PR #1961, opened 2026-06-16
Fix the Progress Bar Breaking Cycamore Build

dean-krueger:
# Summary of Changes

Apparently the progress bar was breaking Cycamore's build (I guess I somehow NEVER checked that, which is odd because usually I'm pretty careful about that, but I guess this slipped through the cracks). This PR addresses the root cause of that issue by adding a non-default copy constructor to the Timer class which explicitly avoids copying the Progress Bar, which is of type `std::unique_ptr`, which cannot be copied (the cause of the error)

## Design Notes

The root cause of this one was apparently:

1. MockSim starts up and tries to copy the Timer
2. Timer doesn't have an explicit copy function (I think) so the compiler tries to invent one (this is normal and standard I think, the same thing happened with the constructors when I just deleted them since they were default)
3. The copy made by the constructor is blind to what's actually going on in timer, so it just says "yeah grab everything, copy it, and throw it over there"
4.  the progress_bar_  object is an std::unique_ptr because the one you found online says it has to be
5. std::unique_ptr objects are non-copyable (I don't know the details about this, but apparently this is the fundamental problem)
6. when the compiler-generated constructor tries to copy the progress bar, the error gets thrown because it can't copy.
7. Because the progress bar can't copy, C++ throws the "deleted function" error, since it THINKS that timer should have that constructor (at least I think it was the constructor that threw the error) but it can't use it because the progress bar couldn't be copied and is blocking it. 
8. Broken Cycamore build.

I had Codex draft up what it thought would be a fix (to give Timer a copy constructor that was careful about the progress bar so the compiler didn't invent one) and it basically looks like this:

```
Timer::Timer(const Timer& other) {
  *this = other;
}

Timer& Timer::operator=(const Timer& other) {
  if (this != &other) {
    ctx_ = other.ctx_;
    time_ = other.time_;
    si_ = other.si_;
    want_snapshot_ = other.want_snapshot_;
    want_kill_ = other.want_kill_;
    tickers_ = other.tickers_;
    cpp_tickers_ = other.cpp_tickers_;
    py_tickers_ = other.py_tickers_;
    build_queue_ = other.build_queue_;
    decom_queue_ = other.decom_queue_;
    progress_bar_.reset();
    progress_update_frequency_ = other.progress_update_frequency_;
    progress_origin_ = other.progress_origin_;
    progress_span_ = other.progress_span_;
    quiet_ = other.quiet_;
  }
  return *this;
}
```

Which on a first pass allows Cyclus and Cycamore to build and both sets of tests to pass. When I asked Codex to explain what it had done, it flagged that the two major steps were this:

```
Timer::Timer(const Timer& other) {
  *this = other;
}
```

Which is the "copy constructor" it was talking about. This function literally just makes a new timer object  (`this`) and then points the new (copied) Timer to `other`  (the original timer, I have no idea why they're named like that but I assume it's some standard). HOWEVER, what does `Timer = Timer` mean? Well, we define that here:

```
Timer& Timer::operator=(const Timer& other) {
  if (this != &other) {
    ctx_ = other.ctx_;
    time_ = other.time_;
    si_ = other.si_;
    want_snapshot_ = other.want_snapshot_;
    want_kill_ = other.want_kill_;
    tickers_ = other.tickers_;
    cpp_tickers_ = other.cpp_tickers_;
    py_tickers_ = other.py_tickers_;
    build_queue_ = other.build_queue_;
    decom_queue_ = other.decom_queue_;
    progress_bar_.reset();
    progress_update_frequency_ = other.progress_update_frequency_;
    progress_origin_ = other.progress_origin_;
    progress_span_ = other.progress_span_;
    quiet_ = other.quiet_;
  }
  return *this;
}
```
Which just steps through everything that a timer is and sets it normally EXCEPT when it gets to the progress bar part. There, it says "don't copy the progress bar. Just reset it and make a new one" so the `std::unique_ptr` never gets a copy called on it and doesn't complain.

- [x] API for Cyclus modules
- [x] Input files

# Related CEPs and Issues

Fixes #???

# Associated Developers

Codex 🖥️

# Testing and Validation

I clean built cyclus and cycamore and ran BOTH sets of tests with this new fix. 

### DIFF
--- a/.github/workflows/changelog_test.yml
+++ b/.github/workflows/changelog_test.yml
@@ -15,7 +15,7 @@ jobs:
     name: Is Changelog up-to-date ?
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
 
       - run: |
           git config --global --add safe.directory ${GITHUB_WORKSPACE}
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ Since last release
 ====================
 
 **Added:**
-
+* Added a copy constructor to Timer to allow MockSim to copy it without upsetting the progress bar (#1961)
 * Added progress bar to the simulation loop (#1912)
 * Added a warning for when a facility trades with itself (#1895)
 * Added a new datatype to the backend for tariff region (#1922)
--- a/src/timer.cc
+++ b/src/timer.cc
@@ -19,6 +19,31 @@
 
 namespace cyclus {
 
+Timer::Timer(const Timer& other) {
+  *this = other;
+}
+
+Timer& Timer::operator=(const Timer& other) {
+  if (this != &other) {
+    ctx_ = other.ctx_;
+    time_ = other.time_;
+    si_ = other.si_;
+    want_snapshot_ = other.want_snapshot_;
+    want_kill_ = other.want_kill_;
+    tickers_ = other.tickers_;
+    cpp_tickers_ = other.cpp_tickers_;
+    py_tickers_ = other.py_tickers_;
+    build_queue_ = other.build_queue_;
+    decom_queue_ = other.decom_queue_;
+    progress_bar_.reset();
+    progress_update_frequency_ = other.progress_update_frequency_;
+    progress_origin_ = other.progress_origin_;
+    progress_span_ = other.progress_span_;
+    quiet_ = other.quiet_;
+  }
+  return *this;
+}
+
 void Timer::RunSim() {
   LogLevel saved_level = Logger::ReportLevel();
   if (quiet_) {
--- a/src/timer.h
+++ b/src/timer.h
@@ -25,6 +25,10 @@ class Timer {
   friend class ::SimInitTest;
 
  public:
+  Timer() = default;
+  ~Timer() = default;
+  Timer(const Timer& other);
+  Timer& operator=(const Timer& other);
 
   /// Sets intial time-related parameters for the simulation.
   ///
--- a/tests/timer_tests.cc
+++ b/tests/timer_tests.cc
@@ -344,6 +344,36 @@ TEST_P(TimerTestsFixture, DoubleDecom) {
   cyclus::PyStop();
 }
 
+TEST_P(TimerTestsFixture, Copy) {
+  cyclus::Recorder rec;
+  cyclus::Timer ti;
+  cyclus::Timer ti2;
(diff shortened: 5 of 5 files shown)


gonuke (changes requested):
Thanks @dean-krueger - can we add some simple tests for the new copy and assignment constructors?

dean-krueger:
Sure can!

dean-krueger:
okay, it's saying that the only failing test is the changelog, but I double checked and I 100% updated that, so not sure what's going on there. Otherwise this should be good to go @gonuke.

gonuke:
> okay, it's saying that the only failing test is the changelog, but I double checked and I 100% updated that, so not sure what's going on there. Otherwise this should be good to go @gonuke.

It may be a change under the hood at GitHub.

Can you try to update the version of the `checkout` action in the `changelog` test to v7?  (Not sure what this test is failing on this and not the others?)

dean-krueger:
> Can you try to update the version of the `checkout` action in the `changelog` test to v7? (Not sure what this test is failing on this and not the others?)

Done. Let's see if that helped.

dean-krueger:
> Done. Let's see if that helped.

It did not.

gonuke:
So... I'm OK with moving away from a rebasing workflow - looks like you merged in changes from the upstream - and let's go with that!!

But somehow we have ended up with many of the changes that you merged in still appearing as changes/diffs in this PR... so I wonder how to fix that without rebasing???

dean-krueger:
oh man, see that's weird because my git settings have `git config pull.rebase` set to `true`, so I thought I was rebasing? Maybe when I sync my remote branch with the main cyclus one it does this? Either way, they shouldn't appear as new in the diff, though, I would think. Clearly there's something about git that I don't understand.

Do you have any advice on how to proceed both with:

1. This specific PR
2. PRs in the future?

Happy to try whatever. It should be fairly easy in this PR to tell what is and isn't progress bar (if not I can outline it), but in a larger PR with a bigger difference between it and the main branch this could become very troublesome.

gonuke:
Can you try merging `main` into this branch to see if it cleans things up - I'm trying to let go of my rebase habit

dean-krueger:
okay, I think it did? The timer stuff is all still there by the looks of it, and a bunch of those files that looked changed last time are gone. Sorry it took me a second to get around to this!

gonuke (approved):
LGTM - waiting for tests

PR #1966, opened 2026-06-30
Apply changes for deacy_absorb in clean branch

### LINKED ISSUE #1841: Bulk ResBufs have dubious decay behavior

gonuke:
# Summary of Changes
Force both materials involved in an absorption to decay before absorption.
Replaces #1918 after some git drama

## Design Notes
Previous behavior would result in dubious decay times for materials following absorption

**Importantly**, this PR introduces a new constraint on material decay:
* tracked materials can only decay forward and up to the current context time

- [X] API for Cyclus modules
- [X] Input files

# Related CEPs and Issues
Fixes #1841

# Associated Developers
@dean-krueger 

# Testing and Validation

- [ ] I have compiled and run the code locally
- [X] I have updated the changelog

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 *.sw*
 *.orig
-build
+build/
 .DS_Store
 debug
 core
@@ -88,3 +88,7 @@ cyclus/cycpp.py
 # Rever
 rever/
 .vscode
+
+# Coverage
+tests/html/*
+*coverage.info
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -40,7 +40,7 @@ Since last release
 * Users can specify for random seed to be created for random number generation (#1950)
 
 **Changed:**
-* Clarify behavior of `Material::Absorb()` with respect to decay ()
+* Clarify behavior of `Material::Absorb()` with respect to decay (#1966)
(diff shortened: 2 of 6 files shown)


gonuke:
@dean-krueger I think it's fine for you to review/merge this

dean-krueger:
Sounds good. I take it that means you don't think we should have a 3rd party do that in some capacity? Happy to be the one to give it another look over, just wanted to make sure.

gonuke:
Go for it...

dean-krueger (approved):
Thanks @gonuke! Sorry this took me a minute to get around to. This appears to have everything we wanted in it, and I double checked that this builds alongside Cycamore main and that both sets of tests pass on my machine. I think this is good to go!

PR #1970, opened 2026-07-24
Set booleans with `true` and `false`

### LINKED ISSUE #1969: [Bug] Booleans cannot be set with `true` or `t`, only `1` despite clear intent for those to be supported in `infile_tree.h`
# Describe the bug.

infile_tree.h has a function called `inline bool OptionalQuery(args)` 

```
template <>
inline bool OptionalQuery(InfileTree* tree, std::string query,
                          bool default_val) {
  if (tree->NMatches(query) == 0) {
    return default_val;
  }

  std::string s = tree->GetString(query);
  boost::trim(s);
  std::transform(s.begin(), s.end(), s.begin(), ::tolower);
  return s == "true" || s == "t" || s == "1";
}
```

which shows clear intent for cyclus XML reading to support users entering boolean "TRUE" values as `true`, `t`, or `1`, but currently it seems like Cyclus ONLY supports booleans being set with 1 or 0. This appears to be a problem with cycpp.py and the way it reads `#pragma cyclus var foo {...}` values. Specifically, when cycpp.py reads the `#pragma` bit, it replaces it with some generated code, and the generated code uses the generic Query:

```
template <typename T>
inline T Query(InfileTree* tree, std::string query, int index = 0) {
  return boost::lexical_cast<T>(tree->GetString(query, index).c_str());
}
```

which for some reason I didn't care to really look into only accepts bools as `0` or `1`. This seems unintentional, since I think that `OptionalQuery` function just never gets used as it is, and it would certainly be nicer as input to have:

```
<foobar>true</foobar>
```

instead of:

```
<foobar>1</foobar> // harder to see that foobar is a bool here
```

# Steps to reproduce the bug.

1. Ran a cyclus input which had a facility which takes a bool and set that bool to `true` in the input
2. Error: bad lexical cast

```
terminate called after throwing an instance of 'boost::wrapexcept<boost::bad_lexical_cast>'
  what():  bad lexical cast: source type value could not be interpreted as target
Aborted (core dumped)
```

 - Operating System:
 - Cyclus Version:
 - Python Version:

# Additional context.

dean-krueger:
# Summary of Changes

This PR adds a boolean-specific Query function to infile_tree.h which cycpp.py can grab to allow cyclus XML input for booleans to be entered as `true` and `false` instead of just `1` and `0`.

This PR also switches the order of the `prev_decay_time` update and the record inside `Material::Transmute`. This was done here instead of in its own PR so that I could give Amer a branch that worked for him to continue making progress. I figured since it was such a small change it was fine to glob onto this.

## Design Notes

There are maybe (not necessarily "probably") more tests which could be done as a part of a more thorough investigation about how this all works and whether other variables have similar problems, but this PR is meant mostly as a patch to get things working properly quickly to support Amer.

The bug that this fixes was discovered by Cypher use (Cypher tries to insert `true` and `false` not `1` or `0`) and the options were either to change cypher (worse, because 1 and 0 are less obviously booleans in the XML) or to do this in Cyclus.

The issue describes `t` and `f` as possibilities, but some testing during implementation revealed that this broke something else if you tried it, and that was a rabbit hole I did not want to go down, since this is a little time-sensitive.

### NOTE: 
This PR does NOT make it so you can use `True` or `False` or `TRUE` or `FALSE` (or `trUE` or `FaLsE`). It must be `true` or `false` or `1` or `0` and that's an unrelated thing with the way Cyclus handles booleans in general. If we want to fix that we (probably) can, but I would STRONGLY suggest that we do so as a separate issue with a separate PR since that feels like yet another rabbit hole.

- [X] API for Cyclus modules
- [X] Input files

# Related CEPs and Issues

Fixes #1969 

# Associated Developers

Codex

# Testing and Validation

Manual testing and unit tests both added. Code compiles and runs on my computer. Tests all pass.

- [X] I have compiled and run the code locally
- [X] I have added or updated relevant tests

### DIFF
--- a/src/infile_tree.h
+++ b/src/infile_tree.h
@@ -90,6 +90,18 @@ inline std::string Query(InfileTree* tree, std::string query, int index) {
   return tree->GetString(query, index);
 }
 
+template <> inline bool Query(InfileTree* tree, std::string query, int index) {
+  std::string s = tree->GetString(query, index);
+  boost::trim(s);
+  if (s == "true" || s == "1") {
+    return true;
+  }
+  if (s == "false" || s == "0") {
+    return false;
+  }
+  return boost::lexical_cast<bool>(s.c_str());
+}
+
 template <> inline int Query(InfileTree* tree, std::string query, int index) {
   std::string s = tree->GetString(query, index);
   boost::trim(s);
--- a/src/infile_tree.h
+++ b/src/infile_tree.h
@@ -93,10 +93,10 @@ inline std::string Query(InfileTree* tree, std::string query, int index) {
 template <> inline bool Query(InfileTree* tree, std::string query, int index) {
   std::string s = tree->GetString(query, index);
   boost::trim(s);
-  if (s == "true" || s == "1") {
+  if (s == "true") {
     return true;
   }
-  if (s == "false" || s == "0") {
+  if (s == "false") {
     return false;
   }
   return boost::lexical_cast<bool>(s.c_str());
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Since last release
 ====================
 
 **Added:**
+* Added a boolean-specific Query function to infile_tree.h to support archetype bools using true and false (#1970)
 * Added a copy constructor to Timer to allow MockSim to copy it without upsetting the progress bar (#1961)
 * Added progress bar to the simulation loop (#1912)
 * Added a warning for when a facility trades with itself (#1895)
--- a/src/material.cc
+++ b/src/material.cc
@@ -156,7 +156,6 @@ void Material::Absorb(Material::Ptr mat) {
 
 void Material::Transmute(Composition::Ptr c) {
   comp_ = c;
-  tracker_.Modify();
 
   // Presumably the user has chosen the new composition to be accurate for
   // the current simulation time.  The next call to decay should not include
@@ -168,6 +167,9 @@ void Material::Transmute(Composition::Ptr c) {
   if (ctx_ != NULL && ctx_->time() > prev_decay_time_) {
     prev_decay_time_ = ctx_->time();
   }
+
+  // Record the composition and updated decay timestamp together.
+  tracker_.Modify();
 }
 
 Resource::Ptr Material::PackageExtract(double qty,
--- a/tests/infile_tree_tests.cc
+++ b/tests/infile_tree_tests.cc
@@ -152,3 +152,25 @@ TEST_F(InfileTreeTest, optional_queries) {
   EXPECT_EQ(str_val, OptionalQuery<string>(&qe, str_str, str_other));
   EXPECT_EQ(str_other, OptionalQuery<string>(&qe, other, str_other));
 }
+
+// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+TEST_F(InfileTreeTest, bool_queries) {
+  using cyclus::Query;
+
+  std::stringstream ss;
+  ss << "<root>"
+     << "<true> true </true>"
+     << "<false>false</false>"
+     << "<one>1</one>"
+     << "<zero>0</zero>"
+     << "</root>";
+
+  cyclus::XMLParser parser;
+  parser.Init(ss);
+  cyclus::InfileTree qe(parser);
+
+  EXPECT_TRUE(Query<bool>(&qe, "true"));
+  EXPECT_FALSE(Query<bool>(&qe, "false"));
+  EXPECT_TRUE(Query<bool>(&qe, "one"));
+  EXPECT_FALSE(Query<bool>(&qe, "zero"));
+}


src/infile_tree.h
```diff
@@ -90,6 +90,18 @@ inline std::string Query(InfileTree* tree, std::string query, int index) {
   return tree->GetString(query, index);
 }
 
+template <> inline bool Query(InfileTree* tree, std::string query, int index) {
+  std::string s = tree->GetString(query, index);
+  boost::trim(s);
+  if (s == "true" || s == "1") {
```
gonuke: Should we correct for case? Something like this? (Not sure of syntax)
```suggestion
  if (s.to_lower() == "true" || s == "1") {
```

src/infile_tree.h
```diff
@@ -90,6 +90,18 @@ inline std::string Query(InfileTree* tree, std::string query, int index) {
...
+    return true;
+  }
+  if (s == "false" || s == "0") {
+    return false;
+  }
+  return boost::lexical_cast<bool>(s.c_str());
```
gonuke, line 102: Doesn't this handle the case of 0/1?

gonuke (changes requested):
A couple of robustness questions... also, do we want to handle "t" and "f"?

src/infile_tree.h
dean-krueger: So we CAN do that (edit: we cannot, I spoke too soon), but it won't help since the throw happens with the XML itself and not with parsing it. When you give things a "boolean" type XML is allowed to accept "true/false" or "1/0", but XML doesn't recognize "True/False", so it will throw a fit BEFORE we get to this part where we try to do `s.to_lower()`. The same is true if we tried "t/f", so we can't support that either (I think). The most upvoted answer in [this stack overflow](https://stackoverflow.com/questions/1308491/xsdboolean-element-type-accept-true-but-not-true-how-can-i-make-it-accept) puts it nicely.

dean-krueger: Further evidence, we get a failure on Schema Validation when we try "True" instead of a bad_lexical_cast, which indicates that the C++ casting isn't the issue, but rather that XML is complaining because it doesn't like `True`. Note that this file runs when I use `true` instead. The same thing happens when I do `t`. It's just that XML doesn't accept those as options.

Screenshot 2026-08-24 at 10 06 35 AM

src/infile_tree.h
dean-krueger: Yes? Not totally sure what your question is here, but perhaps if you rephrased I could give you the insight you're looking for? :)

dean-krueger:
# NOTE
I added one extra thing to this PR: inside of `Material::Transmute` I simply switched the order of the `prev_decay_time` update and the record. Previously, it recorded and THEN updated the time, which meant the time was wrong. I just made it so that it now records after so that the time is correct in the output database.

src/infile_tree.h
gonuke: Gotcha ... wonder if it's worth doing anyway to future-proof an evolving XML capability/standard

dean-krueger: I don't think so, especially because it is my understanding that there is no equivalent to a `std::string.to_lower()` function. There's a char version that you'd have to loop over the whole string for I think, called `tolower()`, but that seems like a pain, especially for something that doesn't do anything. An example of how to lowercase a string I found is below:

```
    // Convert the string to lowercase in-place
    std::transform(str.begin(), str.end(), str.begin(), [](unsigned char c) {
        return std::tolower(c);
    });
```

I have no idea why C++ doesn't have a built-in function to do this, but it seems like it doesn't.

src/infile_tree.h
gonuke: I think I was wondering if you needed to handle "1" and "0" explicitly above if it will just be done here?

dean-krueger: oohhhhhhh, maybe not? Let me check if it works without the 1/0 bit.

dean-krueger: Yeah they seem to be redundant. Good catch!

dean-krueger:
Also, I really do not understand why these tests are failing. I don't think it's something I did, so maybe this is another valgrind angle?

gonuke:
Might need to run tests locally to see if we can figure out what's going on...

dean-krueger:
I mean, they all pass for me, certainly. You're talking about spinning up one of these container types and running them that way or something?

gonuke:
> I mean, they all pass for me, certainly. You're talking about spinning up one of these container types and running them that way or something?

Nope - was just making sure they were tested somewhere other than CI

dean-krueger:
ah, gotcha. I'll re-run them on my machine now just to double check that everything is good to go, but I definitely reran them yesterday also. One sec!

dean-krueger:
609/609 passing on my machine!

gonuke:
It's only failing on MacOS now and that looks like a Mac specific issue - I'll try to figure it out soon

PR #1974, opened 2026-09-11
Fix decay absorb

### LINKED ISSUE #1973: [Bug] In `lazy` Decay Mode, if Decay Threshold causes decay to be skipped and then Absorb happens, materials never decay
# Describe the bug.

If you're in `lazy` decay mode, and you're working with a material/timestep combination (say, tritium - 12hrs) that causes Material::Decay to skip decay (below the threshold which is currently pegged to tritium at 1d timesteps) and then immediately call absorb on that material (this happens in DecayStorage when you're bringing in new Tritium inventory from CANDUs), Material::Absorb "decays" both of those materials (which does nothing) and then updates the prev_decay_time_ of the combined material to the **current timestep**. This causes those materials to NEVER decay. 

# Steps to reproduce the bug.

1. Simulation dt = 12 hrs (very non-standard, but supposedly allowed/supported)
2. An experimental source which allows just-in-time material creation supplies tritium to Tricycle::DecayStorage. This tritium was created on the current timestep.
3. DecayStorage absorbs the new tritium material into its bulk resource buffer. This calls Decay on both materials which does nothing because the threshold for material decay is pegged to tritium decaying on 1d timesteps.
4. Material::Absorb explicitly/manually sets the combined material's prev_decay_time to the current timestep.
5. This happens EVERY time step, causing the tritium inventory of DecayStorage to never decay.

## What inputs and outputs are involved?

There are a few files involved here but the bug description should hopefully be enough. I'm probably the one to fix this anyway. Hopefully that doesn't come back to bite anyone in the butt.

 - Operating System:
 - Cyclus Version: dean-krueger/cyclus `bool-time` branch
 - Python Version: 3.12.3
 - Other Relevant Software Versions: dean-krueger/tricycle `add-power` branch, unpushed cycamore `just-in-time-source` branch (the Source creates its material every time step instead of all at the beginning of the simulation)

The DecayStorage should be able to handle this and decay the tritium correctly.

# Additional context.

It is probably possible to fix this quickly by lowering the threshold, but the "correct" fix will be in Material::Absorb and involve checking whether the material ACTUALLY decayed before setting the prev_decay_time_ to the current time step.

Fixing the bug? idk. Change Material::Absorb.

dean-krueger:
# NOTE:
This PR branch was based off of my `bool-time` branch because I needed it to be all together for Amer to use (if it works). Because of that the diff lists bool-time changes too. If we merge bool-time first hopefully that all goes away and the diff gets cleaned up. If you're interested in the actual changes in THIS pr, they are:

- everything in `resource.h`
- everything in `material.h`
- changes to `Decay()` and `Absorb()` in `material.cc`, but not to `Transmute()`
- the appropriate change in changelog

# Summary of Changes

Decay in Material::Absorb was misbehaving. In lazy decay mode, if new material entered a bulk resbuf with a halflife/timestep combination (ie Tritium, 12h) it would never decay. This PR adds a "forced" decay mode to Material and Resource which ignores the `eps_decay` check and decays regardless of what it sees, then uses that mode in Absorb.

## Design Notes

I originally thought of making Decay a bool and just returning whether it was successful, but Codex pointed out to me that this ran the risk of trading an under-decay bug for an over-decay bug. Imagine a simulation in lazy mode:

> T0 - Tritium in DecayStorage with `prev_decay_time_` = 0
> T1 - Tritium enters DecayStorage with `prev_decay_time` = 1. Absorb determines that no decay can happen.
> 
> Scenario A (Under/Never Decay):
> T2 - The combined material's `prev_decay_time` was updated to T1, this causes the infinite-update-never-decay loop we are trying to avoid
> 
> Scenario B (Over Decay):
> T2 - The combined material's `prev_decay_time` was left at T0. Now we meet the threshold for decay, and decay 2 time steps. However, qty_new / qty_old fraction of the material only needed to be decayed by one time step! 

So instead I decided that there needed to be some way to just **force** decay to happen in Absorb so that we could guarantee that the `prev_decay_time_` was correct.

One other option I didn't fully consider because it seemed like more work and I was trying to test that this was actually the issue I was seeing in a simulation was to just WARN the user that the decay wasn't happening because of the threshold and then give them a route to update that threshold manually through the control block or something, but that feels like it'd basically be the same thing with more steps like 99+% of the time. I think if this comes up as an edge case where someone is Absorbing Pu in a bulk resbuf and their simulations are taking too long we can revisit this then, perhaps.

- [x] API for Cyclus modules
- [x] Input files

# Related CEPs and Issues

Fixes #1973 

# Associated Developers

Codex

# Testing and Validation

## TESTS HAVE NOT BEEN ADDED DO NOT MERGE

Tests were performed manually, no formal tests added yet, but I'll do that once general approval for this method is given.

- [ x I have compiled and run the code locally

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Since last release
 ====================
 
 **Added:**
+* Added forced decay mode for use inside Material::Absorb() to correctly handle lazy mode with small timesteps and half-lives (#1974)
 * Added a boolean-specific Query function to infile_tree.h to support archetype bools using true and false (#1970)
 * Added a copy constructor to Timer to allow MockSim to copy it without upsetting the progress bar (#1961)
 * Added progress bar to the simulation loop (#1912)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Since last release
 ====================
 
 **Added:**
+* Added a boolean-specific Query function to infile_tree.h to support archetype bools using true and false (#1970)
 * Added a copy constructor to Timer to allow MockSim to copy it without upsetting the progress bar (#1961)
 * Added progress bar to the simulation loop (#1912)
 * Added a warning for when a facility trades with itself (#1895)
--- a/src/infile_tree.h
+++ b/src/infile_tree.h
@@ -90,6 +90,18 @@ inline std::string Query(InfileTree* tree, std::string query, int index) {
   return tree->GetString(query, index);
 }
 
+template <> inline bool Query(InfileTree* tree, std::string query, int index) {
+  std::string s = tree->GetString(query, index);
+  boost::trim(s);
+  if (s == "true" || s == "1") {
+    return true;
+  }
+  if (s == "false" || s == "0") {
+    return false;
+  }
+  return boost::lexical_cast<bool>(s.c_str());
+}
+
 template <> inline int Query(InfileTree* tree, std::string query, int index) {
   std::string s = tree->GetString(query, index);
   boost::trim(s);
--- a/src/infile_tree.h
+++ b/src/infile_tree.h
@@ -93,10 +93,10 @@ inline std::string Query(InfileTree* tree, std::string query, int index) {
 template <> inline bool Query(InfileTree* tree, std::string query, int index) {
   std::string s = tree->GetString(query, index);
   boost::trim(s);
-  if (s == "true" || s == "1") {
+  if (s == "true") {
     return true;
   }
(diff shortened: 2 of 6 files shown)


gonuke:
Leaving a note here from discussion:

If we wanted the user to be able to guide the threshold for when decay occurs, it is probably some function of the following:
* longest half-life of interest
* duration of the simulation
* time step of the simulation


That is, for half-lives of interest that are much longer than the duration of the simulation, we may not care whether decay happens accurately.

